PR #26, opened 2021-06-07
Implement nback.llvm support for structure types and external function deceleration

### LINKED ISSUE #24: Implement nback.llvm support for struct types
Add support for:

- Define structure types (minimally `{i64, i1}`) and use them as a type of a value / return type of function 
- Use `extractvalue` so we can get values out of them by index

heshanpadmasiri:
## Purpose
- Add minimum support to implement structures so we can use structure types such as `{i64,i1}`
- Change the function code so we can support external (functions without a body) function declarations 
Resolve #24 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/nback.llvm/llvm.bal
+++ b/compiler/modules/nback.llvm/llvm.bal
@@ -21,7 +21,12 @@ public type PointerType readonly & record {|
     Alignment align;
 |};
 
-public type Type IntType|PointerType;
+// Corresponds to llvm::StructType
+public type StructType readonly & record {|
+    Type[] elementTypes;
+|};
+
+public type Type IntType|PointerType|StructType;
 
 // A RetType is valid only as the return type of a function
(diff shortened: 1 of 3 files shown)


jclark (changes requested):
The declare/define stuff is a separate change, and I am not convinced that part is right. Please do that separately.

In particular, there is a specific API for declaring intrinsics: LLVMGetIntrinsicDeclaration

There should also be a function to create a struct type following LLVMStructType

heshanpadmasiri:
> The declare/define stuff is a separate change, and I am not convinced that part is right. Please do that separately.
> 
> In particular, there is a specific API for declaring intrinsics: LLVMGetIntrinsicDeclaration
> 
> There should also be a function to create a struct type following LLVMStructType

Fixed by removing external function and adding functions to create struct types

PR #27, opened 2021-06-07
Serialize SemTypes

### LINKED ISSUE #22: Serialize SemTypes to a pseudo-Ballerina like text format
We should serialize SymType to a pseudo-Ballerina text format with possible addition of `not` operator. 
Since SemTypes have references to TypeCheckContext we can't trivially serialize it. Some type references may be needed.

Prerequisite for #15.

manuranga:
Not completed yet but I'll keep the PR open as a draft
Fixes #22


jclark (changes requested):
You can do something that is closer to real Ballerina.

1. When you have both the _rw and the _ro parts of basic type, then you can use the Ballerina syntax for the whole basic type
2. When you have just the _ro bit of the basic type, you can use readonly&B, where B is the Ballerina syntax for the whole basic type; for the_rw bit you can use !readonly&B
3. You can use correct Ballerina syntax for whole basic types: object is `object {}`, mapping is `map<any|error>`, list is `(any|error)[]`, table is `table<map<any|error>>`
4. When you have everything except error, you can use `any`
5. When you have everything, you can use `any|error`

manuranga:
Yes I am planing to make them more ballerina-ish, I send early just so you can give some preliminary feedback.
Will continue with above changes.

jclark:
I would also suggest separating the part that handles a non-complex semtype into a separate function:

```
function nonComplexSemTypeToString(t:UniformTypeBitSet bits) returns string {
}
```

(I am reluctant to use the word *simple*, since that has a specific meaning in the context of Ballerina types.)

jclark:
We can also add experimental support for  !T in nballerina, so this output would be parseable in our current implementation. This also needs `object { }` as a type descriptor.

manuranga:
Ah yes, I was wondering how to test `!` cases because I am relying on `types.json` for testing. Will be adding `!` to that first.

manuranga:
I updated the code a bit, still need to serialize subtype data.
One issue is, I had to make a bunch of definitions public.

compiler/modules/types/core.bal
```diff
@@ -119,11 +120,11 @@ public readonly class ComplexSemType {
     // For a uniform type with code c,
     // all & (1 << c) is non-zero iff this type contains all of the uniform type
     // some & (1 << c) is non-zero iff this type contains some but not all of the uniform type
-    UniformTypeBitSet all;
-    UniformTypeBitSet some;
+    public UniformTypeBitSet all;
```
jclark, line 123: Making these things public is not OK. The public interface exposed by ComplexSemType should be type-safe

compiler/modules/types/core.bal
```diff
@@ -169,7 +170,7 @@ public readonly class ComplexSemType {
 // There's usually a hardware instruction for this
 // typically called PopCpount
 // This is __builtin_popcount in GCC and clang
-function bitCount(int bits) returns int {
+public function bitCount(int bits) returns int {
```
jclark, line 173: Should not be public

compiler/modules/types/int.bal
```diff
@@ -1,10 +1,10 @@
 // Implementation specific to basic type int.
 
-type IntSubtype readonly & Range[];
+public type IntSubtype readonly & Range[];
```
jclark, line 3: OK for these to be public

compiler/modules/types/string.bal
```diff
@@ -1,6 +1,6 @@
 // Implementation specific to basic type string.
 
-type StringSubtype readonly & record {|
+public type StringSubtype readonly & record {|
```
jclark, line 3: OK for this to be public

jclark (changes requested):
Some of the changes to public are OK, some are not OK.

I don't think `emit` is a good word to use for the module. Maybe `print`: LLVM uses this for conversion to human-readable format.

compiler/modules/types.emit/emit.bal
```diff
@@ -0,0 +1,186 @@
...
+        return nonComplexSemTypeToString(ty);
+    } 
+    else {
+        return complexSemTypeToString(ty);
+    }
+}
+
+function nonComplexSemTypeToString(t:UniformTypeBitSet ty) returns string {
+    // JBUG should work without ether casts 
+    if <int>ty == <int>t:NEVER {
```
jclark, line 29: The casts are ugly here.

I changed types to declare t:NEVER as a UniformTypeBitSet rather than a SemType, so you shouldn't need casts (although jBallerina has some bugs here).

compiler/modules/types.emit/emit.bal
```diff
@@ -0,0 +1,186 @@
...
+        // TODO: escape
+        string[] quoted = from var s in subTy.values
+                          select "\"" + s + "\"";
+        return "|".'join(...quoted);
+    }
+    if subTy is t:IntSubtype {
+        string[] ranges = from t:Range r in subTy
+                          select r.min == r.max ?
+                                 r.min.toString() :
+                                 r.min.toString() + ".." + r.max.toString();
```
jclark, line 140: Better to use `...` here, since that is already Ballerina range syntax.

compiler/modules/types.emit/emit.bal
```diff
@@ -0,0 +1,186 @@
...
+        return "|".'join(...ranges);
+    }
+    if subTy is t:BooleanSubtype {
+        return subTy.value.toString();
+    }
+    // TODO: looks like `subTy is boolean` case is unused figure out why
+    panic err:unimplemented("subtype `boolean` case is not implemented");
+}
+
+function bddToString(bdd:Node node, string[] branch) returns string {
```
jclark, line 150: You can implement this more cleanly. Have a look at e.g. bddEvery.

manuranga:
This is outdated now.

PR #60, opened 2021-06-14
lvm.nback support for named alloca

### LINKED ISSUE #56: llvm.nback should make use of name argument to alloca
These correspond to local variables, so there's an obvious semantic.

Follow on from #55.

heshanpadmasiri:
## Purpose
lvm.nback should make use of name argument to alloca
Resolves #56 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/nback.llvm/llvm.bal
+++ b/compiler/modules/nback.llvm/llvm.bal
@@ -243,8 +243,14 @@ public class FunctionDefn {
         if self.variableNames.hasKey(regName) {
             int? count = self.variableNames.get(regName);
             if count is int {
-                self.variableNames[regName] = count + 1;
-                regName = regName + "." + count.toString();
+                self.variableNames[regName] = count + 1; // increment the count of the base name
+                string newName = regName + "." + count.toString();
+                while (self.variableNames.hasKey(newName)) {
+                    count += 1;
+                    newName = regName + "." + count.toString();
+                }
+                regName = newName;
+                self.variableNames[regName] = 1; // save the augmented name in case user use the same
             } else {
                 err:unreached();
             }
--- a/compiler/modules/nback.llvm/llvm.bal
+++ b/compiler/modules/nback.llvm/llvm.bal
@@ -236,7 +236,7 @@ public class FunctionDefn {
     }
 
     function genReg(string? name = ()) returns string {
-        string regName = "R" + self.variableNames.length().toString();
+        string regName = "_" + self.variableNames.length().toString();
         if name is string {
             regName = name;
         }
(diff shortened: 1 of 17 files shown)


compiler/modules/nback.llvm/llvm.bal
```diff
@@ -179,7 +179,7 @@ public class FunctionDefn {
     final string functionName;
 
     private BasicBlock[] basicBlocks = [];
-    private int varCount = 0;
+    private string[] variableNames = [];
```
jclark: Don't think this the right data structure.

jclark: You should have a map from variable name to count.

compiler/modules/nback.llvm/llvm.bal
```diff
@@ -235,9 +235,18 @@ public class FunctionDefn {
         return label;
     }
 
-    function genReg() returns string {
-        string reg = "%" + "R" + self.varCount.toString();
-        self.varCount += 1;
+    function genReg(string? name = ()) returns string {
```
jclark, line 238: To be completely bullet proof here, you need deal with the case where the name passed to genReg ends  with a `.`  followed by a number

compiler/modules/nback.llvm/llvm.bal
```diff
@@ -235,9 +235,18 @@ public class FunctionDefn {
...
+        string regName = "R" + self.variableNames.length().toString();
```
jclark: If we are now using user-defined names, then `R` is no longer a choice (it looks too much like a user variable). Something like `_` would be better if it works.

compiler/modules/nback.llvm/llvm.bal
```diff
@@ -531,6 +540,19 @@ function createLine(string[] words, string indent = "") returns string {
     return string:concat(indent, ...parts);
 }
 
+function incrementName(string name, int count) returns string{
+    if count == 0 {
+       return name + "." + count.toString(); 
+    } else {
+        int? splitPoint = name.lastIndexOf(".");
```
jclark: In this sort of case, just do `int splitPoint = <int>name.lastIndexOf(".");`

But I don't think you need this with the right data structure.

jclark (changes requested):

compiler/modules/nback.llvm/llvm.bal
heshanpadmasiri: Fixed by changing variable name store to map

compiler/modules/nback.llvm/llvm.bal
heshanpadmasiri: Fixed

compiler/modules/nback.llvm/llvm.bal
heshanpadmasiri: Fixed

compiler/modules/nback.llvm/llvm.bal
heshanpadmasiri: Fixed

Issue #52, opened 2021-06-14
Test cases for subset 1


jclark:
We need a full set of test cases for subset 1, using the format described in https://github.com/ballerina-platform/nballerina/issues/12#issuecomment-859983087 

These test cases must follow the  on the [definition of correctness](https://github.com/ballerina-platform/nballerina/blob/main/docs/correctness.md) and the [definition of subset 1](https://github.com/ballerina-platform/nballerina/blob/main/docs/subset01.md).

If stuff doesn't work yet, we can commit the cases as failing cases.

The semantic E cases will need #39 to work.

jclark:
We need some U tests (which needs #81)

- syntax not implemented (not many tests needed)
- limitations on `io:println`:
   - `io:println(true)` 
   - `io:println()`
   - `io:println(1, 2)`

jclark:
I added the cases for https://github.com/ballerina-platform/nballerina/issues/52#issuecomment-864480589

PR #64, opened 2021-06-14
Add function call related tests

KavinduZoysa:
## Purpose
$subject. This PR will contain test cases related to function calls with the rest of the features in state1.

Related to #52

### DIFF
--- a/tests/Vcall01.bal
+++ b/tests/Vcall01.bal
@@ -1,0 +1,17 @@
+import ballerina/io;
+
+public function main() {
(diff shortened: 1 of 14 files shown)


jclark:
This looks OK.

I would add some tests for deeper nesting like `foo(bar(baz(foobar())))`.

KavinduZoysa:
> This looks OK.
> 
> I would add some tests for deeper nesting like `foo(bar(baz(foobar())))`.

@jclark, I have added few test cases.

Issue #12, opened 2021-06-07
Run tests from annotated ballerina files


manuranga:
**Description:**

Write tests such that it is execute without writing any extra code. Expectation and input should be mentioned in the file itself.
Slimier to LLVM LIT system errors should be mentioned inline.

manuranga:
```ballerina
// @invoke 1, 2
// @expect 3
function test(int x, int y) returns int {
    int diff = x - y;
    return diff;
}
```
how about a format like above.

jclark:
I was thinking of

```
import ballerina/io;

public function main() {
  io:println(add(1, 2)); // @output 3
}

function add(int x, int y) returns int {
   return x + y;
}
```

jclark:
I have used a convention where the name of a valid test case (a program that should compile and run without panicking) starts with the letter `V`; these have `@output` comments specifying the expected output.

  I envisage having other letters for other categories of test case, including

* `P`: this compiles but panics when executed: this would have a `@panic` comment on the line which should panic
* `E`: this should get a compilation error: this would have an `@error` on the line on which an error should be reported; this should only be for cases that are errors according to the full Ballerina spec (XXX need a way to distinguish whether parse error or semantic error is expected)
* `U`: this should be like a `V` test, but is for something that is not in the subset: this should result in an error of type parse or unsupported not not semantic; when a feature is implemented U tests can be renamed to V tests
* `F`: this is for current failing test cases; it should be followed by one of the other letters

jclark:
We should use a DataProvider with testerina to run all the test cases: this should leave `.ll` files for all the `V` cases. These `.ll` cases can separately be tested using the `testll.sh` script.

jclark:
Still need to do #81 and #89, but this is enough for subset01

PR #65, opened 2021-06-14
Compile all tests via testarina

manuranga:
Compile all tests via testarina.

Running following form `tests` dir should run all tests
```
bal test --tests "testCompile*" ../compiler
sh testll.sh
```

Part of #12

### DIFF
--- a/compiler/tests/integration.bal
+++ b/compiler/tests/integration.bal
@@ -1,0 +1,21 @@
+import ballerina/test;
+import ballerina/file;
+
+final string SOURCE_DIR = check file:joinPath("..", "tests");
+
+@test:Config {
+    dataProvider: listPositive
+}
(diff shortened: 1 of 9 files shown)


compiler/tests/integration.bal
```diff
@@ -0,0 +1,21 @@
...
+import ballerina/file;
+
+final string SOURCE_DIR = check file:joinPath("..", "tests");
+
+@test:Config {
+    dataProvider: listPositive
+}
+function testPositive(string path) returns error? {
+    // JBUG due to #31146 let's assert so we see the file name
+    error? err = main(path);
```
jclark: I don't like calling `main` here. Better to factor out the relevant part of main into a `compileFile` function and call that.

jclark (changes requested):

PR #74, opened 2021-06-15
Change test script file permissions to executable

manuranga:
Calling sh file form another (like in `compile.sh`) requires files to be set executable on my machine (`GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin19)`)

```
 manu > bash ./compile.sh
./compile.sh: line 2: ./findJava.sh: Permission denied
```

This will fix it.


jclark:
I had the files checked out on Windows and was using that via WSL, and curiously I was seeing them as executable on the Linux. It looks like there's a special extra step to follow on Windows https://code.adonline.id.au/how-to-change-file-premissions-in-git-on-windows/

Issue #50, opened 2021-06-14
Improve maintainability of parsing tests 


jclark:
The parsing tests should be designed in a way that when the AST data structure is changed, it is only a little work to get the tests working.

jclark:
Some thoughts on how to do this:

- expected output should be in .json files, so they can easily be edited
- the AST should be transformed into a JSON format: the tests should work by comparing the expected and actual results in this format
- the JSON format should not be the same as the AST format; the AST format has design decisions that may change; instead it should be a neutral tuple-based format e.g. an addition can be represented by `["+", L, R]`.
- positions should not be in the JSON format: it is not practical to verify that they are correct by hand; alternative ways should be found to check position information
    - first of all check that the tokenizer is giving the right positions (since tokens are literal strings, this should be easy)
    - then check that the position of each node in the AST corresponds to a token of the right type (i.e. an addition should correspond to a "+" token)

jclark:
Note that with this approach, parsing tests can be shared with jBallerina.

PR #77, opened 2021-06-15
Add AST serialization for function def

manuranga:
Draft of AST json serialization. Opened for preliminary feedback.

Part of #50

### DIFF
--- a/compiler/modules/front/jsonAst.bal
+++ b/compiler/modules/front/jsonAst.bal
@@ -1,0 +1,37 @@
+import wso2/nballerina.err;
+
+function functionDefToJson(FunctionDef func) returns json {
+    return [ "function",
+             func.vis ?: "non-public",
+             func.name,
+             func.typeDesc.args.map(typeDescToJson),
+             typeDescToJson(func.typeDesc.ret),
+             func.body.map(stmtToJson) ];
+}
+
+function stmtToJson(Stmt stmt) returns json {
+    if stmt is VarDeclStmt {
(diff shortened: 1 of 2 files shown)


jclark:
I would suggest using Ballerina keywords as much as possible: so `var` instead of `var-decl`; `return` instead of `ret`.

For function qualifiers like "public": I would use a sorted list of qualifiers. So with subset 1, either `["public"]` or `[]`.

manuranga:
I am planing to add a flag to CLI that will write the ast json to a file, so we can use it from `compile.sh`.

jclark:
CLI flag is good, but why do you want to use it from `compile.sh`?

jclark:
Sorry I pressed the wrong button

manuranga:
I was thinking to keep the ast jsons in `test/ast` dir simile to the ll files (perhaps not for all cases) and verify them using `diff`.

jclark:
Presumably we are going to have parse tests, together with expected output that we have manually written, in the front module; then testerina will check that.

If we have that, I'm not convinced saving the ast in test/ast is going to be of much practical benefit.

jclark:
On second thoughts, perhaps it could be set up in a way that would be useful, e.g. when you run the tests, if one or more ASTs have changed, it will print a message saying "the ASTs for 17 files have changed", and if that's not expected, you can do something about it.  We should probably do the same for `.ll`.

manuranga:
putting this on hold in favor of #84

PR #79, opened 2021-06-16
Add missing tests for codeGen

ushirask:
## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


test/ll/Vwhile05.ll
```diff
@@ -0,0 +1,15 @@
...
+  store i1 0, i1* %_0, align 8
+  ret void
+}
+define internal void @_B_printWhile () {
+  %_0 = alloca i1, align 8
+  br label %L1
+  call void @_Bio__println (i64 0)
+  store i1 0, i1* %_0, align 8
+  ret void
+L1:
```
ushirask: @jclark this is the ll for while false statement. Is this intended as while false is redundant?

compiler/testSuite/Ecall05.bal
```diff
@@ -0,0 +1,8 @@
+public function main() {
+    int x = 1;
+    foo(x); // @error
+}
+
+type foo(
```
jclark: That's not correct `type` definition syntax

compiler/testSuite/Vbooleanorder01.bal
```diff
@@ -0,0 +1,11 @@
+import ballerina/io;
+
+public function main() {
+    printComp(); //@output 1
+}
+
+function printComp() {
+    if (true > false) {
```
jclark: No parentheses here please

compiler/testSuite/Vifelse07.bal
```diff
@@ -0,0 +1,20 @@
+import ballerina/io;
+
```
jclark, line 2: Isn't this already handled by Vifelse06?

compiler/testSuite/Vbooleanorder01.bal
```diff
@@ -0,0 +1,11 @@
+import ballerina/io;
```
jclark, line 1: What does this add to Vbooleanorder00?

compiler/testSuite/Ewhile01.bal
```diff
@@ -0,0 +1,5 @@
+public function main() {
```
jclark: This should be a U case

test/ll/Vwhile05.ll
jclark: Have you tested the output? I don't think that's valid `ll` syntax.

If you find bugs in the code generation, please create a separate issue.

Edited: I just checked: that's no valid `ll` syntax, so it's a bug. No need to report this one.

jclark (changes requested):

compiler/testSuite/Vifelse07.bal
ushirask: directly using the boolean value as if condition than using a stored value takes a different code path so this was added for code coverage

compiler/testSuite/Vbooleanorder01.bal
ushirask: directly comparing the boolean values without assigning is tested here for code coverage

test/ll/Vwhile05.ll
jclark: I fixed the bug for `while false`. You can remove this one.

compiler/testSuite/Uwhile00.bal
```diff
@@ -0,0 +1,5 @@
+public function main() {
```
jclark, line 3: A U case like a V case. With a println @output etc.  When I implement it, I should be able to rename U to V and it should work.

jclark (changes requested):

PR #84, opened 2021-06-16
Add AST serialization

manuranga:
## Purpose
Add string serialization of the AST

## Goals

Main goal
- Do round-trip parse and serialization as a way to verify the parser.

Side goals
- Offer a way to diff parser outputs
- Be a useful tool to see AST while debugging etc.


compiler/modules/front/astString.bal
```diff
@@ -0,0 +1,95 @@
+import wso2/nballerina.err;
+
+type Code [-1|0|1]|string;
```
jclark: * `Code` is not a good name - too easily confused with code as in codeGen.
* Why are you using a 1-tuple here?
* _IN/_OUT would be better as _INDENT/_OUTDENT

I suggest:

```
const LF_INDENT = 1;
const LF_OUTDENT = -1;
const LF = 0;

type Word string|LF_INDENT|LF_OUTDENT|LF;
```

compiler/modules/front/tests/testAstString.bal
```diff
@@ -0,0 +1,12 @@
+import ballerina/test;
+
+@test:Config
+function testFunctionParse() returns error? {
+    string expected = "function foo ( ) {\n" +
```
jclark: Don't want space before semi-colon in the normalization.

jclark (changes requested):
This is fine so far.

I'll merge when it's a fix to #50 i.e. you've done the whole AST and changed the tests to use it.

compiler/modules/front/astString.bal
manuranga: I was considering `Fragment` but decided against it because it's too long. Will use `Word`.
`const` approach looks good for the definition, the downside is we'll need extra casts due to jBal shortcomings, but we shouldn't base our design on that. Will change.

manuranga:
I am planing to have a `parserInputs.txt` with a structure like below
```
@case("number"):
    @canonical
    @expression:
        1
@case("number neg"):
    @canonical
    @expression:
        -1
@case("number max valid"):
    @canonical
    @expression:
        9223372036854775807
@case("number large than max"):
    @expression:
        9223372036854775808
    @error
        "error parsing expression \"9223372036854775808\""
@case("number hex"):
    @unimplemented
    @canonical
    @expression:
        0xf
@case("add number"):
    @canonical
    @expression:
        1 + 1
@case("add with new lines"):
    @expression:
        1
        +
        1
    @expect:
        1 + 1
@case("add no white space"):
    @expression:
        1+1
    @expect:
        1 + 1
@case("if"):
    @statement
    @expression:
        if x {
        }
@case("if nested else if"):
    @statement:
        if x {
        }
        else {
            if {
            }
        }
    @expect:
        if x {
        }
        if else {
        }
@case("function no return"):
    @canonical
    @definition:
        function name() {
        }

```
Will look like below in VSCode
Screenshot 2021-06-17 at 5 29 49 PM

jclark:
That’s just horrible. Inventing some new ad-hoc (ugly) text format is not a good idea at all. Just use one test case per file, and have each file be a Ballerina module. If you really want to put them in one file, then please use a standard format e.g. XML (which you can parse easily with Ballerina).

jclark:
Also please don’t bake particular error messages into the test cases: only distinct error types.

manuranga:
Good thing I posted before fully implementing the parser :sweat_smile:. Did try the file approach before posting, but none of the following options felt right.

1.a) Having a lot of small .bal file pairs each containing an expression or a statement ect; and expected output. I Need to know what kind of entity is in each so I can use the correct parsing funcion, We could use the file name for that. Eg: `expr-int.bal` `expr-int-expected.bal`
1.b) Same as above, but each expression/statement is warped by a fake function so each file becomes a valid ballerina source.
2) Couple of bigger bal files. Here we lose the ability to test `parseExpr` `parseStmt` ect indivitly.

What kind of thing did you have in mind?

jclark:
Either 1a or 1b is fine.

manuranga:
I updated, but still have more to do. Keeping tests in files got a bit out of hand, so I now have them inlined in the test file, if that doesn't look good I can still break them into files. 
Found couple of  failing cases as well.

compiler/modules/front/tests/testParser.bal
```diff
@@ -0,0 +1,91 @@
...
+            ["V", "expr", "0", "0"],
+            ["V", "expr", "1", "1"],
+            ["FE", "expr", "01", ""],
+            ["V", "expr", "9223372036854775807", "9223372036854775807"],
+            ["E", "expr", "9223372036854775808", ""],
+            ["V", "expr", "true", "true"],
+            ["V", "expr", "false", "false"],
+            // unary op
+            ["FV", "expr", "+ 1", "+ 1"],
+            ["FV", "expr", "+1", "+ 1"],
```
jclark: Unary `+` isn't in subset 1, so please comment these out for now. I think the normalized form of `+1` should be `+1` (with no space between the `+` and the `1`.

compiler/modules/front/tests/testParser.bal
```diff
@@ -0,0 +1,91 @@
...
+            ["E", "expr", ")", ""],
+            ["E", "expr", "(1", ""],
+            // statements
+            ["E", "stmt", "", ""],
+            ["E", "stmt", ";", ""],
+            ["V", "stmt", "int i = 10;", "int i = 10;"],
+            ["V", "stmt", "boolean i = 10;", "boolean i = 10;"],
+            ["V", "stmt", "boolean b = false;", "boolean b = false;"],
+            // statements with blocks
+            ["FV", "stmt", "if ok {}", "if ok {\n" +
```
jclark: I think this approach is good for things that fit on a single line, but not for multi-line things. So I suggest a hybrid approach. Use this approach for single-line cases and then make separate files to test multi-line files (I suggest complete Ballerina module-parts, so include the function wrapper).

jclark (changes requested):

jclark:
I made a test an issue for the bug you found https://github.com/ballerina-platform/nballerina/issues/88

jclark (commented):
Please delete existing tests when you have covered everything they do.

manuranga:
@jclark I thinks this is ready to be reviewed now. There are couple of `O` cases that is not covered by this pr. Also need to work on line number/column number validation. Hoping to do that separately.

compiler/modules/front/tests/testParser.bal
```diff
@@ -0,0 +1,238 @@
...
+            continue;
+        }
+        string case = check readCase(caseFile.absPath);
+        string parentDir = check file:parentPath(caseFile.absPath);
+        string canonCase = check file:joinPath(parentDir, canonCaseName(base));
+        string expected = case;
+        if check file:test(canonCase, file:EXISTS) {
+            expected = check readCase(canonCase);
+        }
+        string[] baseParts = regex:split(base, "-");
```
jclark: Please get rid of the regex dependency.

compiler/modules/front/tests/testParser.bal
```diff
@@ -0,0 +1,238 @@
...
+    }
+    return s;
+}
+
+function readCase(string path) returns string|error {
+    string[] lines = check io:fileReadLines(path);
+    string[] caseLines = [];
+    boolean inCase = false;
+    int indented = 0;
+    foreach var l in lines {
```
jclark: I try to avoid using `l` (by itself) as a variable name: too like `I` and `1` in many fonts.

compiler/modules/front/front.bal
```diff
@@ -60,7 +60,9 @@ class Module {
 
 public function loadModule(string filename, bir:ModuleId id) returns bir:Module|err:Any|io:Error {
     string contents = check io:fileReadString(filename);
-    ModulePart part = check parseModulePart(contents);
+    Tokenizer tok = new (contents);
+    check tok.advance();
```
jclark: This change is not OK. This exposes front.bal to Tokenizer, whereas now it's encapsulated in the parser.

jclark (changes requested):
Looks good mostly. I made a few minor points.

What's this @case/@end stuff about?

manuranga:
Cases read from files may have a wrapper function to make the file a valid module. `// @case` and `// @end` marks the subject under the test, rest is ignored.

compiler/modules/front/front.bal
```diff
@@ -60,7 +60,7 @@ class Module {
 
 public function loadModule(string filename, bir:ModuleId id) returns bir:Module|err:Any|io:Error {
     string contents = check io:fileReadString(filename);
-    ModulePart part = check parseModulePart(contents);
```
jclark: This change is not an improvement IMHO

compiler/modules/front/parse.bal
```diff
@@ -2,9 +2,13 @@
 
 import wso2/nballerina.err;
 
-function parseModulePart(string str) returns ModulePart|err:Syntax {
```
jclark: This changes is not an improvement IMHO

compiler/modules/front/astString.bal
```diff
@@ -0,0 +1,217 @@
+import wso2/nballerina.err;
+
+
+// zero-width joiner
+const ZWJ = ();
```
jclark: ZWJ has a well-established meaning as U+200D, which is not what you mean here, so choose another name.

jclark (changes requested):

compiler/modules/front/parse.bal
manuranga: I need a function that parses `ModulePart` given a `Tokenizer`
All the other `parse*` in this file accept a `Tokenizer tok`, So I though it would be more consistent if didn't create a different functions and change the signature of this to match others.

PR #86, opened 2021-06-18
Test operator precedence

ushirask:
## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/testSuite/Vintop02.bal
```diff
@@ -0,0 +1,14 @@
+import ballerina/io;
+
+public function main() {
+  printBoolean(7 == 8); // @output 0
```
ushirask: @jclark this fails with error `error: intersection of operands of operator '==' is empty` but `7 + 1 == 8` does not.

manuranga: In that case this file should have a name starting with with `F`

ushirask: Fixed.

ushirask: > In that case this file should have a name starting with with `F`

Fixed.

jclark (commented):
Please don't include the .ll files

compiler/testSuite/Fintop00.bal
```diff
@@ -0,0 +1,14 @@
+import ballerina/io;
+
+public function main() {
+  printBoolean(7 == 8); // @output 0
```
jclark: This is not legal according to the Ballerina spec. https://ballerina.io/spec/lang/master/#section_6.29

> it is a compile time error if the intersection of the static types of the operands is empty

which in this case it is

jclark (changes requested):
Please leave out the .ll files

ushirask:
> Please leave out the .ll files

Removed.

manuranga:
Did you remove the `F` case you had? It's still good to have that, let's convert that to an `E` case.

ushirask:
> Did you remove the `F` case you had? It's still good to have that, let's convert that to an `E` case.

Added as an error test

compiler/testSuite/Eintop00.bal
```diff
@@ -0,0 +1,14 @@
...
+}
+
+
+function printBoolean(boolean x) {
+  if x {
+    io:println(1);
+  } else {
+    io:println(0);
+  }
+}
```
jclark: Better to end files with a newline. Not having one confuses some tools.

jclark (commented):
You still needs tests for the precedence of unary operators e.g. `-1 + 2` parses as `(-1) + 2`.

`!true == false` parses as `(!true) == false`.

PR #123, opened 2021-06-23
Add tokenizer tests

### LINKED ISSUE #92: More tokenizer tests
We need more tokenizer tests to give complete coverage, particularly for ill-formed cases like unterminated strings, strings crossing newline etc.

The tokenizer tests should also be setup to check line/column information.

manuranga:
Fixes #92
More tests to be added, general framework is ready to be reviewed.

### DIFF
--- a/compiler/modules/front/tests/testParser.bal
+++ b/compiler/modules/front/tests/testParser.bal
@@ -10,7 +10,7 @@ final int CASE_START_LENGTH = CASE_START.length();
 const CASE_END = "// @end";
 
 @test:Config {
-    dataProvider: sourceFragments
+    dataProvider: validTokenSourceFragments
 }
 function testParser(string k, string rule, string subject, string expected) returns err:Syntax|io:Error? {
     if k.includes("U") {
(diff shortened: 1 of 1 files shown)


compiler/modules/front/tests/testParser.bal
```diff
@@ -42,6 +42,63 @@ function testParser(string k, string rule, string subject, string expected) retu
     test:assertEquals(actual, expected, "wrong ast");
 }
 
+@test:Config {
+    dataProvider: sourceFragments
+}
+function testTokenizer(string k, string src) returns error? {
+    int[] lines = findLineFeeds(src);
```
jclark: The tokenizer recognizes `\r`, `\n`, and `\r\n` as line terminators.

compiler/modules/front/tests/testParser.bal
```diff
@@ -42,6 +42,63 @@ function testParser(string k, string rule, string subject, string expected) retu
...
+        else {
+            panic error("tokenizer error for '" + subject + "'", e);
+        }
+    }
+    return tok.current();
+}
+
+function tokenToString(Token t) returns string {
+    if t is VariableLengthToken {
+        return t[1];
```
jclark: This won't work for escapes in strings: in this case `t[1]` is the string after escapes have been processed. In general, you cannot recover the original string from the processed string, because you can have multiple escapes with the same meaning (e.g. `\n` and `\u{A}`). So in this case, all you can do is check the source token is longer than the processed token and that the source token starts with `"`.

compiler/modules/front/tests/testParser.bal
```diff
@@ -66,17 +123,33 @@ function reduceToWords(string rule, string fragment) returns err:Syntax|Word[] {
...
     string[][] s = 
         [["E", "expr", "", ""],
         // literals
          ["V", "expr", "0", "0"],
          ["V", "expr", "1", "1"],
-         ["E", "expr", "01", ""],
          ["V", "expr", "()", "()"],
          ["V", "expr", "-()", "-()"],
          ["V", "expr", "-true", "-true"],
-         ["E", "expr", "\"", ""],
```
jclark, line 79: I suggest you use a string template when the source includes Ballerina strings: otherwise quoting will get confusing.

jclark (changes requested):

compiler/modules/front/tests/testParser.bal
```diff
@@ -42,6 +42,63 @@ function testParser(string k, string rule, string subject, string expected) retu
...
+}
+
+function tokenToString(Token t) returns string {
+    if t is VariableLengthToken {
+        return t[1];
+    }
+    return <string>t;
+}
+
+function findLineFeeds(string str) returns int[] {
```
jclark: To deal with line terminators of different lengths (2 and 1), I suggest you compute the index of the first character of every line, i.e. of the character following all the characters (2 or 1) of the line terminator.

PR #128, opened 2021-06-25
Add support for global variables in nback.llvm

### LINKED ISSUE #127: Add support for global variables in nback.llvm
This is to support #91.

@jclark for this I am thinking of adding a new function `getOrInsertGlobal(string name, Type ty) returns Value` similar to [`getOrInsertGlobal`](https://llvm.org/doxygen/classllvm_1_1Module.html#a1ec20a6f23ced1a328d4b1223fa22d96) in C api

heshanpadmasiri:
## Purpose
Add support for global variables in nback.llvm
Resolves #127.

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/nback.llvm/llvm.bal
+++ b/compiler/modules/nback.llvm/llvm.bal
@@ -107,10 +107,7 @@ public type IntrinsicFunctionName IntegerArithmeticIntrinsicName|GeneralIntrinsi
 
 # Corresponds to llvm::Module class
 public class Module {
-    private final FunctionDefn[] functions = [];
-    private final map<FunctionDecl> intrinsicFunctions = {};
-    private final map<PointerValue> globalVariables = {};
-    private final FunctionDecl[] functionDecls = [];
+    private final map<FunctionDefn|FunctionDecl|PointerValue> declarations = {};
     private final Context context;
 
     function init(Context context) {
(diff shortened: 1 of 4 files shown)


compiler/modules/nback.llvm/llvm.bal
```diff
@@ -169,6 +170,25 @@ public class Module {
         }
     }
 
+    // Corresponds to LLVMAddGlobal
+    public function addGlobal(Type ty, string name) {
```
jclark: Should return PointerValue, like addFunctionDefn/Decl return a value.

C API returns a ValueRef.

compiler/modules/nback.llvm/llvm.bal
```diff
@@ -169,6 +170,25 @@ public class Module {
...
+        if self.globalVariables.hasKey(name) {
+            panic error(createLine(["Variable", name, "already", "declared"]));
```
jclark: Does the C API panic in this case?

Our existing code for addFunctionDefn/Decl doesn't deal with adding something with the same name. We should be consistent here.

Also we should not allow a global function and global variable with the same name.

Probably need a combined `map<FunctionDefn|FunctionDecl|GlobalVariable>`.

jclark (changes requested):

compiler/modules/nback.llvm/llvm.bal
heshanpadmasiri: C api don't panic instead it appends to the name (for example if we add two variables named `g1` one will be `g1` other `g1.1`). But then when you need to get the second variable user must use the appended name (`g1.1`) otherwise (`g1`) they will get the first variable. I thought this is bit confusing to the user because they use one name to create variables and another to access them

compiler/modules/nback.llvm/llvm.bal
heshanpadmasiri: Fixed

compiler/modules/nback.llvm/llvm.bal
heshanpadmasiri: Same is true for functions as well but we don't provide get methods for functions so the user must be using the function object returned at the time of creation

compiler/modules/nback.llvm/llvm.bal
```diff
@@ -169,6 +170,27 @@ public class Module {
...
+        return val;
+    }
+
+    // Corresponds to LLVMGetNamedGlobal
+    public function getNamedGlobal(string name) returns PointerValue {
+        PointerValue? val = self.globalVariables[name];
+        if val is PointerValue {
+            return val;
+        } else {
+            panic error(createLine(["No", "global", "variable", "by", "name", name]));
```
jclark: Use
```
err:format(`No global variable named ${name}`)
```

instead of `createLine` here.

compiler/modules/nback.llvm/llvm.bal
jclark: How is your code handling name clashes between global variables and functions?

I agree it's weird to add a function with one name, and then require a different name to retrieve it. But we need to be consistent in our treatment of functions, global variables and local variables. Seems to me easiest way is to eliminate `getNamedGlobal`.

jclark (changes requested):

heshanpadmasiri:
Changed the way in which module store functions and global variables such that unique names are now enforced among both functions and global variables. Made intrinsic function names reserved so users can't declare functions and variables using same names. Though this is more restrictive than the C API I thought this will make generated code more readable and make `llvm.bal` simpler

PR #137, opened 2021-06-25
Add ability to set target triple in nback.llvm

### LINKED ISSUE #9: Generate module target triple
At the moment, compiling the .ll file generated by the compiler produces the warning:

```
warning: overriding the module target triple with x86_64-pc-linux-gnu [-Woverride-module]
```

We need to generate the appropriate target triple to deal with this.

This should be a configurable value. Eventually, it should be possible to override with command-line option.

heshanpadmasiri:
## Purpose
Add ability to set the target triple in nback.llvm
Add configurable variable `target` in `nback.bal` to set the taget triple (default to `x86_64-pc-linux-gnu`)
Resolves #9

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/nback/nback.bal
+++ b/compiler/modules/nback/nback.bal
@@ -9,10 +9,11 @@ public const OUTPUT_EXTENSION = ".ll";
 configurable string target = "x86_64-pc-linux-gnu";
 
 // If outputFilename is non-nil write the output to this file.
-public function compileModule(bir:Module mod, string? outputFilename) returns err:Any|io:Error? {
+public function compileModule(bir:Module mod, string? outputFilename) returns err:Any|io:Error|error? {
     llvm:Context context = new;
     llvm:Module llMod = check buildModule(mod, context);
-    llMod.setTarget(<llvm:TargetTriple>target);
+    llvm:TargetTriple targetTriple = check llvm:stringToTargetTriple(target);
+    llMod.setTarget(targetTriple);
     if outputFilename != () {
         return llMod.writeFile(outputFilename);
     }
--- a/compiler/modules/nback/nback.bal
+++ b/compiler/modules/nback/nback.bal
@@ -7,14 +7,23 @@ import wso2/nballerina.nback.llvm;
 # The preferred output extension for the output filename.
 public const OUTPUT_EXTENSION = ".ll";
 configurable string target = "x86_64-pc-linux-gnu";
+public type ConfigError distinct error;
 
 // If outputFilename is non-nil write the output to this file.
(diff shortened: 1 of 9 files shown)


compiler/modules/nback/nback.bal
```diff
@@ -6,11 +6,13 @@ import wso2/nballerina.nback.llvm;
 
 # The preferred output extension for the output filename.
 public const OUTPUT_EXTENSION = ".ll";
+configurable string target = "x86_64-pc-linux-gnu";
 
 // If outputFilename is non-nil write the output to this file.
 public function compileModule(bir:Module mod, string? outputFilename) returns err:Any|io:Error? {
     llvm:Context context = new;
     llvm:Module llMod = check buildModule(mod, context);
+    llMod.setTarget(<llvm:TargetTriple>target);
```
jclark: This isn't OK because the user will get a panic if they've set an unacceptable value for target.

Two possible approaches:
1. Make the type of `target` be `llvm:TargetTriple`: the configurable implementation should take care of the reporting an error if the user supplies an invalid string. If the configurable implementation does not handle this, then please file an issue with nBal needs.
2. If (1) does not work than have two variables: a configurable variable of type string and another variable of type llvm:TargetTriple with an initializer that uses `check` to initialize from the configurable variable.

jclark (changes requested):

compiler/modules/nback/nback.bal
```diff
@@ -6,11 +6,14 @@ import wso2/nballerina.nback.llvm;
 
 # The preferred output extension for the output filename.
 public const OUTPUT_EXTENSION = ".ll";
+configurable string target = "x86_64-pc-linux-gnu";
 
 // If outputFilename is non-nil write the output to this file.
-public function compileModule(bir:Module mod, string? outputFilename) returns err:Any|io:Error? {
+public function compileModule(bir:Module mod, string? outputFilename) returns err:Any|io:Error|error? {
```
jclark: This should use a distinct error type. In nback.bal, add


public type ConfigError distinct error;

jclark (changes requested):

compiler/modules/nback/nback.bal
heshanpadmasiri: I created a issue for 1 (link:https://github.com/ballerina-platform/ballerina-lang/issues/31447) but I don't have permission to set labels for issues.

compiler/modules/nback/nback.bal
heshanpadmasiri: Fixed

PR #146, opened 2021-06-27
Implement binary bitwise expressions

### LINKED ISSUE #96: Implement binary bitwise expressions
As specified here: https://ballerina.io/spec/lang/2020R1/#section_6.34

manuranga:
Fixes #96


jclark (changes requested):
This is great!

You are missing a couple of minor things:

- you need some E tests to make sure that values of type other than `int` are not being accepted
- to make those tests pass you need to modify bir/verify.bal (just add to IntBinaryInsn)

I would also like to include xor in the same subset (which I think needs an addition to nback.llvm)

compiler/modules/front/codeGen.bal
```diff
@@ -322,6 +322,14 @@ function codeGenExpr(CodeGenContext cx, bir:BasicBlock bb, Scope? scope, Expr ex
             bb.insns.push(insn);
             return [result, nextBlock];
         }
+        var {bitwiseOp: op, left, right} => {
```
jclark: I like to add a space after `{` and before `}` when braces are used in a single line.

PR #171, opened 2021-07-02
Implement shift expressions

### LINKED ISSUE #103: Implement shift expressions
Specified in:

https://ballerina.io/spec/lang/2021R1/#shift-expr

Note the point about "all except the bottom 6 bits of the shift amount are masked out (as if by x & 0x3F)". This should be done by the backend.

manuranga:
Fixes #103


manuranga:
Work in progress, currently on hold.

compiler/modules/front/token.bal
```diff
@@ -250,77 +282,81 @@ class Tokenizer {
                 return [DECIMAL_NUMBER, digits];
             }
             else if ch == "\"" {
-                string content = "";
-                while true {
+                return self.tokenizeStr();
```
manuranga, line 285: Had to reduce the size of this method due to a known bug in jBallerina  ballerina-platform/ballerina-lang#29137.
Github diff is confusing, you may append `&w=1` to the URL to ignore white space changes in UI.

compiler/modules/front/token.bal
```diff
@@ -80,6 +80,10 @@ final readonly & map<MultiCharDelim> WITH_EQUALS = {
     ">": ">="
 };
 
+const MODE_NORMAL = 1;
+const MODE_TYPE = 2;
```
jclark: I think `MODE_TYPE_DESC` would be clearer than `MODE_TYPE`.

I would make `MODE_NORMAL` be 0, rather than 1.

compiler/modules/front/token.bal
jclark: Diffs are pretty unintelligible but the function needs splitting up anyway.

jclark (changes requested):
Looks pretty good.

I'm starting to feel the need for hex int literals #235

Issue #141, opened 2021-06-26
Implement `foreach` with `..<`


jclark:
Implement a subset of `foreach` that uses `..<` e.g.

```
foreach int i in 0 ..< v.length() {
   //...
}
```
This can turn into the same as a `while`, but for incrementing the loop variable, we can use an addition without the need to check for overflow.

This is useful with #130 or #135

jclark:
According to https://llvm.org/docs/Frontend/PerformanceTips.html#restricted-operation-semantics we should also use `nsw` on the addition.

jclark:
We really need three different varieties of add:

1. may panic on overflow
2. guaranteed by compiler not to overflow
3. wrap on overflow

The NO_PANIC BIR instructions should be type 2 and should generate LLVM IR with `nsw`. We can have langlib functions to explicitly generate type 3.

manuranga:
With addition of `nsw` (and other optimized instructions) it'll be useful to incorporate `opt -passes=poison-checking`. Otherwise tests can pass due to undefined behavior.

jclark:
@manuranga Is there a clang argument to do that?

manuranga:
> Is there a clang argument to do that?

Couldn't find a way to directly do it. Running though `opt -S` command and then passing the output to `clang` works.

manuranga:
What would be the best way to code gen foreach:

## 1) Desuger and call codeGenStmts
```ballerina
function codeGenForeachStmt(⋯, ForeachStmt stmt) returns CodeGenError|bir:BasicBlock? {
    Stmt[] desugared = check desugerForeachStmt(cx, stmt);
    return check codeGenStmts(⋯, desugared);
}
```
### Pro:
We could formalize desguring pattern and have a set of unit tests, similar to how LLVM tests passes.

### Con:
Need new AST elements to represent non-panicking add in addition to new bir instruction
Need check how to deal with scope leak

## 2) Desuger and codeGen each
```ballerina
function codeGenRangedForeachStmt(⋯, ForeachStmt stmt) returns CodeGenError|bir:BasicBlock? {
    VarDeclStmt varDecl = {⋯, stmt.varName};
    var [curBlock, inner] = check codeGenVarDeclStmt(⋯, varDecl);
    ⋯
    WhileStmt loop = { condition, body };
    return codeGenWhileStmt(⋯, loop);
}
```
### Pro:
More control over scopes ect.

## 3) Directly code gen
### Con:
Foreach and while will have slimier code

[...]

PR #172, opened 2021-07-03
Parse range expression

manuranga:
Parse range expression

Only intended to be used with `foreach`
Part of #141


jclark:
There seems to be a misunderstanding here: subset03 does not have a range expressions. Rather the syntax for range expressions is folded into the foreach statement:

```
foreach-stmt = "foreach" typedesc identifier "in" expression "..<" expression stmt-block
```

(There's actually a spec bug there: it should be `additive-expr` not expressions.)

Your code is parsing lots of things that is not in a subset. The goal is to parse *exactly* the subset, so that we not have to waste time later rejecting things not in the subset.

So we can have a RangeExpr and a parseRangeExpr but they should be only parsed from/referenced from the foreach-stmt, not as part of expressions generally.

manuranga:
Hmm I though it's OK since I am giving an error at CodeGen, let me restrict it from the parser.

jclark:
Syntax errors need to be caught during parsing for distinct error types to work right.

PR #179, opened 2021-07-06
Implement `foreach` with `..< `

### LINKED ISSUE #141: Implement `foreach` with `..<`

manuranga:
Fixes  #141


manuranga:
Integer addition optimization is not handled here and moved to #180.

compiler/modules/front/ast.bal
```diff
@@ -28,9 +28,9 @@ type FunctionDef record {|
     bir:FunctionSignature? signature = ();
 |};
 
-type Stmt VarDeclStmt|AssignStmt|CallStmt|ReturnStmt|IfElseStmt|WhileStmt|BreakStmt|ContinueStmt;
+type Stmt VarDeclStmt|AssignStmt|CallStmt|ReturnStmt|IfElseStmt|WhileStmt|ForeachStmt|BreakStmt|ContinueStmt;
 type CallStmt FunctionCallExpr|MethodCallExpr;
-type Expr SimpleConstExpr|BinaryExpr|UnaryExpr|FunctionCallExpr|MethodCallExpr|VarRefExpr|TypeCastExpr|ConstructorExpr|MemberAccessExpr;
+type Expr SimpleConstExpr|BinaryExpr|UnaryExpr|FunctionCallExpr|MethodCallExpr|VarRefExpr|TypeCastExpr|ConstructorExpr|MemberAccessExpr|RangeExpr;
```
jclark, line 33: In this subset , we don't allow an Expr to be a `RangeExpr`, so this should not be here.

compiler/modules/front/ast.bal
```diff
@@ -57,6 +57,15 @@ type WhileStmt record {|
     Stmt[] body;
 |};
 
+type ForeachStmt record {|
+    InlineTypeDesc td;
```
jclark, line 61: Don't need this: we only accept int.

compiler/modules/front/ast.bal
```diff
@@ -148,6 +158,12 @@ type MemberAccessLExpr record {|
     err:Position pos;
 |};
 
+type RangeExpr record {|
+    Expr lower;
+    Expr upper;
+    RangeOp op;
```
jclark, line 164: Don't need this: we only have "..<".

compiler/modules/front/astString.bal
```diff
@@ -89,6 +89,14 @@ function stmtToWords(Word[] w, Stmt stmt) {
         exprToWords(w, stmt.condition);
         blockToWords(w, stmt.body);
     }
+    else if stmt is ForeachStmt{
```
jclark, line 92: Missing space

compiler/modules/front/codeGen.bal
```diff
@@ -159,6 +162,59 @@ function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Scope? initialScope,
     return curBlock;
 }
 
+final readonly & map<bir:OrderOp> RANGE_TO_ORDER = {
```
jclark, line 165: Don't need this: `...` is not in the subset.

compiler/modules/front/tests/testParser.bal
```diff
@@ -250,6 +250,8 @@ function validTokenSourceFragments() returns string[][]|error {
          ["V", "expr", "a | b", "a | b"],
          ["V", "expr", "a & b", "a & b"],
          ["V", "expr", "a ^ b", "a ^ b"],
+         ["U", "expr", "a ..< b", "a ..< b"],
+         ["U", "expr", "a ... b", "a ... b"],
```
jclark, line 254: Not in the subset.

compiler/modules/front/parseStmt.bal
```diff
@@ -197,3 +201,17 @@ function parseWhileStmt(Tokenizer tok) returns WhileStmt|err:Syntax {
     Stmt[] body = check parseStmtBlock(tok);
     return { condition, body };
 }
+
+function parseForeachStmt(Tokenizer tok) returns ForeachStmt|err:Syntax {
+    InlineTypeDesc td = check parseInlineTypeDesc(tok);
```
jclark, line 206: The subset just accepts `int` here.

compiler/modules/front/codeGen.bal
```diff
@@ -159,6 +162,59 @@ function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Scope? initialScope,
...
+    "..<": "<"
+};
+
+function codeGenForeachStmt(CodeGenContext cx, bir:BasicBlock startBlock, Scope? scope, ForeachStmt stmt) returns CodeGenError|bir:BasicBlock? {
+    Expr range = stmt.iterable;
+    var { varName, semType} = stmt;
+    if lookup(varName, scope) !== () {
+        return cx.semanticErr(`duplicate declaration of ${varName}`);
+    }
+    if !(range is RangeExpr) {
```
jclark, line 176: The ForeachStmt should have a range member: then you don't need this extra level of indentation.

compiler/modules/front/codeGen.bal
```diff
@@ -474,6 +530,10 @@ function codeGenExpr(CodeGenContext cx, bir:BasicBlock bb, Scope? scope, Expr ex
                 return cx.semanticErr("cannot apply member access to constant of simple type");
             }
         }
+        // Range
+        var {upper, lower, op} => {
```
jclark, line 534: We don't need this if we don't include ranges in expressions.

compiler/modules/front/codeGen.bal
```diff
@@ -607,7 +667,7 @@ function mustLookup(CodeGenContext cx, string name, Scope? scope, boolean forAss
         return cx.semanticErr(`variable ${name} not found`);
     }
     else if forAssign && binding.isFinal {
-        return cx.semanticErr(`cannot assign to parameter ${name}`);
+        return cx.semanticErr(`cannot assign to final variable ${name}`);
```
jclark, line 670: How does this deal with the parameter case?

jclark (commented):
You have done a bunch of things that make this more complicated than is needed for this subset.

I will fix them.

compiler/modules/front/codeGen.bal
manuranga: Both assignment to loop var and assignment to parameter come here. Currently way to distinguish.

Issue #202, opened 2021-07-11
llvm.print should not generate a space before `(`


jclark:
At the moment, llvm.print is generating a space before `(`, which is unnecessary and looks ugly. The examples on the LLVM ref page don't do it.

jclark:
Fixed in https://github.com/ballerina-platform/nballerina/commit/ee5cd553ec5f825510ab156f597501d50b3e1711

PR #208, opened 2021-07-12
Use LLVM unnamed local identifiers

### LINKED ISSUE #153: Use LLVM unnamed local identifiers
At the moment, llvm.nback uses a named LLVM local identifier starting with `_`  (e.g. `%_17`) when the API user does not specify a `name` argument. It would be better for it to use an unnamed LLVM identifier (e.g. `%17`). This would improve the similarity of code generated using llvm.nback and using #19.

This isn't as simple as just leaving out the `_`: LLVM has a specify method of numbering basic blocks and variables that the use of unnamed identifiers must comply with.

heshanpadmasiri:
## Purpose
Make output from `print.llvm` more similar to `jni.llvm`
Resolves #153, #202

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/print.llvm/llvm.bal
```diff
@@ -314,20 +323,74 @@ public class FunctionDefn {
...
+            return true;
+        }
+        return false;
+    }
 
-    public function addEnumAttribute(EnumAttribute attribute){
+    public function addEnumAttribute(EnumAttribute attribute) {
         if self.attributes.indexOf(attribute) == () {
-           self.attributes.push(attribute); 
+            self.attributes.push(attribute);
```
jclark, line 362: Extra space

compiler/modules/print.llvm/llvm.bal
```diff
@@ -314,20 +323,74 @@ public class FunctionDefn {
...
+    // Verify name is a valid LLVM name otherwise must escape
+    function nameValid(string name) returns boolean{
+        string nameLower = name.toLowerAscii();
+        // Currently check only the first character
+        string firstChar = nameLower.substring(0,1);
+        if firstChar == "_" || firstChar == "$" {
+            return true;
+        }
+        int first = firstChar.getCodePoint(0);
+        if firstChar >= "a" && firstChar <= "z" {
```
jclark: This shouldn't compile. You can't compare an `int` and a `string`.

compiler/modules/print.llvm/llvm.bal
```diff
@@ -314,20 +323,74 @@ public class FunctionDefn {
...
+        if name is string && !self.nameValid(regName) {
+            regName = "\"" + regName + "\"";
+        }
         string reg = "%" + regName;
         return reg;
-    } 
+    }
+
+    // Verify name is a valid LLVM name otherwise must escape
+    function nameValid(string name) returns boolean{
```
jclark: missing space before `{`

compiler/modules/print.llvm/llvm.bal
```diff
@@ -314,20 +323,74 @@ public class FunctionDefn {
         else {
             self.variableNames[regName] = 1;
         }
+        if name is string && !self.nameValid(regName) {
```
jclark: Better to have a function `escapeString` that properly escapes a string using LLVM rules. It doesn't have to handle everything. But it should panic on everything it does not handle.

compiler/modules/print.llvm/llvm.bal
jclark: This whole function needs rewriting.

compiler/modules/print.llvm/llvm.bal
```diff
@@ -343,38 +406,38 @@ public class Builder {
...
 
     // Corresponds to LLVMBuildStore
-    public function store(Value val, PointerValue ptr, Alignment? align = ()) {
+    public function store(Value val, PointerValue ptr, Alignment? align=()) {
         Type ty = ptr.ty.pointsTo;
         if ty != val.ty {
             panic error("store type mismatch: " + typeToString(val.ty) + ", " + typeToString(ptr.ty));
         }
-        addInsnWithAlign(self.bb(), ["store", typeToString(ty), val.operand, ",", typeToString(ptr.ty), ptr.operand], align);
+        addInsnWithAlign(self.bb(), [createLine(["store", typeToString(ty)]), val.operand, createLine([",", typeToString(ptr.ty)]), ptr.operand], align);
```
jclark: Adding all these calls to `createLine` is not a good solution to the problem.

jclark (changes requested):
I think we can do this much better by using Ballerina in a more idiomatic way, and leveraging unions.  We are not restricted to having lists of strings in Ballerina.  So we can use something other than a string to represent an unnamed variable that we can choose a number for later. I think you can probably do it by representing an unnamed variable as an `int`.  Instead of words being a `string[]`, it's a `(int|string)[]`, where there an int represents an unnamed variable that you are going to need to renumber later. Then the lines are represented as an  `(int|string)[]` also, where the strings are maximal chunks that are concatenated with spaces. Then at the end you build an array of strings that maps from the `int` to the final `%N` that you need to use, and use that to map from an ` (int|string)[]`  to `string[]` which you then concatenate.

Does this make sense?

compiler/modules/print.llvm/llvm.bal
heshanpadmasiri: In the older one there was a space missing (only 3 instead of 4) there

compiler/modules/print.llvm/llvm.bal
heshanpadmasiri: Function replace with new `escapeName` function

compiler/modules/print.llvm/llvm.bal
heshanpadmasiri: Function replace with new `escapeName` function

compiler/modules/print.llvm/llvm.bal
heshanpadmasiri: `escapeName` function added

compiler/modules/print.llvm/llvm.bal
heshanpadmasiri: Function replace with new `escapeName` function

compiler/modules/print.llvm/llvm.bal
heshanpadmasiri: Replaced with new function `concat` that properly handle merging `string` chunks and removed changes to `createLine` function

heshanpadmasiri:
> I think we can do this much better by using Ballerina in a more idiomatic way, and leveraging unions. We are not restricted to having lists of strings in Ballerina. So we can use something other than a string to represent an unnamed variable that we can choose a number for later. I think you can probably do it by representing an unnamed variable as an `int`. Instead of words being a `string[]`, it's a `(int|string)[]`, where there an int represents an unnamed variable that you are going to need to renumber later. Then the lines are represented as an `(int|string)[]` also, where the strings are maximal chunks that are concatenated with spaces. Then at the end you build an array of strings that maps from the `int` to the final `%N` that you need to use, and use that to map from an ` (int|string)[]` to `string[]` which you then concatenate.
> 
> Does this make sense?

Fixed such that positive numbers represent "variables" and negative numbers represent basic blocks

manuranga:
Using negative numbers don't feel right to me. Is it possible to using something like `([Kind, int]|string)[]`

heshanpadmasiri:
> Using negative numbers don't feel right to me. Is it possible to using something like `([Kind, int]|string)[]`

Fixed by adding new type `UnnamedIdentifier` instead of int

compiler/modules/print.llvm/llvm.bal
```diff
@@ -287,47 +297,124 @@ public class FunctionDefn {
...
-            }
-            regName = newName;
-            self.variableNames[regName] = 1; // save the augmented name in case user use the same  
-        }
-        else {
-            self.variableNames[regName] = 1;
+    }
+
+    function escapeName(string name) returns string {
+        if regex:matches(name, "[-a-zA-Z$._][-a-zA-Z$._0-9]*") {
```
jclark: Please don't use `regex`: we are avoiding a dependency on that.

compiler/modules/print.llvm/llvm.bal
```diff
@@ -287,47 +297,124 @@ public class FunctionDefn {
...
+            return name;
+        } else {
+            return "\"" + name + "\"";
```
jclark: Some characters will need quoting with `\`

compiler/modules/print.llvm/llvm.bal
```diff
@@ -240,14 +241,18 @@ public class FunctionDefn {
     private Value[] paramValues;
     private Linkage linkage = "external";
     private final Context context;
+    private map<string> nameTranslation = {};
```
jclark: Can't you make this a `string[]` indexed by the `int` you stored to represent the unnamed identifier?

jclark: If you need to distinguish the case of a basic block label vs a operand, you can use another parallel array for that.

jclark (changes requested):

compiler/modules/print.llvm/llvm.bal
```diff
@@ -772,8 +876,26 @@ function createLine(string[] words, string indent = "") returns string {
     return string:concat(indent, ...parts);
 }
 
+function concat(string... words) returns string {
```
jclark, line 888: This has duplicate functionality with createLine

jclark:
> Using negative numbers don't feel right to me. Is it possible to using something like `([Kind, int]|string)[]`

You don't need to do that. When you allocate an integer to represent an unnamed block or an operand, you can store in an array a flag saying which it was (block vs operand).

compiler/modules/print.llvm/llvm.bal
```diff
@@ -7,12 +7,13 @@
 
 import nballerina.err;
 import ballerina/io;
+import ballerina/regex;
 
 # Corresponds to LLVMValueRef 
 public readonly distinct class Value {
-    string operand;
+    string|int operand;
```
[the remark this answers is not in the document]

compiler/modules/print.llvm/llvm.bal
```diff
@@ -343,38 +430,38 @@ public class Builder {
...
 
     // Corresponds to LLVMBuildStore
-    public function store(Value val, PointerValue ptr, Alignment? align = ()) {
+    public function store(Value val, PointerValue ptr, Alignment? align=()) {
         Type ty = ptr.ty.pointsTo;
         if ty != val.ty {
             panic err:illegalArgument("store type mismatch: " + typeToString(val.ty) + ", " + typeToString(ptr.ty));
         }
-        addInsnWithAlign(self.bb(), ["store", typeToString(ty), val.operand, ",", typeToString(ptr.ty), ptr.operand], align);
+        addInsnWithAlign(self.bb(), [concat("store", typeToString(ty)), val.operand, concat(",", typeToString(ptr.ty)), ptr.operand], align);
```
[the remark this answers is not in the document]

compiler/modules/print.llvm/llvm.bal
```diff
@@ -343,38 +430,38 @@ public class Builder {
...
     }
 
     // Corresponds to LLVMBuildNSW{Add,Mul,Sub}
     public function iArithmeticNoWrap(IntArithmeticOp op, Value lhs, Value rhs, string? name=()) returns Value {
         BasicBlock bb = self.bb();
-        string reg = bb.func.genReg();
+        string|int reg = bb.func.genReg();
```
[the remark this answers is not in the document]

compiler/modules/print.llvm/llvm.bal

compiler/modules/print.llvm/llvm.bal
heshanpadmasiri: Fixed. We are wrapping names with `"` if it doesn't match the given pattern (similar to what `jni.llvm` does) only character that I think we have to escape is `"` it self.

compiler/modules/print.llvm/llvm.bal

compiler/modules/print.llvm/llvm.bal
heshanpadmasiri: Fixed

jclark (approved):
This looks great.

jclark:
Unfortunately when I run `make test`, it fails

llvm-link-11: out/ll/Vforeach03.ll:50:3: error: instruction expected to be numbered '%16'
  %15 = load i64, i64* %i
  ^

heshanpadmasiri:
> Unfortunately when I run `make test`, it fails
> 
> llvm-link-11: out/ll/Vforeach03.ll:50:3: error: instruction expected to be numbered '%16'
> %15 = load i64, i64* %i
> ^

@jclark I have made a new PR #219 that fixes this issue. It seems to be caused due to having an  un-referenced basic block in that test case. This PR adds a work around to temporarily make it output a valid code in such cases. `jni` generated ll file (which marks predecessors of each basic blocks) for reference
[Vforeach03.txt](https://github.com/ballerina-platform/nballerina/files/6809529/Vforeach03.txt)

PR #209, opened 2021-07-12
Migrate to map based DataProvider 

manuranga:
Migrate to map based DataProvider.

Since now I need maps instead of lists for data providers, I had to get rid of the query expressions. Is there something more elegant than a foreach loop for this?

Please be aware some tests can't be run in isolation after this due to a bug https://github.com/ballerina-platform/ballerina-lang/issues/31672

### DIFF
--- a/compiler/modules/front/tests/testParser.bal
+++ b/compiler/modules/front/tests/testParser.bal
@@ -9,6 +9,8 @@ const CASE_START = "// @case";
 final int CASE_START_LENGTH = CASE_START.length();
 const CASE_END = "// @end";
 
+// JBUG #31673 can't specify the first type to be "V"|"E"|ect
+type ParserTestCase [string, string, string, string];
 @test:Config {
     dataProvider: validTokenSourceFragments
 }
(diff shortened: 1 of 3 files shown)


compiler/modules/front/tests/testParser.bal
```diff
@@ -341,7 +346,6 @@ function validTokenSourceFragments() returns string[][]|error {
          ["E", "expr", "(f()())", ""],
          ["V", "expr", "(((x)))", "x"],
          // statement
-         ["E", "stmt", "", ""],
```
jclark, line 344: Why are we removing this?

compiler/modules/front/tests/testParser.bal
```diff
@@ -209,9 +216,6 @@ function validTokenSourceFragments() returns string[][]|error {
          ["UV", "expr", string`"\""`, string`"\""`],
          ["UV", "expr", string`"what"`, string`"what"`],
          ["UV", "expr", string`"Say \"what\" again."`, string`"Say \"what\" again."`],
-         // ref
-         ["V", "expr", "x", "x"],
```
jclark, line 213: Why are we removing this?

jclark (changes requested):

compiler/modules/front/tests/testParser.bal
manuranga: 2 tests where being repeated. Noticed because now the test-cases get de-duplicated due to map.

compiler/modules/front/tests/testParser.bal
manuranga: explained in the previous comment.

compiler/tests/baltest.bal
```diff
@@ -2,26 +2,33 @@ import ballerina/test;
...
-    return from var entry in check file:readDir("tests/data")
-           let string path = entry.absPath
-           where path.endsWith(".bal")
-           let string[]|error res = readTestCase(path)
-           // skip files with no subtype lines on them
-           where res is string[] && res[1].length() > 0
-           select res;
+function subtypeCaseProvider() returns map<SubtypeTestCase>|error {
+    map<SubtypeTestCase> cases = {};
+    foreach var file in check file:readDir("tests/data") {
```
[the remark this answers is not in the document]
manuranga: Is there a way to get a `map` from `from`? I tried to see if I can get a `table` and trivially convert it to a `map`, but there isn't a way.

Or should I get a list and then convert?

compiler/modules/front/tests/testParser.bal
jclark: Oh I see. OK.

compiler/tests/baltest.bal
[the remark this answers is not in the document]
jclark: Use a do-clause to build the map: https://ballerina.io/spec/lang/2021R1/#do-clause

Issue #190, opened 2021-07-08
Remove workarounds no longer needed with slbeta2


jclark:
The code has a lot of ugly workarounds for jBallerina bugs.

We should get rid of as many as possible as we can now that slbeta2 is out.

PR #210, opened 2021-07-12
Remove couple of JBUG comments

manuranga:
Part of #190

### DIFF
--- a/compiler/modules/nback/build.bal
+++ b/compiler/modules/nback/build.bal
@@ -163,11 +163,7 @@ class Scaffold {
         self.functionDefns = functions;
         self.importedFunctions = importedFunctions;
         self.birBlocks = code.blocks;
-        // JBUG 31008 if this is a query expression
-        final Repr[] reprs = [];
-        foreach var reg in code.registers {
-            reprs.push(check semTypeRepr(reg.semType));
-        }
+        final Repr[] reprs = from var reg in code.registers select check semTypeRepr(reg.semType);
         self.reprs = reprs;
         self.retRepr = check semTypeRetRepr(defn.signature.returnType);
         self.nParams = defn.signature.paramTypes.length();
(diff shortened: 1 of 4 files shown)


compiler/modules/nback/build.bal
```diff
@@ -379,7 +374,6 @@ function buildAssign(llvm:Builder builder, Scaffold scaffold, bir:AssignInsn ins
 function buildCall(llvm:Builder builder, Scaffold scaffold, bir:CallInsn insn) returns BuildError? {
     // Handler indirect calls later
     bir:FunctionRef funcRef = <bir:FunctionRef>insn.func;
-    // JBUG #31008 cannot write this with select
```
jclark, line 382: But you have not rewritten this one with `select`

compiler/modules/nback/build.bal
```diff
@@ -163,12 +163,7 @@ class Scaffold {
         self.functionDefns = functions;
         self.importedFunctions = importedFunctions;
         self.birBlocks = code.blocks;
-        // JBUG 31008 if this is a query expression
-        final Repr[] reprs = [];
-        foreach var reg in code.registers {
-            reprs.push(check semTypeRepr(reg.semType));
-        }
-        self.reprs = reprs;
+        self.reprs = from var reg in code.registers select check semTypeRepr(reg.semType);
```
jclark: I prefer to o:

```
final Repr[] reprs = from ...;
self.reprs = reprs;
```

The use reprs rather than self.reprs later.

jclark (changes requested):

compiler/modules/nback/build.bal
manuranga: When the comment was originally written, only `insn.args` was being iterated. Now that the `paramTypes` is being iterated along with it, I can't do it trivially.

I may if I write a `zip` iterator, is that a good option?

jclark: Oh I see. Never mind then.

jclark:
Can we close #190 or are there bugs you have not yet looked at?

manuranga:
Can't close yet, I haven't gone through all.

PR #220, opened 2021-07-13
Run tests on each pr and commit

manuranga:
Make github run tests on each pr and commit

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -1,0 +1,37 @@
+name: Make test
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+  workflow_dispatch:
+  
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


jclark:
Looks good, but I would like to make sure I completely understand what this is doing. Perhaps you can explain it to me tomorrow.

manuranga:
`apt-get` turned out to be the better option (at least for llvm-11) since it is pre-installed in github.

I noticed Make is compiling the ballerina code with `--skip-tests`. I like to run ballerina tests in this action as well.

Option 1: Run `bal test` before `make test`
Option 2: Have a flag like `make baltest=true  -j2 test`

Which one is the better approach?

manuranga:
updated with option 1

jclark:
If you do `bal test` before `make test`, then won't `make test` rebuild?

I don't think there's a perfect solution because of how `bal build` works, but I think this might work better: change the top-level Makefile so that

* test does not depend on the compiler jar
* test does `cd compiler && $(BAL) build --offline` before doing a recursive make

manuranga:
Updated to just do `make test`.

Issue #177, opened 2021-07-06
Make return value of `_bal_alloc` have `noalias` attribute


jclark:
Semantics are:

> On function return values, the noalias attribute indicates that the function acts like a system memory allocation function, returning a pointer to allocated storage disjoint from the storage for any other object accessible to the caller.

Note that this requires the LLVM backend to attach attributes to return values not just to functions.

jclark:
Also `_bal_mapping_construct`.

jclark:
Only true if the implementation of _bal_alloc is compiled separately (not just to `.bc`) from the other files

PR #230, opened 2021-07-15
Add support for argument/return value attributes in llvm

### LINKED ISSUE #227: LLVM interface needs support for `signext` and `zeroext` attributes on arguments/returns values
We need this to get compatibility with clang for boolean arguments.

This requires extending attributes to support applying to argument/return value rather than to function.

heshanpadmasiri:
## Purpose
Extending attributes to support applying to argument/return value.
Resolves #227
Related to #177

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/print.llvm/llvm.bal
+++ b/compiler/modules/print.llvm/llvm.bal
@@ -266,11 +266,16 @@ public class FunctionDecl {
     final FunctionType functionType;
     final string functionName;
     string? gcName = ();
-    final EnumAttribute[] attributes = [];
+    final FunctionEnumAttribute[] functionAttributes = [];
+    final ReturnEnumAttribute[] returnAttributes = [];
+    final ParamEnumAttribute[][] paramAttributes = [];
 
     function init(Context context, string functionName, FunctionType functionType) {
         self.functionName = functionName;
         self.functionType = functionType;
+        foreach var _i in 0..< functionType.paramTypes.length() {
+           self.paramAttributes.push([]); 
+        }
     }
 
     function output(Output out) {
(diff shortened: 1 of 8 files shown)


compiler/modules/print.llvm/llvm.bal
```diff
@@ -266,20 +266,46 @@ public class FunctionDecl {
...
     }
 
     public function addEnumAttribute(EnumAttribute attribute) {
-        if self.attributes.indexOf(attribute) == () {
-           self.attributes.push(attribute); 
+        if attribute is FunctionEnumAttribute {
+            self.pushAttributeToContainer(self.functionAttributes, attribute);
+        }
+        else {
+            if attribute[0] == "return" {
```
jclark: You should be able to write this without casting e.g.

```
if attribute is (readonly & ["return", ReturnEnumAttribute]) {
}
```

If that doesn't work, put a JBUG comment

compiler/modules/print.llvm/llvm.bal
```diff
@@ -266,20 +266,46 @@ public class FunctionDecl {
...
-    final EnumAttribute[] attributes = [];
+    final FunctionEnumAttribute[] functionAttributes = [];
+    final ReturnEnumAttribute[] returnAttributes = [];
+    final ParamEnumAttribute[][] paramAttributes = [];
 
     function init(Context context, string functionName, FunctionType functionType) {
         self.functionName = functionName;
         self.functionType = functionType;
+        foreach var _i in 0..< functionType.paramTypes.length() {
+           self.paramAttributes.push([]);
```
jclark: Should be able to use `setLength` here.

compiler/modules/print.llvm/llvm.bal
```diff
@@ -266,20 +266,46 @@ public class FunctionDecl {
...
+                if paramIndex >= self.paramAttributes.length() {
+                    panic err:illegalArgument("Invalid index for parameter attribute");
+                } else {
+                    self.pushAttributeToContainer(self.paramAttributes[paramIndex], attrib);
+                }
+            }
+        }
+    }
+
+    function pushAttributeToContainer((FunctionEnumAttribute|ParamEnumAttribute|ReturnEnumAttribute)[] container,
```
jclark: Not a great function name. You are using the list to represent a set. So maybe `addAttributeToSet`.  I would just use `string` instead of `FunctionEnumAttribute|ParamEnumAttribute|ReturnEnumAttribute`, since it isn't type-safe here (and should be `private`).

compiler/modules/print.llvm/llvm.bal

compiler/modules/print.llvm/llvm.bal

compiler/modules/print.llvm/llvm.bal

PR #241, opened 2021-07-19
Fix ast-to-string's function visibility order 

manuranga:
Fix ast-to-string's function visibility order.
Add details to testParserOnTestSuite failures .


manuranga:
moved to #242

PR #243, opened 2021-07-19
Test llvm.print string escape functions

### LINKED ISSUE #221: Fix llvm.print escaping of identifiers
escapeName has a number of problems including:

* `-` is allowed in identifiers
* when using double quotes not only `"` needs escaping, but also `\` and control characters
* doesn't handle Unicode at all

heshanpadmasiri:
## Purpose

- Add tests to verify `charArray` and `escapeIdent` correctly escape strings
- Update module function/global variables to correctly escape names

Resolves #221 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/print.llvm/llvm.bal
```diff
@@ -117,27 +117,29 @@ public class Module {
 
     // Corresponds to LLVMAddFunction
     public function addFunctionDefn(string name, FunctionType fnType) returns FunctionDefn {
-        if name is IntrinsicFunctionName {
+        string fnName = escapeIdent(name);
```
jclark: You've got this common code in three places (escape and panic if it's an intrinsic function name), so I suggest you factor this out into a separate function (e.g. `escapeGlobalIdent`).

jclark (changes requested):

compiler/modules/print.llvm/llvm.bal

PR #248, opened 2021-07-20
Implement int:toHexString

### LINKED ISSUE #239: Implement int:toHexString in C
This will allow us to easily generate strings for testing.

manuranga:
Fixes #239


runtime/int.c
```diff
@@ -0,0 +1,20 @@
+#include <stdio.h>
+#include "balrt.h"
+
+
+// lang.int:MIN_VALUE is the longest, which is 17 chars
+#define MAX_HEX_INT_LEN 17
+
+GC TaggedPtr _Bint__toHexString(int64_t i) {
```
jclark: You don't need GC here: TaggedPtr already has that.

runtime/int.c
```diff
@@ -0,0 +1,20 @@
...
+    int size = smallStringSize(MAX_HEX_INT_LEN);
```
jclark: I would do this differently, so that you don't allocate unnecessary space (also allow for nul byte):

```
char buf[1 + 2*8 + 1]; // allow for sign, 2 hex digits for each byte and the nul byte
len = sprintf(buf, ...)
SmallStringPtr p = _bal_alloc(len);
memcpy(p->bytes, buf, len)
```

runtime/int.c
```diff
@@ -0,0 +1,20 @@
...
+    SmallStringPtr p = _bal_alloc(size);
+    int len;
+    if (i < 0) {
+        // simply -i worked, but doing this to avoid UB
+        len = sprintf(p->bytes, "-%llx", ((uint64_t) -(i + 1)) + 1);
```
jclark: `-(uint64_t)i` is, I believe, sufficient to avoid UB

Passing a `uint64_t` to a `%llx` isn't ideal, because there's nothing to say that `long long` is exactly 64 bits. I just had a look at this and it turns out there's a PRIx64 macro for this (from inttypes.h).

(I need to fix print.c to use this also.)

jclark (changes requested):

manuranga:
Should I `memset(p, 0, size)` ?

jclark:
The padding should be zeroed. It's tricky so this should be factored out into _bal_string_alloc in string.c (and reused in testrt.c).

valgrind should give an error if you use == on this: let me see if it does. (I checked it does.)

jclark:
If you use memcpy, then you need to include `<string.h>`.

jclark:
I will fix this.

PR #249, opened 2021-07-21
LLVM module support for const expressions

### LINKED ISSUE #234: LLVM module support for const expressions
In the initializer of a global we need to be able to have at least:

- getElementPtr
- bitCast
- addressSpaceCast

heshanpadmasiri:
## Purpose
Add `GEP`, `bitcast` and `addrspacecast` constant expressions
Resolves #234 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/print.llvm/llvm.bal
+++ b/compiler/modules/print.llvm/llvm.bal
@@ -100,9 +100,12 @@ public class Context {
     }
 
     // Corresponds to LLVMConstGEP
-    function constGetElementPtr(ConstValue ptr, ConstValue[] indices, "inbounds"? inbounds=()) returns ConstValue {
-        var body = gepBody(ptr, indices, inbounds);
-        string[] words = <string[]>body[0];
+    function constGetElementPtr(PointerValue ptr, ConstValue[] indices, "inbounds"? inbounds=()) returns ConstValue {
+        var body = gepBody(ptr, indices, inbounds, "constantExp");
+        string[] words = [];
+        foreach var word in body[0] {
+           words.push(<string> word); 
+        }
         string operand = concat(...words);
         PointerType resultPtrType = body[1];
         return new (resultPtrType, operand);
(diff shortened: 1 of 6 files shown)


compiler/modules/print.llvm/llvm.bal
```diff
@@ -36,14 +36,22 @@ public readonly class PointerValue {
 # Subtype of Value that refers to a constant
 public readonly class ConstValue {
     *Value;
-    string operand;
+    string|Unnamed operand;
```
jclark: This makes ConstValue the same type as Value...  How can a ConstValue have referenced to Unnamed?

compiler/modules/print.llvm/llvm.bal
```diff
@@ -1031,6 +1022,88 @@ function escapeIdentChar(string:Char ch) returns string {
...
+        words.push(")");
+    }
+    PointerType resultPtrType = pointerType(resultType, resultAddressSpace);
+    return [words, resultPtrType];
+}
+
+function bitCastBody(Value val, PointerType destTy, "constantExp"? constantExp=()) returns (string|Unnamed)[] {
+    (string|Unnamed)[] words = [];
+    words.push("bitcast");
+    if constantExp != () {
```
jclark: These repeated `if constantExp` don't feel clean.  I think this would work better

- make the function be `bitCastArgs`
- do not have a `constantExp` argument
- pass `words` as the argument to bitCastArgs
- `bitCastArgs` constructs the part that needs to go in parentheses when it's const (i.e. not including `bitcast`)
- insert the parentheses in `constBitCast` arounds calls to `bitCastArgs`

Similarly for the other two.

jclark (changes requested):

compiler/modules/print.llvm/llvm.bal
heshanpadmasiri: It was there to allow casting between `ConstValue` and `PointerValue`. Fixed by adding explicit functions to create `ConstValue` from `Value` and `PointerValue`, and `PointerValue` from `ConstValue`.

compiler/modules/print.llvm/llvm.bal

jclark:
It should be `constAddrSpaceCast` not  `costAddrSpaceCast`.

All this toPointerValue/toConstValus stuff is not the right way to do things. I will fix this.

PR #251, opened 2021-07-21
Implement string concatenation

### LINKED ISSUE #152: Implement string concatenation
This provides the `+` operator on strings, which requires the ability to construct new strings at runtime.

This builds on #130.

manuranga:
Implement string concatenation
Fixes #152

### DIFF
--- a/docs/subset04.md
+++ b/docs/subset04.md
@@ -203,7 +203,6 @@ Language spec syntax references:
 
 The implementation of `string` has the following additional restrictions:
 
- * `+` operator is not supported with operands of type `string`
  * member access `s[i]` is not supported when `s` has type `string`
 
 Method call syntax can be used for calling the following langlib functions:
@@ -238,6 +237,7 @@ Add `string` and `map<any>`:
 Existing syntax extended:
 
 * `===`, `!==` for string and map
+* `+` for string
 * `==`, `!=` for string
 * `<`, `<=`, `>`, `>=` for string
 * `<T>E` for `string` and `map<any>`
--- a/runtime/balrt_inline.c
+++ b/runtime/balrt_inline.c
@@ -1,3 +1,4 @@
+#include <string.h>
 #include "balrt.h"
 
 #define FLAG_INT_ON_HEAP 0x20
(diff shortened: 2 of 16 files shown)


runtime/testrt.c
```diff
@@ -35,6 +35,11 @@ static TaggedPtr copySmallString(TaggedPtr tp) {
     return copy;
 }
 
+int64_t stringLen(TaggedPtr p) {
```
manuranga, line 38: using `_Bstring__length` didn't work, got `ld: symbol(s) not found` but only with `-O2`.

jclark: Maybe need to link in balrt_inline.bc

docs/subset04.md
```diff
@@ -238,6 +238,7 @@ Add `string` and `map<any>`:
 Existing syntax extended:
 
 * `===`, `!==` for string and map
+* `+` for string
```
jclark, line 240: IIRC the doc says somewhere that `+` is not allowed for string.

jclark (changes requested):

runtime/testrt.c
```diff
@@ -227,6 +255,7 @@ int main() {
     testStringCmp();
     testStringEq();
     testStringHash();
+    testStringConcatAssociative();
```
jclark, line 298: We should have a bit more testing that this.

I suggest concatenating multiple strings and then checking with memcmp that the StringData of the things that are concatenated appears at the right place in the StringData of the concatenation.

jclark (changes requested):

runtime/testrt.c
manuranga: It is already linked, maybe since it's a `.bc` symbol is not getting picked up.

https://github.com/ballerina-platform/nballerina/blob/f637cf79e0a521272c5ff74b98b4575387b77ee1/runtime/Makefile#L24

runtime/balrt_inline.c
```diff
@@ -142,3 +143,15 @@ bool _bal_string_eq(TaggedPtr tp1, TaggedPtr tp2) {
     }
     return true;
 }
+
+TaggedPtr _bal_string_concat(TaggedPtr tp1, TaggedPtr tp2) {
+    StringData s1 = _bal_tagged_to_string(tp1);
+    StringData s2 = _bal_tagged_to_string(tp2);
+    int64_t byteLen = s1.lengthInBytes + s2.lengthInBytes;
```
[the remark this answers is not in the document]
jclark: I have made the first argument to _bal_string_alloc be unsigned, so we can deal with length > INT_MAX there.

jclark (changes requested):
You also need to implement string concatenation in const.bal so as to recognize e.g. `if ("x" == "y")` is false (which impacts reachability).

PR #261, opened 2021-07-25
Implement unary `~` operator

### LINKED ISSUE #214: Implement unary `~` operator
This doesn't need to be in BIR (can use xor with all ones).

Both codegen and constant folding need this.

manuranga:
Implement unary `~` operator.

Fixes #214


jclark (changes requested):
Please also use "complement" or an abbreviation in the test case name, not "twiddle".

PR #266, opened 2021-07-26
Use LLVM JNI to generate object files directly

### LINKED ISSUE #240: Use LLVM JNI to generate object files directly
The goal is to be able to generate object files directly using the LLVM JNI interface.

The jni.llvm module should have additional methods that provide access to the needed APIs of LLVM.  Then write a separate main.bal in llvm_jni that uses these methods to implement a program that compiles a .bal file into an object file. This needs to combine the module with the `runtime/balrt_inline.bc` module (as done by llvm-link) before compiling/optimizing into an object file. I suggest storing the `balrt_inline.bc` in the compile jar file.

Later this main.bal can also call the jBallerina front-end.

heshanpadmasiri:
## Purpose
Add ability to generate object files directly from `jni.llvm`
Resolves #240 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


llvm_jni/modules/jni.llvm/Module.bal
```diff
@@ -7,9 +7,19 @@ function intrinsicNameToId(IntrinsicFunctionName name) returns int {
     return jLLVMLookupIntrinsicID(java:fromString(str_name), str_name.length());
 }
 
+// JBUG: #31447 make possible values for these strict
+public type LLVMCodeGenOptLevel string;
+public type LLVMRelocMode string;
+public type LLVMCodeModel string;
+
+configurable string optLevel = "Default";
```
jclark: I don't think something like optLevel should be handled via configurable. This is something that should be a command-line option.

llvm_jni/main.bal
```diff
@@ -4,10 +4,11 @@ public function main() {
     llvm:Context context = new;
     llvm:Builder builder = context.createBuilder();
     llvm:Module m = context.createModule();
-    llvm:Function mainFunction = m.addFunctionDefn("main", {returnType: "void", paramTypes: ["i64", "i1"]});
+    llvm:Function mainFunction = m.addFunctionDefn("main", {returnType: "i64", paramTypes: ["i64", "i1"]});
     llvm:BasicBlock initBlock = mainFunction.appendBasicBlock();
```
jclark: We need a proper command-line driver program that takes a `.bal` file plus some command-line arguments and compiles it to a `.o` file. You should use Ballerina's built-in CLI handling capabilities like the nballerina module does.

jclark (changes requested):

llvm_jni/main.bal
heshanpadmasiri: Fixed by modifying the driver to directly take the optional command line arguments to object file generation and changed the `print.llvm` to panic if asked to generate object file

jclark: That's not what I'm looking for. I want a **separate** driver program in llvm_jni. Doesn't have to be fancy: just enough to talk a file and compile it to `.o`.

heshanpadmasiri: For clarification what you are suggesting is move the parts that has to be different (`main.bal` and `nback.bal`) in to the llvm_jni (and sym link the reset of compiler in to llvm_jni because above 2 depend on them for development purposes) and at the build time (when are currently replace print.llvm with jni.llvm) replace those different files with the versions from jni.llvm

jclark: I am suggesting a new `main.bal` file (completely separate from the current one) that provides a `main` function that uses `nback`, `front` and `jni.llvm` modules to compile a .bal file into a `.o` file. This may require refactoring/extending `nback` so that `main` has access to the llvm:Module and llvm:Context (e.g. we could make `buildModule` public). Please suggest something.

The reason is that a `main` that can compile all the way to `.o` is (eventually) going to have a ton of options (which will be passed to LLVM) that will be irrelevant for print.llvm. This `main` will also (eventually) be responsible for managing the jBallerina front-end.

heshanpadmasiri: I think the easiest way will be to change the `compileFile` function in `main.bal` and `nback`'s `compileModule` functions. We will change `compileModule` so that it will return a `llvm:module` (similar to `front:loadModule` return a `bir:Module`) instead of taking care of output as well. Then `main.bal` will decided how to handle output. This way `main` will have direct access to all three `nback`, `front` and `jni.llvm`|`print.llvm` allowing it to pass arguments to any of the three without relying on others to do that.

jclark: `compileModule` will then be doing the same thing that `buildModule` currently does. So we can just

- move `nback:compileModule` into `main.bal`
- make `nback:buildModule` public (and perhaps rename it)

heshanpadmasiri: Fixed

compiler/main.bal
```diff
@@ -50,16 +52,26 @@ function validGcName(string? gcName) returns string|error? {
...
+        llMod.setTarget(nback:target);
+    }
+
+    if outputFileName != () {
+        check llMod.printModuleToFile(outputFileName);
+    }
 }
 
-function chooseOutputFilename(string sourceFilename, string? outDir) returns string|error? {
+function chooseOutputFilename(string sourceFilename, string outputExtension, string? outDir) returns string|error? {
```
jclark: Don't need the outputExtension parameter

llvm_jni/modules/nback/build.bal
```diff
@@ -0,0 +1,1384 @@
+// Copy of build.bal from compiler with imports changed to jni.llvm. Not used in final build
```
jclark: Why is this checked in?

llvm_jni/modules/jni.llvm/Module.bal
```diff
@@ -77,10 +122,90 @@ public distinct class Module {
     }
 
     public function setTarget(TargetTriple targetTriple) {
+        self.targetTriple = targetTriple;
         jLLVMSetTarget(self.LLVMModule, java:fromString(targetTriple));
     }
 }
 
+
+function getLLVMCodeGenOpLevel(LLVMCodeGenOptLevel codeGenLevel) returns int {
```
jclark: OptLevel not OpLevel

jclark (changes requested):

llvm_jni/modules/nback/build.bal
heshanpadmasiri: We need this file (file that import `jni.llvm` instead of `print.llvm` and has the correct interface for `main.bal`) so vs code won't errors.

jclark: This should be generated by transforming the original build.bal at build time.

heshanpadmasiri: We do that when building with gradle but but if I symlink this file (similar to other common files) since this import `print.llvm` in the original file vs code show this file as an error

llvm_jni/modules/jni.llvm/Module.bal
heshanpadmasiri: Fixed

compiler/main.bal
heshanpadmasiri: Fixed

llvm_jni/modules/nback/build.bal
heshanpadmasiri: Fixed by changing file to a symlink

Issue #262, opened 2021-07-26
Implement `float` type


jclark:
Requires:

- `float` type descriptor
- floating point literals
- extend arithmetic operators to float
- extend relational operators to float (these follow IEEE rules)
- extend `==` and `===` to float (these are not the same as IEEE numeric equality)
- extend type cast to float
- io:println on float
- extend tagged pointers to represent float
- float constant evaluation
- BIR instructions for float
- some langlib float functions (exact list TBD)

Note that LLVM hexadecimal float constants can be produced using the langlib float:toBitsInt function.

jclark:
Not sure what `==` and `===` will turn into: maybe need to use LLVM `select` instruction.

jclark:
Also requires extending semtypes to do singleton float types.

jclark:
Another problem is that with the addition of `float`, and integer literal is either of type `int` or `float` depending on the contextually expected type. I will try to deal with this when I have another look at how we handle constants.

manuranga:
~~Until then, I'll restrict the syntax to mandate `f` suffix for non-dotted float literals.~~
A syntax level restriction will not work.

manuranga:
Interesting tokenizer case :`0.e0f()`

manuranga:
Another difficult case `10..toHexString()`.

manuranga:
To handle

jclark:
Please explain the problem. The lexer chooses the longest valid token.

jclark:
That is to say, the longest string that is a token.

manuranga:
Please disregard this, my concerns are baseless since there are no functions in `lang.int` starting with `f` or `e`. I'll use this fact.
I was thinking, if there was,`32.factor()` would be tokenized as `32.f` `actor()`.

manuranga:
When we are parsing a `1.toHexString()` and we see `.`, we still need to see the next char to determine if it's a decimal or float. After setting `t` we need to unget `t` **and** `.` and then return `[DECIMAL_NUMBER, "1"]`. 

(And we are not returning the longest token which is `[DECIMAL_FLOATING_POINT_NUMBER, "1."]`)

jclark:
The way tokenization is supposed to work is to return the longest string that is a valid token. So `1.toHexString()` lexes as the tokens  `1.`   `toHexString`  `(`  `)` which will give a parse error.

jclark:
If you want to do something else, please report it as a spec issue.

[...]

PR #269, opened 2021-07-28
Tokenize floats

manuranga:
Part of #262

### DIFF
--- a/compiler/modules/front/tests/tokenTests.bal
+++ b/compiler/modules/front/tests/tokenTests.bal
@@ -12,6 +12,25 @@ function test1() {
     test:assertEquals(tokenization(string`"abc\r\n\"" | "" | "123"`),
                       [[STRING_LITERAL, "abc\r\n\""], "|", [STRING_LITERAL, ""], "|", [STRING_LITERAL, "123"]]);
     test:assertEquals(tokenization("ab.xy . f()"), [[IDENTIFIER, "ab"], ".", [IDENTIFIER, "xy"], ".",  [IDENTIFIER, "f"], "(", ")"]);
+    test:assertEquals(tokenization("32.factor()"), [[DECIMAL_NUMBER, "32"], ".", [IDENTIFIER, "factor"], "(", ")"]);
+
+    test:assertEquals(tokenization("0.0 0f 0.0f 0e01 .00"), [[DECIMAL_FP_NUMBER, "0.0", ()], [DECIMAL_FP_NUMBER, "0", "f"], [DECIMAL_FP_NUMBER, "0.0", "f"], [DECIMAL_FP_NUMBER, "0e01", ()], [DECIMAL_FP_NUMBER, ".00", ()]]);
+    test:assertEquals(tokenization("-4f 2.0"), ["-", [DECIMAL_FP_NUMBER, "4", "f"], [DECIMAL_FP_NUMBER, "2.0", ()]]);
+    test:assertEquals(tokenization("1.2.floor()"), [[DECIMAL_FP_NUMBER, "1.2", ()], ".", [IDENTIFIER, "floor"], "(", ")"]);
+    test:assertEquals(tokenization("1.3e1 1.3e1F"), [[DECIMAL_FP_NUMBER, "1.3e1", ()], [DECIMAL_FP_NUMBER, "1.3e1", "f"]]);
+    test:assertEquals(tokenization("1.e1"), [[DECIMAL_NUMBER, "1"], ".", [IDENTIFIER, "e1"]]);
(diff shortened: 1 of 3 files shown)


compiler/modules/front/tests/tokenTests.bal
```diff
@@ -6,12 +6,23 @@ function test1() {
...
+
+    test:assertEquals(tokenization("0.0 0.0f 0e01 .00"), [[DECIMAL_FP_NUMBER, "0.0"], [DECIMAL_FP_NUMBER, 0.0], [DECIMAL_FP_NUMBER, "0e01"], [DECIMAL_FP_NUMBER, ".00"]]);
+    test:assertEquals(tokenization("-4f 2.0"), ["-", [DECIMAL_FP_NUMBER, 4.0], [DECIMAL_FP_NUMBER, "2.0"]]);
+    test:assertEquals(tokenization("1.2.floor()"), [[DECIMAL_FP_NUMBER, "1.2"], ".", [IDENTIFIER, "floor"], "(", ")"]);
+    test:assertEquals(tokenization("1e-5 1e-5f"), [[DECIMAL_FP_NUMBER, "1e-5"], [DECIMAL_FP_NUMBER, 1.0E-5]]);
+    test:assertEquals(tokenization("1.e1 1.e1F"), [[DECIMAL_FP_NUMBER, "1.e1"], [DECIMAL_FP_NUMBER, 10.0]]);
+    test:assertEquals(tokenization(".5e10 .5e10f"), [[DECIMAL_FP_NUMBER, ".5e10"], [DECIMAL_FP_NUMBER, 5.0E9]]);
+    test:assertEquals(tokenization("-10.0e-2 -10.0e-2f"), ["-", [DECIMAL_FP_NUMBER, "10.0e-2"], "-", [DECIMAL_FP_NUMBER, 0.1]]);
+
+    test:assertEquals(tokenization(".e1"), [".", [IDENTIFIER, "e1"]]);
```
manuranga, line 26: Currently  `1.e-()` gives an error when torkenizing, perhaps it's more correct to torkenize it as `[[DECIMAL_FP_NUMBER, 1.0], [IDENTIFIER, "e"], "-", "(", ")"]]`. That may require significant change.

jclark: Yes, that is the correct answer (except `[DECIMAL_FP_NUMBER, "1."]`.

We need to be able to do:

`1. e = 1.0`

jclark:
There seems to be a lot of breakage from allowing trailing `.` in a floating point number. We have maybe got do something about that at the spec level. So can we simply not allow that for now?

I have made a spec issue for this. https://github.com/ballerina-platform/ballerina-spec/issues/905

compiler/modules/front/token.bal
```diff
@@ -282,6 +304,80 @@ class Tokenizer {
         return self.err("invalid token");
     }
 
+    private function nextDecimalFPNumber(string preDot) returns Token|err:Syntax {
```
jclark: Ballerina convention in a case like this is  `Fp` not `FP`

compiler/modules/front/token.bal
```diff
@@ -282,6 +304,80 @@ class Tokenizer {
...
+            }
+            else {
+                self.ungetc(ch);
+                break;
+            }
+        }
+        return [DECIMAL_FP_NUMBER, digits];
+    }
+
+    private function nextDecimalExpedFPNumber(string preExp, Char e) returns Token|err:Syntax {
```
jclark: No idea what `exped` means

jclark (changes requested):
If we don't accept trailing `.`, we can get rid of all the test case changes.

compiler/modules/front/token.bal
manuranga: I meant to say "Exp-ed" meaning "Exponentiated", I'll go with "nextDecimalExpFpNumber" instead.

compiler/modules/front/token.bal
```diff
@@ -7,9 +7,10 @@ const DECIMAL_NUMBER = 1;
 const STRING_LITERAL = 2;
 const BOOLEAN_LITERAL = 3;
 const HEX_INT_LITERAL = 4;
+const DECIMAL_FP_NUMBER = 5;
 
 // Use string for DECIMAL_NUMBER so we don't get overflow on -int:MAX_VALUE
-type VariableLengthToken [IDENTIFIER, string]|[DECIMAL_NUMBER, string]|[STRING_LITERAL, string]|[HEX_INT_LITERAL, string];
+type VariableLengthToken [IDENTIFIER, string]|[DECIMAL_NUMBER, string]|[STRING_LITERAL, string]|[HEX_INT_LITERAL, string]|[DECIMAL_FP_NUMBER, float|string];
```
jclark: On further reflection I think a cleaner way to represent the token is: `[DECIMAL_FP_NUMBER, string, FpTypeSuffix?]` where

```
type FpTypeSuffix "f";
```

jclark (changes requested):

Issue #263, opened 2021-07-26
Improve runtime library testing


jclark:
The current testing is rather skimpy and ad hoc.

We should
- put tests in separate directory
- split up into separate testing programs
- have a Makefile that allows tests to be run in parallel
- ensure we have good test coverage using clang's code coverage support https://clang.llvm.org/docs/SourceBasedCodeCoverage.html

jclark:
@heshanpadmasiri I suggest you start with a simple pull request that just splits up the current testrt.c into multiple programs in a test directory and modifies the makefiles accordingly.

jclark:
We should provide a way to test io:println. We can use this as part of staging #262

PR #273, opened 2021-07-30
Runtime tests for strings

heshanpadmasiri:
## Purpose

- Add tests to validate `mediumStringSize` and `largeStringSize`
- Add tests to validate `immediateStringLength` and additional tests for `taggedStringLength`
- Generate test coverage reports using `make`

Related to #263

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


runtime/tests/immediateString.c
```diff
@@ -0,0 +1,53 @@
...
+void testImmediateStringLength() {
+	// test ASCII
+	validateImmediateStringLength(0xABFFFFFFFFFFFFFF, 0, 0); // ""
+	validateImmediateStringLength(0xABFFFFFFFFFF4142, 2, 2); // "AB"
+	validateImmediateStringLength(0xABFFFFFFFF613223, 3, 3); // "a2#"
+	validateImmediateStringLength(0xABFFFFFF256C6420, 4, 4); // "%ld "
+	validateImmediateStringLength(0xABFFFF6162606364, 5, 5); // "ab`cd"
+	validateImmediateStringLength(0xABFF2A2625242C3F, 6, 6); // "*&%$,?"
+	validateImmediateStringLength(0xAB61626364656667, 7, 7); // "abcdefg"
+	// testUnicode
```
heshanpadmasiri, line 30: @jclark I think there is a bug in `immediateStringLength` with unicode characters. I have commented the wrong values I got here.

jclark: This is not correct UTF-8. In the UTF-8 encoding of a non-ASCII character, all bytes have the top bit set and the first byte has the top 2 bits set. https://en.wikipedia.org/wiki/UTF-8#Encoding

In 0xABFFFFFFFFFF265E, 0x5E is supposed to be the first byte but it does not have the top bit set.

And why is the top byte 0xAB. It should be 0x2A, shouldn't it.

heshanpadmasiri: I fixed the above problems but I still get the number of bytes wrong (1).

jclark: The lo byte is still wrong. For a UTF-8 sequence of three bytes, the first byte must be 0b1110xxxx whereas 0x9E is 0b10011110.

heshanpadmasiri: To be clear if I get a utf-8 encoding as `E2` `99` `9E` (for "♞") then it will get encoded as `0x2AFFFFFFFF9E99E2` (first byte as the right most one) not not `0x2AFFFFFFFFE2999E` (first byte as the left most after padding). It work fine in the former case.

jclark: Yes, the UTF-8 bytes are in little-endian order in the `int`.

PR #286, opened 2021-08-06
Add float singleton type

### LINKED ISSUE #284: Add float singleton types by generifying string type logic
Generify string type union, intersect, diff algorithms by passing a comparator. 
Use that to create float singleton type.

Part of #262.

manuranga:
Fixes #284

### DIFF
--- a/compiler/modules/types/enumerable.bal
+++ b/compiler/modules/types/enumerable.bal
@@ -1,0 +1,225 @@
+import wso2/nballerina.err;
+
+// EnumerableTypes are types in which each expressible subtype can be reasonably
+// represented by a list of values. counter-eg: `int` is not a EnumerableType since
+// uint32 can't be reasonably represented by listing all values.
+type EnumerableType float|string;
+
+type EnumerableSubType FloatSubtype|StringSubtype;
+
+const LT = -1;
+const EQ = 0;
+const GT = 1;
+
+type Order GT|EQ|LT;
+
+function enumerableSubtypeUnion(EnumerableSubType t1, EnumerableSubType t2) returns SubtypeData {
+    boolean b1 = t1.allowed;
+    boolean b2 = t2.allowed;
+    boolean allowed;
+    EnumerableType[] values = arrayOfType(t1.values);
+    if b1 && b2 {
+        enumerableListUnion(t1.values, t2.values, values);
+        allowed = true;
+    }
+    else if !b1 && !b2 {
+        enumerableListIntersect(t1.values, t2.values, values);
+        allowed = false;
+    }
+    else if b1 && !b2 {
+        enumerableListDiff(t2.values, t1.values, values);
+        allowed = false;
+    }
+    else {
+         // !b1 && b2
+        enumerableListDiff(t1.values, t2.values, values);
+        allowed = false;
+    }
+    return createEnumerableSubtype(allowed, values);
+}
+
+function enumerableSubtypeIntersect(EnumerableSubType t1, EnumerableSubType t2) returns SubtypeData {
+    boolean b1 = t1.allowed;
+    boolean b2 = t2.allowed;
+    boolean allowed;
+    EnumerableType[] values = arrayOfType(t1.values);
+    if b1 && b2 {
+        enumerableListIntersect(t1.values, t2.values, values);
+        allowed = true;
+    }
+    else if !b1 && !b2 {
+        enumerableListUnion(t1.values, t2.values, values);
+        allowed = false;
+    }
+    else if b1 && !b2 {
+        enumerableListDiff(t1.values, t2.values, values);
+        allowed = true;
+    }
+    else {
+        // !b1 && b2
+        enumerableListDiff(t2.values, t1.values, values);
+        allowed = true;
+    }
+    return createEnumerableSubtype(allowed, values);
+}
+
+function enumerableListUnion(EnumerableType[] v1, EnumerableType[] v2, EnumerableType[] result) {
+    int i1 = 0;
+    int i2 = 0;
+    int len1 = v1.length();
+    int len2 = v2.length();
+    while true {
+        if i1 >= len1 {
+            if i2 >= len2 {
+                break;
+            }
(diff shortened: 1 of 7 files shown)


compiler/modules/types/flaot.bal
```diff
@@ -0,0 +1,98 @@
...
+       return createFloatSubtype(false, floatListDiff(t2.values, t1.values));
+    }
+    else {
+        // !b1 && b2
+       return createFloatSubtype(false, floatListDiff(t1.values, t2.values));
+    }
+}
+
+function floatSubtypeIntersect(SubtypeData d1, SubtypeData d2) returns SubtypeData {
+    FloatSubtype t1 = <FloatSubtype>d1;
```
manuranga: Should I generify this part by creating `type ListableSubtype FloatSubtype|StringSubtype`

compiler/modules/types/listable.bal
```diff
@@ -0,0 +1,148 @@
+// ListableTypes are types in which each expressible subtype can be reasonably
+// represented by a list of values. counter-eg: `int` is not a ListableType since
+// uint32 can't be reasonably represented by listing all values.
+type ListableType float|string;
```
manuranga: is `ListableType` a good name for this ?

jclark: I suggest `EnumerableType`

compiler/modules/types/flaot.bal
jclark: Yes, `EnumerableSubtype`

compiler/modules/types/flaot.bal
```diff
@@ -0,0 +1,98 @@
+// Implementation specific to basic type float.
```
jclark: Filename is misspelled

compiler/modules/types/listable.bal
```diff
@@ -0,0 +1,148 @@
...
+
+const LT = -1;
+const EQ = 0;
+const GT = 1;
+
+type Order GT|EQ|LT;
+
+function listUnion(ListableType[] v1, ListableType[] v2, ListableType[] result) {
```
manuranga: Accepting `result` as a pram feels bit unclean, but this avoids a cast.

jclark: I think adding the param is the right choice, but I would add a word (maybe `into`) to the name to make it clear that this is not working functionally (unlike `int`).

manuranga:
Ready to be reviewed.
Tests don't cover  list intersect, I think we need `!` type constructor to do that easily.

compiler/modules/types/listable.bal
manuranga: I prefixed `add`, could have gone with `collect` or `accumulate`.

compiler/modules/types/float.bal
```diff
@@ -0,0 +1,61 @@
...
+    boolean allowed;
+    float[] values;
+|};
+
+public function floatConst(float value) returns SemType {
+    FloatSubtype st = { allowed: true, values: [value] };
+    return uniformSubtype(UT_FLOAT, st);
+}
+
+function floatSubtypeUnion(SubtypeData d1, SubtypeData d2) returns SubtypeData =>
```
jclark: Don't see why we need this. Can't we just put enumerableSubtypeUnion in the vtable?

jclark (changes requested):

compiler/modules/types/enumerable.bal
```diff
@@ -0,0 +1,225 @@
+import wso2/nballerina.err;
+
+// EnumerableTypes are types in which each expressible subtype can be reasonably
+// represented by a list of values. counter-eg: `int` is not a EnumerableType since
+// uint32 can't be reasonably represented by listing all values.
+type EnumerableType float|string;
+
+type EnumerableSubType FloatSubtype|StringSubtype;
```
jclark, line 8: It's `Subtype` not `SubType`.

compiler/modules/types/string.bal
```diff
@@ -194,10 +50,12 @@ function createStringSubtype(boolean allowed, string[] values) returns SubtypeDa
 }
 
 final UniformTypeOps stringOps = {
-    union: stringSubtypeUnion,
+    union: function (SubtypeData d1, SubtypeData d2) returns SubtypeData =>
```
jclark, line 53: Just put enumerableSubtypeUnion directly in the ops table (and make the parameter type be SubtypeData). Similarly for all the other functions in the ops table.

jclark (changes requested):

jclark:
Also finish this off by

- making resolveTypes use t:floatConst
- adding a float-singleton.bal showTypes test case

jclark:
We can tweak this so that it will work with string:Char (which will need additional members in StringSubtype) as follows:

```
function enumerableSubtypeUnion(EnumerableSubType t1, EnumerableSubType t2) returns SubtypeData {}
```

becomes

```
function enumerableSubtypeUnion(EnumerableSubtype t1, EnumerableSubtype t2, EnumerableSubtype result) {}
```

where EnumerableSubtype is an open record:

```
type EnumerableSubtype readonly & record {
    boolean allowed;
    EnumerableType[] values;
};
```

We will then still need stringSubtypeUnion and floatSubtypeUnion as trivial wrappers around the corresponding enumerable* function.

Please don't use anonymous functions in the ops table.

manuranga:
In order to write showTypes for NaN and Infinity, I'll need #293.

(eg: 
```ballerina
const float NAN = -1.0/0.0;
type NaN NAN
```
)

jclark:
You don't need to be able to put type descriptors in const declarations. You can just say

```
const NaN = 1.0/0.0;
```

This requires expressions to allow floating point literals. Added as #294.

manuranga:
> You don't need to be able to put type descriptors in const declarations.

Great, I was misled since jBallerina required that. Guess that's a spec deviation.

jclark:
> Tests don't cover list intersect, I think we need `!` type constructor to do that easily.

Yeah, we definitely need `!`.

PR #317, opened 2021-08-14
Implement equality for `float`

### LINKED ISSUE #279: Implement equality for `float`
Part of #262 is to make the equality operators work for float.

This will require splitting _bal_eq into two functions, one for `==` and one for `===`.

This is follow on from #278

manuranga:
Implement equality for `float`

Fixes #279

### DIFF
--- a/compiler/modules/front/const.bal
+++ b/compiler/modules/front/const.bal
@@ -200,20 +200,50 @@ function foldBinaryEqualityExpr(FoldContext cx, t:SemType? expectedType, s:Binar
     if leftExpr is s:ConstValueExpr && rightExpr is s:ConstValueExpr {
         SimpleConst left = leftExpr.value;
         SimpleConst right = rightExpr.value;
-        boolean equal = left == right;
+        boolean exact = (<string>expr.equalityOp).length() == 3; // either "===" or "!=="
+        boolean equal = exact ? isExactEqual(left, right) : isEqual(left, right);
+        boolean positive = expr.equalityOp[0] == "=";
+        boolean value = positive == equal;
+        if exact {
+            return <s:ConstValueExpr> { value, multiSemType: t:BOOLEAN };
+        }
         if !equal && simpleConstExprIntersectIsEmpty(leftExpr, rightExpr) {
             return cx.semanticErr(`intersection of types of operands of ${expr.equalityOp} is empty`);
         }
-        boolean positive = expr.equalityOp[0] == "=";
-        if (<string>expr.equalityOp).length() == 2 {
-            return foldedBinaryConstExpr(positive == equal, t:BOOLEAN, leftExpr, rightExpr);
-        }
-        // This is the === or !== case
-        return <s:ConstValueExpr> { value: positive == (left === right), multiSemType: t:BOOLEAN };
+        return foldedBinaryConstExpr(value, t:BOOLEAN, leftExpr, rightExpr);
     }
     expr.left = leftExpr;
     expr.right = rightExpr;
     return expr;
+}
+
+// XXX remove after JBUG #17977 is fixed
+function isEqual(SimpleConst c1, SimpleConst c2) returns boolean{
+    if c1 is float {
+        float c1f = <float> c1;
+        if c2 is float {
+            float c2f = <float> c2;
+            // JBUG #32173 can't use `==` even for non-NaN case, -0.0 == 0.0 is wrong when not casted to float.
+            return float:isNaN(c1f) && float:isNaN(c2f) ? true : c1f == c2f;
+        }
+        else {
+            return false;
+        }
+    }
+    return c1 == c2;
+}
+
+// XXX remove after JBUG #17977 is fixed
+function isExactEqual(SimpleConst c1, SimpleConst c2) returns boolean{
+    if c1 is float && float:isNaN(c1) {
+        if c2 is float && float:isNaN(c2) {
+            return true;
+        }
+        else {
+            return false;
+        }
+    }
+    return c1 === c2;
 }
 
 // Precondition is that the values are !=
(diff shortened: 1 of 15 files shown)


manuranga:
Not fully done yet. Will make non-draft when done.

compiler/modules/front/const.bal
```diff
@@ -200,7 +200,8 @@ function foldBinaryEqualityExpr(FoldContext cx, t:SemType? expectedType, s:Binar
     if leftExpr is s:ConstValueExpr && rightExpr is s:ConstValueExpr {
         SimpleConst left = leftExpr.value;
         SimpleConst right = rightExpr.value;
-        boolean equal = left == right;
+        // XXX remove call to floatEq after JBUG #17977 is fixed
+        boolean equal = left is float && right is float ? t:floatEq(left, right) : left == right;
```
jclark: Better to define an isEqual function.  t:floatEq should not be public.

compiler/testSuite/Efloat00.bal
```diff
@@ -0,0 +1,6 @@
+import ballerina/io;
+
+public function main() {
+    io:println(1.0 == 2.0); // @error
```
jclark, line 4: Need more E tests (e.g. `!=` float/int)

compiler/testSuite/Vfloat01.bal
```diff
@@ -0,0 +1,29 @@
...
+    io:println(eq(nInf, nInf)); // @output true
+    
+    io:println(100.0 == 10e1); // @output true
+    io:println(0.0 == 0.0); // @output true
+    io:println(0.0 == -0.0); // @output true
+    io:println(nan0 == nan1); // @output true
+}
+
+function eq(float f1, float f2) returns boolean {
+    return f1 == f2;
```
jclark: Can test `!=` at the same time. Return some other value if != isn't not ==.

jclark (changes requested):

compiler/testSuite/Efloat05.bal
```diff
@@ -0,0 +1,8 @@
+import ballerina/io;
+
+public function main() {
+    if 0.0 / 0.0 === -0.0 / 0.0 { // @error
+        return;
+    }
+    io:println("impossible");
```
manuranga: @jclark
I think this should be an unreachable error but currently it's not because of `multiSemType`.
I am going to change this to not to produce a `multiSemType`, is that correct?

jclark: No it's not an unreachable error, because type of === is not a singleton https://github.com/ballerina-platform/ballerina-spec/issues/902#issuecomment-887136996

runtime/eq_inline.c
```diff
@@ -16,6 +17,36 @@ bool READONLY _bal_eq(TaggedPtr tp1, TaggedPtr tp2) {
...
+        return 0;
+    }
+    if (tag1 == TAG_STRING) {
+        return taggedStringEqual(tp1, tp2);
+    }
+    if (tag1 == (TAG_INT|FLAG_INT_ON_HEAP)) {
+        IntPtr p1 = taggedToPtr(tp1);
+        IntPtr p2 = taggedToPtr(tp2);
+        return *p1 == *p2;
+    }
```
manuranga, line 45: this part is duplicated, should I move to a function?

compiler/modules/front/const.bal
```diff
@@ -200,22 +200,52 @@ function foldBinaryEqualityExpr(FoldContext cx, t:SemType? expectedType, s:Binar
...
     }
     expr.left = leftExpr;
     expr.right = rightExpr;
     return expr;
 }
 
+// XXX remove after JBUG #17977 is fixed
+function isEqual(SimpleConst c1, SimpleConst c2) returns boolean{
+    if c1 is float {
+        float c1f = <float> c1;
```
jclark, line 223: You shouldn't need a cast there

compiler/modules/front/const.bal
```diff
@@ -200,22 +200,52 @@ function foldBinaryEqualityExpr(FoldContext cx, t:SemType? expectedType, s:Binar
...
+        if c2 is float {
+            float c2f = <float> c2;
```
jclark, line 225: Nor there

compiler/modules/print.llvm/llvm.bal
```diff
@@ -71,8 +71,9 @@ public function constInt(IntType ty, int val) returns ConstValue {
 
 // Corresponds to LLVMConstReal
 public function constReal(RealType ty, float val) returns ConstValue {
-    // XXX do hex conversion
-    return new ConstValue(ty, val.toString());
+    // XXX #307 use decimal for non-special floats 
+    return new ConstValue(ty, "0x" + int:toHexString(float:toBitsInt(val)));
```
jclark, line 75: This is a separate issue

runtime/eq_inline.c
jclark: OK as is for now

jclark (approved):

compiler/modules/front/const.bal
```diff
@@ -200,22 +200,52 @@ function foldBinaryEqualityExpr(FoldContext cx, t:SemType? expectedType, s:Binar
...
-        return <s:ConstValueExpr> { value: positive == (left === right), multiSemType: t:BOOLEAN };
+        return foldedBinaryConstExpr(value, t:BOOLEAN, leftExpr, rightExpr);
     }
     expr.left = leftExpr;
     expr.right = rightExpr;
     return expr;
 }
 
+// XXX remove after JBUG #17977 is fixed
+function isEqual(SimpleConst c1, SimpleConst c2) returns boolean{
```
manuranga, line 221: missing space

jclark: I fixed this

Issue #167, opened 2021-07-02
Implement minimal subset of `error` type


jclark:
This involves:

- support error values
- new type descriptor `error`
- new expression syntax `error(E)`
- `T|error` and `error?` as return values only (including for `main`)
-  `panic E;` statement

Follow ons are
- `check E` , `checkpanic E`  expression #406
- langlib `error:message` function #407

This needs #130.

This would be more useful if we have the `is` operator #144 and unions generally.

jclark:
This would allow us to implement string:fromCodePointInt

jclark:
This will require arrays having an inherent type to distinguish `(any|error)[]` and `any[]`, because this

```
any x = [];
```

creates an array with inherent type `(any|error)[]`.

So we could also add an explicit type `any|error`.

jclark:
I removed `fail`: it's pointless (equivalent to `return`) in the absence on `on fail`.

jclark:
Things to do:

- [x] Runtime support for error values #376 
- [x] BIR ErrorConstructInsn for creating error values #377
- [x] Make panicking use error values #379
- [x] Implement error constructor  #386
- [x] Implement `panic` statement #385
- [x] Add `error` type descriptor #387
- [x] Make type test  `is error` work
- [x] Make type cast `<error>` work
- [x] Allow unions with error
- [x] Make sure `===` works

Related
- Add libbacktrace as third-party library #367
- Add libbacktrace to build system #390
- Snapshot stack when creating error value #46 
- Print out backtrace in error value on panic #6

jclark:
Moved remaining parts into separate issues.

PR #331, opened 2021-08-16
[WIP] Implement simple garbage collector

### LINKED ISSUE #165: Identify GC roots on the stack
Based on #149 and #164, we need to build the capability to identify at runtime the GC roots on the stack.

Concretely, we want to add a C file to the runtime directory that provides a `markRoots` function that can be used by our garbage collector:

```
typedef char *Root;
typedef void (*AddRootFunctionPtr)(Root *root, void *arg);
void markRoots(AddRootFunctionPtr func, void *arg);
```

Not sure if this is the best API. We should probably look MMTK and figure out something that is a good fit.

### LINKED ISSUE #300: Implement simple garbage collector
This follows on from #165.

I think the best approach at this stage is a Cheney-style semispace copying collector, which is described here https://en.wikipedia.org/wiki/Cheney%27s_algorithm

It's very elegant and simple.

Concretely:

1. use mmap to allocate the heap on startup
2.  heap will be page aligned
3. I suggest a default size (including both halves) of 2Gb and multiple of page size; it should be possible to override this with an environment variable
4. allocation rounds up to a multiple of 8
5. each allocation starts with a 64-bit word that usually holds the size in bytes, rounded up, but during collection can also hold the forwarding pointer
6. we need to be able to tell when we have a forwarding pointer; since pointers and sizes are multiples of 8, we can just | with 1 for one of the cases
7. when scanning the stack, if there are derived pointers, we need to be able to find the base pointer i.e. in the derived pointer case, the marking function needs to get an extra argument that is the base pointer
8. when it does an allocation, if the heap is full, it will run a collection; if that doesn't free enough space, then it aborts
9. in between collections, we should mprotect (as PROT_NONE) the semispace that is not being used in order to detect bugs

KavinduZoysa:
## Purpose
Implement simple garbage collector.
Resolves #300
Resolved #165


runtime/gc_cheney.c
```diff
@@ -0,0 +1,69 @@
...
+void collect() {
+    printf("%s\n", "Collect");
+    get_roots();
+    // get_roots() -> calls function in gc-roots, should pass rsp as parameter
+    // swap from_space <-> to_space
+    // should update heap_limit_ptr
+}
+
+UntypedPtr _bal_alloc(uint64_t nBytes) {
+    nBytes = nBytes + HEAP_HEADER_SIZE;
```
jclark, line 59: You need to round up nBytes to a multiple of 8.

runtime/gc_cheney.c
```diff
@@ -0,0 +1,69 @@
...
+    to_space_ptr = mmap(NULL, heap_half_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, 0, 0);
+    if (to_space_ptr == MAP_FAILED) {
+        printf("Heap initialization Failed\n");
+        abort();
+    }
+}
+
+void collect() {
+    printf("%s\n", "Collect");
+    get_roots();
```
jclark, line 52: get_roots needs to take a function pointer that it will call for each root found

runtime/gc_cheney.c
```diff
@@ -0,0 +1,69 @@
...
+uint8_t *alloc_ptr;
+uint8_t *scan_ptr;
+uint8_t *heap_limit_ptr;
+
+extern void get_roots();
+
+void _bal_init_heap() {
+    int page_size = getpagesize();
+    const char *heap_env = getenv("BAL_HEAP");
+    uint64_t heap_half_size = (heap_env != NULL) ? atoi(heap_env) : DEFAULT_HEAP_HALF_SIZE;
```
jclark, line 30: Need to check return of atoi

jclark (approved):
Why do we need libbacktrace?

jclark:
Use however big a heap you need to get the tests to run.  We can reduce when GC is working.

runtime/gc_cheney.c
KavinduZoysa: Fixed in https://github.com/ballerina-platform/nballerina/commit/8964a5ea73124ee8d2132121b66b78e9f0be39a4

runtime/gc_cheney.c
KavinduZoysa: Fixed in https://github.com/ballerina-platform/nballerina/commit/8964a5ea73124ee8d2132121b66b78e9f0be39a4

KavinduZoysa:
> Why do we need libbacktrace?

It is used to get the stack trace to find GC roots. https://github.com/ballerina-platform/nballerina/blob/8964a5ea73124ee8d2132121b66b78e9f0be39a4/runtime/gc-roots/roots.c#L50

KavinduZoysa:
> Use however big a heap you need to get the tests to run. We can reduce when GC is working.

Okay, if the heap half size is 1GB, two tests are failing (Vbenchmap and Vbenchmergesort), I have increased it it 3GB. Now all tests are passing.

jclark:
> > Why do we need libbacktrace?
> 
> It is used to get the stack trace to find GC roots.

But I thought statepoint-utils was doing a GC without needing libbacktrace

jclark:
It's not a problem that you are using libbacktrace, I just don't understand.

Also bear in mind that I will want to do #167 soon, which will need libbacktrace. This may happen before you are done with GC.

KavinduZoysa:
> But I thought statepoint-utils was doing a GC without needing libbacktrace

> It's not a problem that you are using libbacktrace, I just don't understand.
> 

The `statepoint-utils` goes through the stack using a method that depends on the architecture, i.e. at the end of each stack frame(stack pointer), it keeps the return address of the previous function. So what `statepoint-utils` does is adding the frame size to stack pointer and get the value of that location. Since this method depends on the architecture, I use libbackatrace to get the return address of each frame.

> Also bear in mind that I will want to do #167 soon, which will need libbacktrace. This may happen before you are done with GC.

Sure, will do

Issue #310, opened 2021-08-14
Implement constReal in jni.llvm


jclark:
Catch up with print.llvm changes in #306.

Needed for #262.

@heshanpadmasiri Can you look at how it decides whether to use hex vs decimal formats for float?

heshanpadmasiri:
Documentation says "the only time hexadecimal floating-point constants are required (and the only time that they are generated by the disassembler) is when a floating-point constant must be emitted but it cannot be represented as a decimal floating-point number in a reasonable number of digits". Based on my testing this "reasonable number of digits" is 7 (ex :1.12345 => 1.123450e+00 will be decimal float but 1.23456 will be hex).
It always uses hex for special values such as

- Infinity = 0x7FF0000000000000
- Nan = 0x7FF8000000000000

jclark:
@heshanpadmasiri Thanks. Please take #307 and fix to be consistent with jni.llvm.

PR #333, opened 2021-08-16
Implement constReal in jni.llvm

### LINKED ISSUE #307: Use hexadecimal for LLVM floating point constants
LLVM requires the use of hexadecimal format (which is a hex integer representing the IEEE format bits) in some cases:

> For example, the form ‘double    0x432ff973cafa8000’ is equivalent to (but harder to read than) ‘double 4.5e+15’. The only time hexadecimal floating-point constants are required (and the only time that they are generated by the disassembler) is when a floating-point constant must be emitted but it cannot be represented as a decimal floating-point number in a reasonable number of digits. For example, NaN’s, infinities, and other special values are represented in their IEEE hexadecimal format so that assembly and disassembly do not cause any bits to change in the constants.

The simplest correct solution is to always use hex, but that will reduce the human readability of the IR.  One approach is to convert finite (not NaN or infinity) values to a string, then parse back into a float, and check whether the bits are the same (using float:toBitsInt).

heshanpadmasiri:
## Purpose

- Implement constReal in jni.llvm
- Use hex / decimal for LLVM floating point constants in print.llvm

Resolves #307 #310

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/print.llvm/llvm.bal
+++ b/compiler/modules/print.llvm/llvm.bal
@@ -69,9 +69,9 @@ public function constInt(IntType ty, int val) returns ConstValue {
     return new ConstValue(ty, val.toString());
 }
 
-final string pINF = (1.0/0.0).toBitsInt().toHexString().toUpperAscii();
-final string nINF = (-1.0/0.0).toBitsInt().toHexString().toUpperAscii();
-final string NAN = (0.0/0.0).toBitsInt().toHexString().toUpperAscii();
+final string posInf = "0x" + (1.0/0.0).toBitsInt().toHexString().toUpperAscii();
+final string negInf = "0x" + (-1.0/0.0).toBitsInt().toHexString().toUpperAscii();
+final string NAN = "0x" + (0.0/0.0).toBitsInt().toHexString().toUpperAscii();
 
 // Corresponds to LLVMConstReal
 public function constReal(RealType ty, float val) returns ConstValue {
@@ -79,14 +79,14 @@ public function constReal(RealType ty, float val) returns ConstValue {
     // Special cases
     if val.isInfinite() {
         if val > 0.0 {
-            valRep = "0x" + pINF;
+            valRep = posInf;
         }
         else {
-            valRep = "0x" + nINF;
+            valRep = negInf;
         }
     }
     else if val.isNaN() {
-        valRep = "0x" + NAN;
+        valRep = NAN;
     }
     else {
         // General case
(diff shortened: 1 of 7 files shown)


compiler/modules/print.llvm/llvm.bal
```diff
@@ -71,9 +71,53 @@ public function constInt(IntType ty, int val) returns ConstValue {
 
 // Corresponds to LLVMConstReal
 public function constReal(RealType ty, float val) returns ConstValue {
-    // XXX #307 use decimal for non-special floats 
-    return new ConstValue(ty, "0x" + int:toHexString(float:toBitsInt(val)));
-
+    string valRep;
+    if val.isInfinite() {
+        if val > 0.0 {
+            valRep = "0x7FF0000000000000";
```
jclark: Clearer to do use a final with something like (1.0/0.0).toBitsInt().toHexString().

compiler/modules/print.llvm/llvm.bal
```diff
@@ -71,9 +71,53 @@ public function constInt(IntType ty, int val) returns ConstValue {
...
+        }
+        else {
+            valRep = "0xFFF0000000000000";
+        }
+    }
+    else if val.isNaN() {
+        valRep = "0x7FF8000000000000";
+    }
+    else {
+        float exp = float:floor(float:log10(float:abs(val)));
```
jclark: This code is far too difficult to understand. I have no idea what is doing.  It either needs a lot of comments or, probably better, do it in a simpler way. I don't understand why it needs to be this complicated.  Can you not convert the float to a string and then based on the string decide whether to use that or the hexadecimal representation?

jclark (changes requested):

compiler/modules/print.llvm/llvm.bal

compiler/modules/print.llvm/llvm.bal
heshanpadmasiri: Moved the part which decide whether to use hex or not to the top and added comments to clarify what each thing is doing. Most of it is to convert the number to scientific notation in the format we need if we are not using the hex representation.

compiler/modules/print.llvm/llvm.bal
```diff
@@ -69,11 +69,68 @@ public function constInt(IntType ty, int val) returns ConstValue {
...
+    else {
+        // General case
+        if float:abs(val).toString().length() > 7 {
+            // Resulting number has more than 7 digits so convert to hex format
+            int repInt = float:toBitsInt(val);
+            valRep = "0x" + int:toHexString(repInt).toUpperAscii();
+        }
+        else {
+            // Convert the number to scientific notation
+            float exp = float:floor(float:log10(float:abs(val))); // Value of the exponent
```
jclark: Why can't we use the result of val.toString() here?

jclark (changes requested):

compiler/modules/print.llvm/llvm.bal
heshanpadmasiri: It doesn't convert the value to scientific notation it just output the value as a string for example for 11.2345 it will output "11.2345" instead of "1.123450e+01" (which is the output for llvm). We can use the string output and generate the value using a combination of bunch of splits and concats but I am not sure whether it will be more easier to understand specially given we'll have lot of special cases like values less than zero (ex 0.001) and large ones.

jclark: So LLVM is using printf `e` format and you are trying to match that, right?

The right way to solve this is to add a toExpString function to lang.float

https://github.com/ballerina-platform/ballerina-spec/issues/770#issuecomment-791258822

We don't have that yet, so we can either

1. not try to match LLVM output in this respect,
2. create a module with some library functions and implement toExpString in that

If we do 2, we should just use the native Java equivalent (which is String.format, I think)

compiler/modules/print.llvm/llvm.bal
```diff
@@ -69,11 +69,68 @@ public function constInt(IntType ty, int val) returns ConstValue {
     return new ConstValue(ty, val.toString());
 }
 
+final string pINF = (1.0/0.0).toBitsInt().toHexString().toUpperAscii();
```
jclark: Include the "0x" in these.

pINF is not camel case. I suggest posInf or positiveInfinity.

jclark (changes requested):
Main poiont is https://github.com/ballerina-platform/nballerina/pull/333/files#r690020835

compiler/modules/print.llvm/llvm.bal

compiler/modules/print.llvm/llvm.bal
heshanpadmasiri: Fixed by option 1 because it meets the objective of being easier to read and compiles correctly (though not exactly equal to LLVM output) while remaining purely ballerina (unlike `jni.llvm`).

jclark (approved):
LGTM

PR #334, opened 2021-08-16
Fix the precision of float printing

### LINKED ISSUE #308: Float to string conversion should use just enough digits to represent the correct IEEE value
Currently using printf %g which does not do the right thing.

The classic paper on this is the Steele/White one:

https://lists.gnu.org/archive/html/gcl-devel/2012-10/pdfkieTlklRzN.pdf

Required for #262

manuranga:
Fix the precision of float printing 
Fixes #308

### DIFF
--- a/runtime/print.c
+++ b/runtime/print.c
@@ -1,6 +1,8 @@
 #include "balrt.h"
 #include <stdio.h>
 #include <stdlib.h>
+#include <math.h>
+#include "third-party/dtoa/emyg_dtoa.h"
 
 #define STYLE_DIRECT 0
 #define STYLE_INFORMAL 1
@@ -37,6 +39,7 @@ static void printStringLiteralChar(FILE *fp, int c) {
 }
 
 static void printTagged(FILE *fp, TaggedPtr p, int style, struct PrintStack *stackPtr) {
+    double d;
     int tag = getTag(p);
     switch (tag & UT_MASK) {
         case 0:
@@ -55,8 +58,22 @@ static void printTagged(FILE *fp, TaggedPtr p, int style, struct PrintStack *sta
             fprintf(fp, "%" PRId64, taggedToInt(p));
             break;
         case TAG_FLOAT:
-            // XXX fix precision
-            fprintf(fp, "%g", taggedToFloat(p));
+            d = taggedToFloat(p);
+            if (isnan(d)) {
+                fprintf(fp, "NaN");
+                break;
+            }
+            if (isinf(d)) {
+                fprintf(fp, d < 0.0l ? "-Infinity" : "Infinity");
+                break;
+            }
+            if (d == 0.0) {
+                fprintf(fp, signbit(d) ? "-0.0" : "0.0");
+                break;
+            }
+            char buf[EMYG_DTOA_BUFFER_LEN];
+            emyg_dtoa_non_special(d, buf);
+            fprintf(fp, "%s", buf);
             break;
         case TAG_LIST_RW:
             if (stackContains(stackPtr, p)) {
--- a/runtime/tests/Makefile
+++ b/runtime/tests/Makefile
@@ -1,4 +1,4 @@
-TESTPROGS=string_hash string_cmp string_concat string_len immediate_string tagged_ptr
+TESTPROGS=string_hash string_cmp string_concat string_len immediate_string tagged_ptr float_print
 # following will test balrt_inline, and must be named inlinert_*
 TESTINLINERTPROGS=inlinert_float
 TESTPROG_STAMPS=$(addsuffix .stamp, $(TESTPROGS) $(TESTINLINERTPROGS))
--- a/compiler/testSuite/Vfloat00.bal
+++ b/compiler/testSuite/Vfloat00.bal
@@ -2,4 +2,19 @@ import ballerina/io;
 public function main() {
     float f = 0.1;
     io:println(f); // @output 0.1
+
+    f = 1.0 / 3.0;
+    io:println(f); // @output 0.3333333333333333
+
+    io:println(0.0); // @output 0.0
+
+    io:println(-0.0); // @output -0.0
+
+    io:println(-2.225073858507202E-308); // @output -2.225073858507202e-308
+
+    io:println(0.0 / 0.0); // @output NaN
+
+    io:println(1.0 / 0.0); // @output Infinity
+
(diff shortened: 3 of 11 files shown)


manuranga:
Should I do some round trip testing ? directly testing `emyg_dtoa` should be straightforward, but if I were to test `printTagged`, I'll need an in-memory file descriptor via a `pipe` or `fmemopen`.

jclark:
Eventually we will want to implement toString and implement println in terms of that, at which point we can test this. Maybe create an issue to remember to do this?

runtime/third-party/emyg/emyg_dtoa.h
```diff
@@ -0,0 +1,7 @@
+// from https://github.com/miloyip/dtoa-benchmark/blob/bf1fb58/src/emyg/emyg_dtoa.h
+/* emyg_dtoa.h */
+
+void emyg_dtoa (double value, char* buffer);
+
+// appended for nballerina use
+void emyg_dtoa_non_special (double value, char* buffer);
```
jclark: Can you explain why we need this?

runtime/print.c
```diff
@@ -55,8 +58,22 @@ static void printTagged(FILE *fp, TaggedPtr p, int style, struct PrintStack *sta
...
+            }
+            if (isinf(d)) {
+                fprintf(fp, d < 0.0l ? "-Infinity" : "Infinity");
+                break;
+            }
+            if (d == 0.0) {
+                fprintf(fp, signbit(d) ? "-0.0" : "0.0");
+                break;
+            }
+            char buf[30]; // buf[25] should be enough, 30 just to be safe
```
jclark: How do you know 25 is enough?

runtime/third-party/emyg/emyg_dtoa.h
manuranga: `emyg_dtoa` doesn't handle NaN, infinity and -0.0, so those checks become redundant. Using it would still give the correct result.
This just avoids extra checks.

runtime/print.c
manuranga: I tested a few million random doubles with `emyg_dtoa`, max len was 25.

It was not specified, but they have [tested everything with 17 significant digits](https://github.com/miloyip/dtoa-benchmark#procedure). 

Wikipidea says:
> The 53-bit significand precision gives from 15 to 17 significant decimal digits precision. If a decimal string with at most 15 significant digits is converted to IEEE 754 double-precision representation, and then converted back to a decimal string with the same number of digits, the final result should match the original string.

1 sign + 17 digits + 1 `.` + 1 `e` + 1 exp sign + 3 exp = 24 so it makes sense.

jclark: I am astounded that this is not documented in the source, which makes me question the dependability of this implementation.

jclark: The benchmark testing code appears to use a buffer size of 1024.

jclark:
Did you test with valgrind?

manuranga:
Just ran valgrind on `4194304` random doubles round-trip test, no leaks detected. Any buffer length less than 25 segfaults.

runtime/third-party/emyg/emyg_dtoa.c
```diff
@@ -0,0 +1,452 @@
...
+		uint64_t tmp = ((uint64_t )(p1) << -one.e) + p2;
+		if (tmp <= delta) {
+			*K += kappa;
+			GrisuRound(buffer, *len, delta, tmp, (uint64_t )(kPow10[kappa]) << -one.e, wp_w.f);
+			return;
+		}
+	}
+
+	// kappa = 0
+	for (;;) {
```
manuranga: Looking at [this paper's `Figure 6` line 8](https://www.cs.tufts.edu/~nr/cs257/archive/florian-loitsch/printf.pdf), I modified this line as below
```suggestion
	while ((*len) < 19) 
```
This should guarantee that there will be no buffer overruns. I see no lose of precision in round-trip test.

jclark: I think it's a good idea to limit *len in that loop.  That's the one place where it's not obvious how many characters can be added to the buffer.

But as far as I understand the algorithm that limit should never be exceeded. So I think we should add `if ((*len) >= 19) abort();`. 

For our buffer limit, I think we should use a multiple of 8 since our stack is 8-byte aligned. If we use 40, it is very obvious that the buffer cannot overflow.

jclark:
We need to do #335 also.

manuranga:
I think it's better add round-trip tests for `emyg_dtoa` now that we are doing modifications.

jclark:
> I think it's better add round-trip tests for `emyg_dtoa` now that we are doing modifications

I expect we will implement float:fromString using strtod, so I suggest checking that our modified dtoa roudntrips with strtod.

jclark:
I am waiting for #335 to be done first.

jclark:
I am guessing this still needs to be fixed to use #335.

runtime/tests/Makefile
```diff
@@ -1,4 +1,4 @@
-TESTPROGS=string_hash string_cmp string_concat string_len immediate_string tagged_ptr
+TESTPROGS=string_hash string_cmp string_concat string_len immediate_string tagged_ptr float_print
```
jclark, line 1: Also needs adding to .gitignore

runtime/print.c
```diff
@@ -55,8 +58,22 @@ static void printTagged(FILE *fp, TaggedPtr p, int style, struct PrintStack *sta
...
+            }
+            if (isinf(d)) {
+                fprintf(fp, d < 0.0l ? "-Infinity" : "Infinity");
+                break;
+            }
+            if (d == 0.0) {
+                fprintf(fp, signbit(d) ? "-0.0" : "0.0");
+                break;
+            }
+            char buf[40]; // buf[25] should be enough, 40 just to be safe, and align
```
jclark: There should be a #define for 40 in emyg_dtoa.h

jclark (changes requested):

runtime/tests/Makefile
manuranga: Fixed. May I create an issue to build all artifacts in a dir like `target` or `build`?

jclark: By all means create an issue, but I think it's better to solve this later:

- we need `bal` command to allow control over `target` dir
- we will want to switch to `cmake` eventually, which will support this for free

PR #342, opened 2021-08-18
Implement float binary arithmetic operators

### LINKED ISSUE #319: Implement FloatArithmeticBinary BIR insn
This issue completes the implementation of binary arithmetic operators for float type.

Front-end has untested support for this.

LLVM support for this is #312.

This requires solving #313.

This is part of #262.

manuranga:
Implement float binary arithmetic operators.

Tests depend on  #334.

Fixes #319


jclark (changes requested):
I think `Float` instead of `BinaryFp` will work best.

PR #343, opened 2021-08-18
Implement hexadecimal floating point literals 

### LINKED ISSUE #326: Implement hexadecimal floating point literals
These are documented here:

https://ballerina.io/spec/lang/2021R1/#HexFloatingPointLiteral

They can be parsed with float:fromHexString.

ushirask:
Resolves #326


ushirask:
Due to bug with `float:fromHexString` this will currently throw `error: invalid float literal` for hex float literals

compiler/modules/front.syntax/scan.bal
```diff
@@ -642,40 +642,32 @@ function scanNumericEscape(int[] codePoints, int startIndex) returns int? {
...
-            |CP_UPPER_F
-            |CP_LOWER_A
-            |CP_LOWER_B
-            |CP_LOWER_C
-            |CP_LOWER_D
-            |CP_LOWER_E
-            |CP_LOWER_F => {
-                i += 1;
-                continue;
+    if i !=() {
```
jclark, line 648: Use match.

compiler/modules/front.syntax/scan.bal
```diff
@@ -764,6 +766,49 @@ function scanExponent(int[] codePoints, int startIndex) returns int? {
     return ();
 }
 
+function scanHexFractionExponent(int[] codePoints, int startIndex) returns int? {
```
jclark, line 769: The common code between decimal and hex floating point should be factored out.

jclark (commented):
Given that `float:fromHexString` isn't working yet, and given the amount of work that this patch still needs, I think you should stop working on this.

PR #344, opened 2021-08-19
Add tests for const float expressions

### LINKED ISSUE #316: Add tests for const float expressions
const expressions involving floats should work, but they need tests

e.g.

```
const X = 1f + 1f;
```

ushirask:
Resolves #316


jclark (changes requested):
These need to handle the difficult, error prone cases specifically

- infinities
- positive/negative zero
- NaN

PR #348, opened 2021-08-21
Implement float to int in LLVM IR

### LINKED ISSUE #288: Implement conversions between `float` and `int` in runtime
Implement conversions between `float` and `int` in runtime
Part of #262

manuranga:
Implement float to int in LLVM IR
Fixes #288

### DIFF
--- a/runtime/balrt_inline.h
+++ b/runtime/balrt_inline.h
@@ -1,4 +1,12 @@
 #include <stdbool.h>
 
+struct FloatToIntResult {
+    uint64_t value;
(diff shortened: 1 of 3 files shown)


runtime/balrt_inline.h
```diff
@@ -1,4 +1,10 @@
 #include <stdbool.h>
 
+struct FloatToIntResult {
+    long value;
```
jclark: uint4_t

jclark (changes requested):

runtime/balrt_inline.h
manuranga: Fixed

PR #354, opened 2021-08-23
Allow type descriptor in constant declarations

### LINKED ISSUE #293: Allow type descriptor in constant declarations
Const declarations should be able to specify a type descriptor so as to provide the expression with a contextually expected type.

manuranga:
Allow type descriptor in constant declarations
Fixes #293

### DIFF
--- a/compiler/modules/front.syntax/ast.bal
+++ b/compiler/modules/front.syntax/ast.bal
@@ -31,6 +31,7 @@ public type FunctionDefn record {|
 public type ResolvedConst readonly & [t:SemType, t:Value];
 public type ConstDefn record {|
     readonly string name;
+    InlineTypeDesc? td;
     Visibility vis;
     Expr expr;
     err:Position pos;
--- a/compiler/modules/front/const.bal
+++ b/compiler/modules/front/const.bal
@@ -62,11 +62,19 @@ function resolveConstDefn(ModuleTable mod, s:ConstDefn defn) returns s:ResolvedC
     else {
         defn.resolved = false;
         ConstFoldContext cx = new ConstFoldContext(defn.name, mod);
-        s:Expr expr = check foldExpr(cx, (), defn.expr);
+        s:InlineTypeDesc? td = defn.td;
+        // JBUG remove cast and use == after beta 3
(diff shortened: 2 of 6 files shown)


compiler/modules/front.syntax/ast.bal
```diff
@@ -31,6 +31,7 @@ public type FunctionDefn record {|
 public type ResolvedConst readonly & [t:SemType, t:Value];
 public type ConstDefn record {|
     readonly string name;
+    InlineBasicTypeDesc? td;
```
jclark: Slight preference for using InlineTypeDesc here.

compiler/modules/front/const.bal
```diff
@@ -51,7 +51,9 @@ function resolveConstDefn(ModuleTable mod, s:ConstDefn defn) returns s:ResolvedC
     else {
         defn.resolved = false;
         ConstFoldContext cx = new ConstFoldContext(defn.name, mod);
-        s:Expr expr = check foldExpr(cx, (), defn.expr);
+        s:InlineBasicTypeDesc? td = defn.td;
+        t:SemType? expectedType = td == () ? () : s:resolveInlineTypeDesc(td);
+        s:Expr expr = check foldExpr(cx, expectedType, defn.expr);
```
jclark, line 68: You also need to check that the value does belong to the type. For example, this is illegal

```
const boolean b = 17;
```

jclark (changes requested):

compiler/modules/front/const.bal
```diff
@@ -62,11 +62,19 @@ function resolveConstDefn(ModuleTable mod, s:ConstDefn defn) returns s:ResolvedC
...
-            s:ResolvedConst r = [t:singleton(expr.value), { value: expr.value }];
-            defn.resolved = r;
-            return r;
+            if expectedType == () || t:containsConst(expectedType, expr.value) {
+                s:ResolvedConst r = [t:singleton(expr.value), { value: expr.value }];
+                defn.resolved = r;
+                return r;
+            }
+            else {
+                return err:semantic(`initializer of ${defn.name} is not a subtype of the contextually expected type`, defn.pos);
```
jclark: `not a subtype of the declared type`

jclark (changes requested):

PR #360, opened 2021-08-24
Use ptrmask intrinsic to implement taggedToPtr

### LINKED ISSUE #201: Make runtime library use ptrmask intrinsic for untagging pointers
At the moment, compiler generates code that uses ptrmask intrinsic for removing the tag from a tagged pointer. However, the runtime library uses inttoptr and ptrtoint, which is not ideal with address spaces, and may not work at all when address space 1 is non-integral (#200).

We can fix this with a hand-written .ll file.

KavinduZoysa:
## Purpose
Fixes #201

### DIFF
--- a/runtime/Makefile
+++ b/runtime/Makefile
@@ -2,19 +2,20 @@ LLVM_SUFFIX?=-11
 export CLANG ?= clang$(LLVM_SUFFIX)
 LLVM_AS ?= llvm-as$(LLVM_SUFFIX)
 LLC ?= llc$(LLVM_SUFFIX)
+OPT ?= opt$(LLVM_SUFFIX)
 export LLVM_LINK ?= llvm-link$(LLVM_SUFFIX)
 VALGRIND ?= valgrind
 OBJS=alloc.o list.o main.o mapping.o panic.o print.o string.o int.o third-party/dtoa/emyg_dtoa.o
-RT_BCS=alloc.bc list.bc main.bc mapping.bc panic.bc print.bc string.bc int.bc third-party/dtoa/emyg_dtoa.bc
-RT_LINKED_BCS=alloc_linked.bc list_linked.bc main_linked.bc mapping_linked.bc panic_linked.bc print_linked.bc string_linked.bc int_linked.bc third-party/dtoa/emyg_dtoa_linked.bc
+RT_OPT_BCS=alloc_opt.bc list_opt.bc main_opt.bc mapping_opt.bc panic_opt.bc print_opt.bc string_opt.bc int_opt.bc third-party/dtoa/emyg_dtoa_opt.bc
 BCS=eq_inline.bc int_inline.bc list_inline.bc mapping_inline.bc string_inline.bc float_ir_inline.bc float_inline.bc
-TAGGED_PTR_BC=tagged_ptr.bc
 INCLUDES=tag.h balrt.h balrt_inline.h
+TAGGED_PTR_BC=tagged_ptr.bc
+TAGGED_PTR_OBJ=tagged_ptr.o
 AR=ar
 export LIB=balrt.a
 export BCLIB=balrt_inline.bc
 
-all: $(TAGGED_PTR_BC) $(RT_BCS) $(RT_LINKED_BCS) $(LIB) $(BCLIB)
+all: $(TAGGED_PTR_BC) $(RT_OPT_BCS) $(LIB) $(BCLIB)
 	$(MAKE) -C tests all
 
 test: all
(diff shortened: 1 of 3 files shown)


runtime/Makefile
```diff
@@ -20,7 +20,7 @@ test: all
 testCoverage: $(LIB) $(BCLIB)
 	$(MAKE) -C tests testCoverage
 
-$(LIB): $(OBJS)
+$(LIB): $(TAGGED_PTR_OBJ) $(OBJS)
```
jclark: We want taggedToPtr to be inlined by the C file.

This means we need to do

- compile .c to .bc
- llvm-link the .bc with tagged-ptr.ll
- llc the result into .o

jclark (changes requested):

runtime/Makefile
KavinduZoysa: With the change de1dc58 I am getting an error as shown below. I am currently looking into that.

```
clang-11  -o ./string_hash string_hash.c ../balrt.a
/usr/bin/ld: ../balrt.a: error adding symbols: archive has no index; run ranlib to add one
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

KavinduZoysa: Fixed that issue in be597b8

runtime/Makefile
```diff
@@ -1,17 +1,20 @@
 LLVM_SUFFIX?=-11
 export CLANG ?= clang$(LLVM_SUFFIX)
 LLVM_AS ?= llvm-as$(LLVM_SUFFIX)
+LLC ?= llc$(LLVM_SUFFIX)
 export LLVM_LINK ?= llvm-link$(LLVM_SUFFIX)
 VALGRIND ?= valgrind
 OBJS=alloc.o list.o main.o mapping.o panic.o print.o string.o int.o third-party/dtoa/emyg_dtoa.o
+RT_BCS=alloc.bc list.bc main.bc mapping.bc panic.bc print.bc string.bc int.bc third-party/dtoa/emyg_dtoa.bc
```
manuranga: Ideally we shouldn't need to all the files again.

KavinduZoysa: I removed as much as possible, and removing rest of the files occur errors when running make.

runtime/tagged_ptr.ll
```diff
@@ -0,0 +1,6 @@
+declare i8 addrspace(1)* @llvm.ptrmask.p1i8.i64(i8 addrspace(1)*, i64) readnone speculatable
+
+define dso_local i8 addrspace(1)* @taggedToPtr(i8 addrspace(1)* %0) {
```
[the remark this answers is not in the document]
KavinduZoysa: Fixed

runtime/Makefile
```diff
@@ -20,14 +24,23 @@ test: all
...
+$(TAGGED_PTR_OBJ):
+	$(CLANG) -c -o $@ $(TAGGED_PTR_BC)
+
 $(BCLIB): $(BCS)
 	$(LLVM_LINK) -o - -S $^ | sed -e '/define .*@_[Bb][a-zA-Z]/s/^define /define linkonce_odr /' -e '/target datalayout/ s/"$$/-ni:1"/' | $(LLVM_AS) >$@
 
-%.o: %.c
-	$(CLANG) $(CFLAGS) -c -o $@ $<
+%.o: %_opt.bc
+	$(LLC) -O2 --filetype=obj -o $@ $<
```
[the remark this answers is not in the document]
KavinduZoysa: Changed

runtime/Makefile
```diff
@@ -1,17 +1,21 @@
 LLVM_SUFFIX?=-11
 export CLANG ?= clang$(LLVM_SUFFIX)
 LLVM_AS ?= llvm-as$(LLVM_SUFFIX)
+LLC ?= llc$(LLVM_SUFFIX)
+OPT ?= opt$(LLVM_SUFFIX)
 export LLVM_LINK ?= llvm-link$(LLVM_SUFFIX)
 VALGRIND ?= valgrind
 OBJS=alloc.o list.o main.o mapping.o panic.o print.o string.o int.o third-party/dtoa/emyg_dtoa.o
+RT_OPT_BCS=alloc_opt.bc list_opt.bc main_opt.bc mapping_opt.bc panic_opt.bc print_opt.bc string_opt.bc int_opt.bc third-party/dtoa/emyg_dtoa_opt.bc
```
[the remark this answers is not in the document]
KavinduZoysa: Using pipe with stdin/out causes to remove these intermediate file names.

runtime/Makefile
```diff
@@ -27,7 +28,11 @@ $(BCLIB): $(BCS)
 	$(LLVM_LINK) -o - -S $^ | sed -e '/define .*@_[Bb][a-zA-Z]/s/^define /define linkonce_odr /' -e '/target datalayout/ s/"$$/-ni:1"/' | $(LLVM_AS) >$@
 
 %.o: %.c
-	$(CLANG) $(CFLAGS) -c -o $@ $<
+	$(CLANG) -c -O2 -emit-llvm -o - $< | \
+	$(LLVM_LINK) tagged_ptr.ll - -o - -S | \
+	$(OPT) -O2 - -o - -S | \
```
jclark, line 38: Why do you need opt separate from llc?

jclark: I don't understand. Won't llc do all the necessary optimizations? Why do we need to do "target-independent optimization" separately?

KavinduZoysa: As I understood it is like this. clang -O2 does all target-independent and target-dependent optimizations. opt -O2 does target-independent optimizations and llc -O2 does target-dependant optimization. 

If we run `clang-11 -O2 -mllvm --debug-pass=Arguments test.c` we can print all passes that were executed. Then if we run `opt-11 -O2 --debug-pass=Arguments test.ll -S` and `llc-11 -O2 --debug-pass=Arguments test.bc` separately we can see passes executed for those commands, and this output is mostly the same to with clang command output.

In this PR, since I removed clang -O2, I added both opt -O2 and llc -O2 to replace clang -O2.

https://groups.google.com/g/llvm-dev/c/b1ckC_hini4/m/rAkhV18bAQAJ, this is not the exact source I learned this.

jclark: IIRC clang can accept .ll or .bc as input, so maybe the simplest thing is to use clang again instead of opt/llc (i.e. clang -emit-llvm then llvm-link then clang again)

KavinduZoysa: To enable GC for runtime in #300, we have to run statepoint generation pass with opt. Therefore shall we keep it like this?

jclark: That's a good point. Clang has an optimization pipeline that is carefully crafted to give good results with C, so it would be good to take advantage of that. Maybe a combo of opt + clang would work?

KavinduZoysa: Yes, changed it.

PR #363, opened 2021-08-25
Implement float/int conversions in backend 

### LINKED ISSUE #323: Implement BIR instructions for float/int conversion
Issue #321 will introduce new BIR instructions for float/int conversion.

These need to be implemented in the backend using #288.

This issue should wait for #321.

This is part of #262.

manuranga:
Implement float/int conversions in backend
Fixes  #323

### DIFF
--- a/compiler/modules/front/codeGen.bal
+++ b/compiler/modules/front/codeGen.bal
@@ -1113,12 +1113,10 @@ function codeGenTypeCast(CodeGenContext cx, bir:BasicBlock bb, Environment env, 
     var reg = <bir:Register>operand; // const folding should have got rid of the const
     t:SemType toType = tcExpr.semType;
     t:UniformTypeBitSet? toNumType = t:singleNumericType(toType);
-    if toNumType != ()
-       // Remove the next line to enable conversion when support in backend is implemented
-       && toNumType == t:NEVER 
-       && !t:isSubtypeSimple(t:intersect(reg.semType, t:NUMBER), toNumType) {
+    if toNumType != () && !t:isSubtypeSimple(t:intersect(reg.semType, t:NUMBER), toNumType) {
         toType = t:diff(toType, t:diff(t:NUMBER, toNumType));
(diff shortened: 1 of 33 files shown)


compiler/modules/front/codeGen.bal
```diff
@@ -1113,12 +1113,10 @@ function codeGenTypeCast(CodeGenContext cx, bir:BasicBlock bb, Environment env,
...
     t:UniformTypeBitSet? toNumType = t:singleNumericType(toType);
-    if toNumType != ()
-       // Remove the next line to enable conversion when support in backend is implemented
-       && toNumType == t:NEVER 
-       && !t:isSubtypeSimple(t:intersect(reg.semType, t:NUMBER), toNumType) {
+    if toNumType != () && !t:isSubtypeSimple(t:intersect(reg.semType, t:NUMBER), toNumType) {
         toType = t:diff(toType, t:diff(t:NUMBER, toNumType));
-        bir:Register result = cx.createRegister(t:intersect(reg.semType, toType));
+        
+        bir:Register result = cx.createRegister(t:union(t:intersect(t:complement(t:NUMBER), reg.semType), toType));
```
jclark: That doesn’t look right. Static type if `<T>E` is intersection of T and static type of E.

jclark: If you are doing ConvertToInt on something with type T, then the type is int|(T - number)

jclark: Compare with the type calculation in verify.bal.

manuranga: > int|(T - number)

This is what I was trying to do, I though `T - number` and `T ∩ ∁(number)` are the same. Let me give another try.

manuranga: >Compare with the type calculation in verify.bal.

Not sure what you meant by this, but I just added bir verifications for these two instructions.

Issue #390, opened 2021-08-31
Build libbacktrace as part of runtime build


jclark:
This follows on from #367.

libbacktrace uses autoconf as its build system, but we are not using (and don't want to use) autoconf.

We also don't want libbacktrace as a separate library: we want it as part of balrt.a. Furthermore, we want to make sure we get only as much as we need of libbacktrace in any build. For example, if we have a non-debug build without symbols, there is no point in including a pile of code in our binary that can interpret debug symbols.

Another important point is that, unlike libbacktrace, we are not trying to be portable to every obsolete Unix variant ever invented: we have a very limited range of systems that we need to be portable to.

This is a prerequisite to #6 and #46.

jclark:
My suggestion would be to run configure on Ubuntu to generate config.h. Also do it on OS  X and see if there are any differences.  Then create our own config.h from this with appropriate #ifdefs. See how it builds and then write our own Makefile that does not build as a shared library. Check in both this config.h and our Makefile.

KavinduZoysa:
@manuranga, can you please clone this repo https://github.com/ianlancetaylor/libbacktrace to your OS X machine and run the command `./configure`. This command will generate a config.h file. Please send that file.

manuranga:
https://gist.github.com/manuranga/7d76846ffa699af8836f77f7f5a8b111
I have also attached a full diff of all the files that changed.

KavinduZoysa:
@jclark, I have compared two config.h files and there are 4 differences.

I will write a new Makefile as you mentioned and that Makefile will generate relevant object files. Then in our runtime Makefile, I will archive them to balrt.a

This is an exact single command libbacktrace uses to generate object file.
`gcc -DHAVE_CONFIG_H -I. -funwind-tables -frandom-seed=mmap.lo -W -Wall -Wwrite-strings -Wstrict-prototypes -Wmissing-prototypes -Wold-style-definition -Wmissing-format-attribute -Wcast-qual -g -O2 -c mmap.c -o mmap.o`

jclark:
What are the 4 differences?

Can you run configure with options so that it uses clang rather than gcc as the compiler? Otherwise, you will need to look at each of these options and determine whether it is needed and if so what the corresponding clang flag is.

"-funwind-tables" may be important and we may need that for the rest of the runtime.

KavinduZoysa:
> What are the 4 differences?
>

| macOS  | Ubuntu |
| ------------- | ------------- |
[...]

PR #391, opened 2021-08-31
Take snapshot of stack for error

### LINKED ISSUE #46: Take a snapshot of the stack that can be used to generate a backtrace
In order to implement the error value #167 and to implement #45, we need not just to print a backtrace as in issue #6, but take a snapshot of the stack, which we can store in the error value and be used subsequently to generate a backtrace.

We want to be as lazy as possible here, i.e. do as little work as possible when taking the snapshot, because the backtrace may never be needed.

KavinduZoysa:
## Purpose
Fixes #46

### DIFF
--- a/runtime/Makefile
+++ b/runtime/Makefile
@@ -25,16 +25,10 @@ all: $(LIB) $(BCLIB)
 test: all
 	$(MAKE) -C tests test
 
-$(LIBBACKTRACE_OBJS): 
-	cd third-party/libbacktrace; \
-	./configure --enable-shared
-	$(MAKE) -C third-party/libbacktrace all
-	$(AR) x third-party/libbacktrace/.libs/libbacktrace.a
-
 testCoverage: $(LIB) $(BCLIB)
 	$(MAKE) -C tests testCoverage
 
-$(LIB): $(LIBBACKTRACE_OBJS) $(OBJS) 
+$(LIB): $(OBJS)
 	$(AR) r $@ $^
 
 $(BCLIB): $(BCS)
@@ -74,10 +68,9 @@ $(LL_BCS): %.bc: %.ll
 $(OBJS) $(BCS): $(INCLUDES)
 
 clean:
-	-rm -f $(OBJS) $(BCS) $(LIB) $(BCLIB) $(LIBBACKTRACE_OBJS)
+	-rm -f $(OBJS) $(BCS) $(LIB) $(BCLIB)
 	$(MAKE) -C tests clean
 
 .PHONY: test all
 # Ensure bogus files get deleted when there's an error
 .DELETE_ON_ERROR:
-
--- a/runtime/Makefile
+++ b/runtime/Makefile
@@ -25,10 +25,16 @@ all: $(LIB) $(BCLIB)
 test: all
 	$(MAKE) -C tests test
 
+$(LIBBACKTRACE_OBJS): 
+	cd third-party/libbacktrace; \
+	./configure --enable-shared
+	$(MAKE) -C third-party/libbacktrace all
+	$(AR) x third-party/libbacktrace/.libs/libbacktrace.a
+
 testCoverage: $(LIB) $(BCLIB)
 	$(MAKE) -C tests testCoverage
 
-$(LIB): $(OBJS)
+$(LIB): $(LIBBACKTRACE_OBJS) $(OBJS) 
 	$(AR) r $@ $^
 
 $(BCLIB): $(BCS)
@@ -68,7 +74,7 @@ $(LL_BCS): %.bc: %.ll
 $(OBJS) $(BCS): $(INCLUDES)
 
 clean:
-	-rm -f $(OBJS) $(BCS) $(LIB) $(BCLIB)
+	-rm -f $(OBJS) $(BCS) $(LIB) $(BCLIB) $(LIBBACKTRACE_OBJS)
 	$(MAKE) -C tests clean
 
 .PHONY: test all
--- a/runtime/Makefile
+++ b/runtime/Makefile
@@ -74,3 +74,4 @@ clean:
 .PHONY: test all
 # Ensure bogus files get deleted when there's an error
 .DELETE_ON_ERROR:
+
--- a/runtime/balrt.h
+++ b/runtime/balrt.h
@@ -123,7 +123,8 @@ typedef GC struct Mapping {
 typedef GC struct Error {
     TaggedPtr message;
     int64_t lineNumber;
-    bool internalErrorOccured;
+    int internalErrorCode;
+    char *internalErrorMessage;
     uint32_t nPCs;
     PC pcs[];
 } *ErrorPtr;
--- a/runtime/balrt.h
+++ b/runtime/balrt.h
@@ -122,8 +122,8 @@ typedef struct {
     uintptr_t pc;
     // Index into the second array of start of string
     // string is terminated with zero
-    unsigned fileOffset;
-    unsigned functionOffset;
+    uint32_t fileOffset;
+    uint32_t functionOffset;
     int lineno;
 } Frame;
 
--- a/runtime/balrt.h
+++ b/runtime/balrt.h
@@ -118,9 +118,20 @@ typedef GC struct Mapping {
     uint8_t tableLengthShift;
 } *MappingPtr;
 
+typedef struct {
+    uintptr_t pc;
+    // Index into the second array of start of string
+    // string is terminated with zero
+    unsigned fileOffset;
+    unsigned functionOffset;
+    int lineno;
+} Frame;
+
 typedef GC struct Error {
     TaggedPtr message;
     int64_t lineNumber;
+    uint32_t nFrames;
+    Frame frames[];
 } *ErrorPtr;
 
 // Both of these are 8-byte aligned and zero-padded so the total size is a multiple of 8
--- a/runtime/balrt.h
+++ b/runtime/balrt.h
@@ -123,7 +123,7 @@ typedef GC struct Mapping {
 typedef GC struct Error {
     TaggedPtr message;
     int64_t lineNumber;
-    uint32_t npcs;
+    uint32_t nPCs;
     PC pcs[];
 } *ErrorPtr;
 
@@ -205,7 +205,7 @@ extern READONLY bool _bal_mapping_eq(TaggedPtr p1, TaggedPtr p2);
 
 extern READNONE UntypedPtr _bal_tagged_to_ptr(TaggedPtr p);
 extern TaggedPtr _bal_error_construct(TaggedPtr message, int64_t lineNumber);
-extern void _bal_error_trace_print(ErrorPtr ep);
+extern void _bal_error_backtrace_print(ErrorPtr ep);
 // Returns an error value
 extern TaggedPtr COLD _bal_panic_construct(PackedPanic err);
 
--- a/runtime/balrt.h
+++ b/runtime/balrt.h
@@ -123,6 +123,7 @@ typedef GC struct Mapping {
 typedef GC struct Error {
     TaggedPtr message;
     int64_t lineNumber;
+    bool internalErrorOccured;
     uint32_t nPCs;
     PC pcs[];
 } *ErrorPtr;
--- a/runtime/balrt.h
+++ b/runtime/balrt.h
@@ -58,6 +58,8 @@ typedef int PanicCode;
 // and line number right-shifted 8.
 typedef uint64_t PackedPanic;
 
+typedef uintptr_t PC;
+
 typedef struct {
     int64_t length;
     int64_t capacity;
@@ -118,20 +120,11 @@ typedef GC struct Mapping {
     uint8_t tableLengthShift;
 } *MappingPtr;
 
-typedef struct {
-    uintptr_t pc;
-    // Index into the second array of start of string
-    // string is terminated with zero
-    uint32_t fileOffset;
-    uint32_t functionOffset;
-    int lineno;
-} Frame;
-
 typedef GC struct Error {
     TaggedPtr message;
     int64_t lineNumber;
-    uint32_t nFrames;
-    Frame frames[];
+    uint32_t npcs;
+    PC pcs[];
 } *ErrorPtr;
 
 // Both of these are 8-byte aligned and zero-padded so the total size is a multiple of 8
@@ -212,6 +205,7 @@ extern READONLY bool _bal_mapping_eq(TaggedPtr p1, TaggedPtr p2);
 
 extern READNONE UntypedPtr _bal_tagged_to_ptr(TaggedPtr p);
 extern TaggedPtr _bal_error_construct(TaggedPtr message, int64_t lineNumber);
+extern void _bal_error_trace_print(ErrorPtr ep);
 // Returns an error value
 extern TaggedPtr COLD _bal_panic_construct(PackedPanic err);
 
--- a/runtime/error.c
+++ b/runtime/error.c
@@ -5,168 +5,97 @@
 #include "balrt.h"
 #include "third-party/libbacktrace/backtrace.h"
 
-#define INITIAL_FRAME_COUNT 1
-#define MAX_FRAME_COUNT 256
-#define INITIAL_BYTES_COUNT 8
-#define MAX_BYTES_COUNT 1 << 16
-#define BACKWARD_FRAMES_COUNT 4
+#define INITIAL_PC_COUNT 1
+#define MAX_PC_COUNT 256
 #define SKIP_FROM_END 1 // This number counts from the last stack frame
 #define THREAD 0
 
 struct backtrace_state *state;
 
 typedef struct {
-    uint32_t nFrames;
-    uint32_t szFrames;
-    uint32_t nBytes;
-    uint32_t szBytes;
-    Frame *frames;
-    char *bytes;
+    uint32_t npcs;
+    uint32_t szpcs;
+    PC* pcs;
 } Trace;
-
-typedef struct {
-    int fileOffset;
-    int functionOffset;
-} Offset;
-
-static void checkForSameStr(Frame *frame, uint32_t nFrames, const char *filename, const char *function, char *bytes, Offset* result) {
-    int i = 1;
-    bool filenameFound = false;
-    bool functionFound = false;
-    while (i <= nFrames && i <= BACKWARD_FRAMES_COUNT) {
-        Frame *f = frame - i++;
-        if (!filenameFound && strcmp(filename, bytes + f->fileOffset) == 0) {
-            result->fileOffset = f->fileOffset;
-            filenameFound = true;
-        }
-        if (!functionFound && strcmp(function, bytes + f->functionOffset) == 0) {
-            result->functionOffset = f->functionOffset;
-            functionFound = true;
-        }
-        if (filenameFound && functionFound)
-            return;
-    }
-}
 
 static void onError(void *data, const char *msg, int errnum) {
     fprintf(stderr, "Error : libbacktrace : %s\n", msg);
     fflush(stderr);
 }
 
-static int onFrame(void *data, uintptr_t pc, const char *filename, int lineno, const char *function) {
-    // This check is valid if there is a posibility not to have debug info
-    if (filename == NULL || function == NULL) {
+static int onFrame(void *data, PC pc, const char *filename, int lineno, const char *function) {
+    if (function == NULL || filename == NULL) 
         return 0;
-    }
-    Trace *trace = (Trace *)data;
-    uint32_t nFrames = trace->nFrames;
-    uint32_t szFrames = trace->szFrames;
-    if (unlikely(nFrames == MAX_FRAME_COUNT))
-        return 1;
-    if (nFrames == szFrames) {
-        szFrames = szFrames << 1;
-        trace->szFrames = szFrames;
-        void *p = realloc(trace->frames, sizeof(Frame) * szFrames);
-        if (p == 0)
-            return 1;
-        trace->frames = (Frame *)p;
-    }
 
-    uint32_t nBytes = trace->nBytes;
-    uint32_t szBytes = trace->szBytes;
-    uint64_t filenameLen = strlen(filename);
-    uint64_t funcnameLen = strlen(function);
-    uint32_t requiredBytes = nBytes + filenameLen + funcnameLen + 2;
-    if (unlikely(requiredBytes > MAX_BYTES_COUNT))
-        return 1;
-    if (requiredBytes > szBytes) {
-        szBytes = szBytes << 1;
-        while (requiredBytes > szBytes) {
-            szBytes = szBytes << 1;
-        }
-        trace->szBytes = szBytes;
-        void *p = realloc(trace->bytes, szBytes);   
-        if (p == 0)
-            return 1;
-        trace->bytes = (char *)p;
-    }
-
-    Frame *frame = trace->frames + nFrames;
-    frame->pc = pc;
-    frame->lineno = lineno;
-
-    Offset offset = {-1, -1};
-    char *bytes = trace->bytes;
-    checkForSameStr(frame, nFrames, filename, function, bytes, &offset);
-
-    if (offset.fileOffset >= 0) {
-        frame->fileOffset = offset.fileOffset;
-    }
-    else {
-        frame->fileOffset = nBytes;
-        memcpy(bytes + nBytes, filename, filenameLen);
-        nBytes = nBytes + filenameLen;
-        *(bytes + nBytes) = '\0';
-        nBytes++;
-    }
-
-    if (offset.functionOffset >= 0) {
-        frame->functionOffset = offset.functionOffset;
-    } 
-    else {
-        frame->functionOffset = nBytes;
-        memcpy(bytes + nBytes, function, funcnameLen);
-        nBytes = nBytes + funcnameLen;
-        *(bytes + nBytes) = '\0';
-        nBytes++;
-    }
-
-    trace->nBytes = nBytes;
-    trace->nFrames = nFrames + 1;
-
+    fprintf(stderr, "%s", function);
+    fprintf(stderr, " %s", filename);
+    fputs(":", stderr);
+    fprintf(stderr, "%" PRIi32, lineno);
+    putc('\n', stderr);
+    fflush(stderr);
     return 0;
 }
 
-static int getBackTrace(Trace* trace) {
+static int onPC(void *data, PC pc) {
+    Trace *trace = (Trace *)data;
+    uint32_t npcs = trace->npcs;
+    uint32_t szpcs = trace->szpcs;
+    if (unlikely(npcs == MAX_PC_COUNT))
+        return 1;
+    if (npcs == szpcs) {
+        szpcs = szpcs << 1;
+        trace->szpcs = szpcs;
+        void *p = realloc(trace->pcs, sizeof(PC) * szpcs);
+        if (p == 0)
+            return 1;    
+        trace->pcs = (PC*)p;
+    }
+
+    PC *pcs = trace->pcs;
+    *(pcs + npcs) = pc;
+
+    trace->npcs = npcs + 1;
+    return 0;
+}
+
+static int getPCs(Trace* trace) {
     state = backtrace_create_state(NULL, THREAD, onError, NULL);
 
-    void *p = malloc(sizeof(Frame) * INITIAL_FRAME_COUNT);
+    void *p = malloc(sizeof(PC) * INITIAL_PC_COUNT);
     if (p == 0)
         return 1;
-    trace->frames = (Frame *)p;
-    p = malloc(INITIAL_BYTES_COUNT);
-    if (p == 0)
-        return 1;
-    trace->bytes = (char *)p;
-
-    return backtrace_full(state, SKIP_FROM_END, onFrame, onError, trace);
+    trace->pcs = (PC *)p;
+    return backtrace_simple(state, SKIP_FROM_END, onPC, onError, trace);
 }
 
 TaggedPtr _bal_error_construct(TaggedPtr message, int64_t lineNumber) {
-    Trace trace = {0, INITIAL_FRAME_COUNT, 0, INITIAL_BYTES_COUNT};
-    int err = getBackTrace(&trace);
+    Trace trace = {0, INITIAL_PC_COUNT};
+    int err = getPCs(&trace);
     if (err) {
         fprintf(stderr, "Error : %s\n", "Error occured when generating backtrace for error value");
         fflush(stderr);
     }
-    uint32_t nFrames = trace.nFrames;
-    uint32_t nBytes = trace.nBytes;
+    uint32_t npcs = trace.npcs;
 
-    uint64_t errStructSize = sizeof(struct Error) + sizeof(Frame) * nFrames + nBytes;
+    uint64_t errStructSize = sizeof(struct Error) + sizeof(PC) * npcs;
     ErrorPtr ep = _bal_alloc(errStructSize);
     ep->message = message;
     ep->lineNumber = lineNumber;
-    ep->nFrames = nFrames;
+    ep->npcs = npcs;
 
-    Frame *frames = trace.frames;
-    char *bytes = trace.bytes;
-    memcpy(ep->frames, frames, sizeof(Frame) * nFrames);
-    memcpy(ep->frames + nFrames, bytes, nBytes);
-
-    free(frames);
-    free(bytes);
+    PC *pcs = trace.pcs;
+    memcpy(ep->pcs, pcs, sizeof(PC) * npcs);
+    free(pcs);
 
     return ptrAddFlags(ep, (uint64_t)TAG_ERROR << TAG_SHIFT);
+}
+
+void _bal_error_trace_print(ErrorPtr er) {
+    PC *pcs = (PC *)er->pcs;
+    uint32_t npcs = er->npcs;
+    for (size_t i = 0; i < npcs; i++) {
+        backtrace_pcinfo(state, *(pcs + i), onFrame, onError, NULL);
+    }
 }
 
 TaggedPtr _Berror__message(TaggedPtr error) {
--- a/runtime/error.c
+++ b/runtime/error.c
@@ -16,7 +16,6 @@
 struct backtrace_state *state;
 
 typedef struct {
-    bool err;
     uint32_t nFrames;
     uint32_t szFrames;
     uint32_t nBytes;
@@ -50,25 +49,6 @@ void checkForSameStr(Frame *frame, uint32_t nFrames, const char *filename, const
     }
 }
 
-static void *memAlloc(uint64_t nBytes) {
-    void *p = malloc(nBytes);
-    if (p != 0)
-        return p;
-    fprintf(stderr, "out of memory\n");
-    fflush(stderr);
-    abort();    
-}
-
-static void *memReAlloc(void *ptr, uint64_t nBytes) {
-    void *p = realloc(ptr, nBytes);
-    if (p != 0) {
-        return p;
-    }
-    fprintf(stderr, "out of memory\n");
-    fflush(stderr);
-    abort();
-}
-
 static void onError(void *data, const char *msg, int errnum) {
     fprintf(stderr, "Error : %s\n", msg);
     fflush(stderr);
@@ -84,7 +64,6 @@ static int onFrame(void *data, uintptr_t pc, const char *filename, int lineno, c
     uint32_t nFrames = trace->nFrames;
     uint32_t szFrames = trace->szFrames;
     if (nFrames >= MAX_FRAME_COUNT) {
-        trace->err = true;
         return 1;
     }
     else if (nFrames == szFrames) {
@@ -92,7 +71,6 @@ static int onFrame(void *data, uintptr_t pc, const char *filename, int lineno, c
         trace->szFrames = szFrames;
         void *p = realloc(trace->frames, sizeof(Frame) * szFrames);
         if (p == 0) {
-            trace->err = true;
             return 1;
         }
         trace->frames = (Frame *)p;
@@ -104,7 +82,6 @@ static int onFrame(void *data, uintptr_t pc, const char *filename, int lineno, c
     uint64_t funcnameLen = strlen(function);
     uint32_t nBytesPerFrame = filenameLen + funcnameLen + 2;
     if (nBytes + nBytesPerFrame >= MAX_BYTES_COUNT) {
-        trace->err = true;
         return 1;
     }
     else if (nBytes + nBytesPerFrame > szBytes) {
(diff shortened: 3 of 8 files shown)


runtime/error.c
```diff
@@ -1,9 +1,116 @@
...
+    framesInfo->capacityOfBytes = MIN_FRAME_COUNT * sizeof(Frame);
+    framesInfo->nBytes = 0;
+
+    backtrace_full(state, SKIP_FROM_END, on_frame, on_error, framesInfo);
+    return framesInfo;
+}
 
 TaggedPtr _bal_error_construct(TaggedPtr message, int64_t lineNumber) {
-    ErrorPtr ep = _bal_alloc(sizeof(struct Error));
+    FramesInfo *framesInfo = getFramesInfo();
```
jclark: It's better to have two separately allocated arrays while doing getFramesInfo:

```
typedef struct {
    uint32_t nFrames;
    uint32_t capacityOfFrames;
    uint32_t nBytes;
    uint32_t capacityOfBytes;
    Frame *frames;
    char *bytes;
} FramesInfo;
```

Then in _bal_error_construct:

```
   FramesInfo info;
   getFramesInfo(&info);
```

runtime/error.c
```diff
@@ -1,9 +1,116 @@
...
+
 #include "balrt.h"
+#include "third-party/libbacktrace/backtrace.h"
+
+#define MIN_FRAME_COUNT 2
+#define SKIP_FROM_END 1 // This number counts from the last stack frame
+#define THREAD 0
+
+typedef struct {
+    uint32_t nFrames;
```
jclark: "nFrames" and "capacityOfFrames" do not make a nice pair of names, because "n" is abbreviated and "capacityOf" is verbose.

runtime/error.c
```diff
@@ -1,9 +1,116 @@
...
+
+    uint64_t filenameLen = strlen(filename);
+    uint64_t funcLen = strlen(function);
+
+    if (nFrames == capacityOfFrames || (nBytes + filenameLen + funcLen + 2) >= capacityOfBytes)
+    {
+        // The current total memory size of bytes array and frames array is multiplied by two.
+        // New allocation should be devided equally between bytes array and frames array.
+        uint32_t newSize = sizeof(FramesInfo) + ((capacityOfFrames * sizeof(Frame) + capacityOfBytes) << 1);
+        framesInfo = realloc(framesInfo, newSize);
```
jclark: You must ALWAYS, ALWAYS, ALWAYS check the return value of malloc/realloc.

runtime/error.c
```diff
@@ -1,9 +1,116 @@
...
+    Frame *frames = framesInfo->frames;
+    uint32_t nFrames = framesInfo->nFrames;
+    uint32_t capacityOfFrames = framesInfo->capacityOfFrames;
+    uint32_t nBytes = framesInfo->nBytes;
+    uint32_t capacityOfBytes = framesInfo->capacityOfBytes;
+
+    uint64_t filenameLen = strlen(filename);
+    uint64_t funcLen = strlen(function);
+
+    if (nFrames == capacityOfFrames || (nBytes + filenameLen + funcLen + 2) >= capacityOfBytes)
```
jclark: How do you know these additions will not overflow?

runtime/error.c
```diff
@@ -1,9 +1,116 @@
...
+    {
+        // The current total memory size of bytes array and frames array is multiplied by two.
+        // New allocation should be devided equally between bytes array and frames array.
+        uint32_t newSize = sizeof(FramesInfo) + ((capacityOfFrames * sizeof(Frame) + capacityOfBytes) << 1);
```
jclark: How do you know this size calculation will not overflow?

runtime/error.c
```diff
@@ -1,9 +1,116 @@
...
+    uint32_t nBytes = framesInfo->nBytes;
+    uint32_t capacityOfBytes = framesInfo->capacityOfBytes;
+
+    uint64_t filenameLen = strlen(filename);
+    uint64_t funcLen = strlen(function);
+
+    if (nFrames == capacityOfFrames || (nBytes + filenameLen + funcLen + 2) >= capacityOfBytes)
+    {
+        // The current total memory size of bytes array and frames array is multiplied by two.
+        // New allocation should be devided equally between bytes array and frames array.
```
jclark: Spelling: "divided"

runtime/error.c
```diff
@@ -1,9 +1,116 @@
...
+
+    return 0;
+}
+
+FramesInfo *getFramesInfo() {
+    struct backtrace_state *state = backtrace_create_state(NULL, THREAD, on_error, NULL);
+
+    // The memory size of bytes array and frames array are equal
+    // Therefore minimum frame size is multiplied by 2
+    FramesInfo *framesInfo = malloc(sizeof(FramesInfo) + MIN_FRAME_COUNT * sizeof(Frame) * 2);
```
jclark: Completely unacceptable to not check return value of malloc.

runtime/error.c
```diff
@@ -1,9 +1,116 @@
+#include <string.h>
+#include <stdlib.h>
+#include <stdio.h>
+
 #include "balrt.h"
+#include "third-party/libbacktrace/backtrace.h"
+
+#define MIN_FRAME_COUNT 2
```
jclark: It is the initial capacity not the minimum count, isn't it.

runtime/Makefile
```diff
@@ -6,6 +6,7 @@ OPT ?= opt$(LLVM_SUFFIX)
 export LLVM_LINK ?= llvm-link$(LLVM_SUFFIX)
 VALGRIND ?= valgrind
 OBJS=alloc.o error.o list.o main.o mapping.o panic.o print.o string.o int.o third-party/dtoa/emyg_dtoa.o tagged_inline.o
+LIBBACKTRACE_OBJS=atomic.o dwarf.o fileline.o posix.o sort.o state.o backtrace.o simple.o elf.o mmapio.o mmap.o
```
jclark: These changes should be in #390.

runtime/error.c
```diff
@@ -1,9 +1,116 @@
...
+    uint32_t capacityOfBytes;
+    Frame frames[];
+} FramesInfo;
+
+void on_error(void *vdata, const char *msg, int errnum) { printf("Error : %s\n", msg); }
+
+int on_frame(void *data, uintptr_t pc, const char *filename, int lineno, const char *function) {
+    // This check is valid if there is a posibility not to have debug info
+    if (filename == NULL || function == NULL)
+    {
```
jclark: Please follow my brace style.

runtime/error.c
```diff
@@ -1,9 +1,116 @@
...
+
+typedef struct {
+    uint32_t nFrames;
+    uint32_t capacityOfFrames;
+    uint32_t nBytes;
+    uint32_t capacityOfBytes;
+    Frame frames[];
+} FramesInfo;
+
+void on_error(void *vdata, const char *msg, int errnum) { printf("Error : %s\n", msg); }
```
jclark: Why do we need this?

runtime/error.c
```diff
@@ -1,9 +1,116 @@
...
+
+int on_frame(void *data, uintptr_t pc, const char *filename, int lineno, const char *function) {
```
jclark: Why isn't this `static`?

jclark (changes requested):

jclark:
Overall, if you are going to write C code, you need to be more careful.

jclark:
When you have fixed the problems, I have identified, get @manuranga to review it also.

jclark:
Let me make some general comments about the C code in the runtime. This is part of the "trusted base" of the language. The security of all code written in Ballerina depends on this. It needs to be written with exceptional care and attention to detail. If a library function returns an error, that must **always** be checked for and handled. For every addition and every multiplication, you must think how can it be proved that this will not overflow. I will check the code when I review it, but I expect the author to have done their own thorough check first.

runtime/balrt.h
```diff
@@ -108,9 +108,20 @@ typedef GC struct Mapping {
     uint8_t tableLengthShift;
 } *MappingPtr;
 
+typedef struct {
+    uintptr_t pc;
+    // Index into the second array of start of string
+    // string is terminated with zero
+    unsigned fileOffset;
```
manuranga: can't we use uint*_t style here?

jclark: It’s going to be uint64_t on the platforms we support, but I think it’s clearer to use uintptr_t since that’s what libbacktrace uses.

runtime/error.c
```diff
@@ -1,9 +1,138 @@
...
+    void *p = realloc(ptr, nBytes);
+    if (p != 0) {
+        return p;
+    }
+    fprintf(stderr, "out of memory\n");
+    fflush(stderr);
+    abort();
+}
+
+static void on_error(void *data, const char *msg, int errnum) {
```
manuranga: way is the name in caterpillar_case when above function is in camlCase. Neither is exposed to generated code, are they?

manuranga (changes requested):
I think there might be an issue with the overall approach. I think we want to just call `backtrace_create_state` and keep the `backtrace_state` pointer around, and only call `backtrace_full` when we want to print something.

I am also concerned about following [libbacktrace comment](https://github.com/ianlancetaylor/libbacktrace/blob/4f57c999716847e45505b3df170150876b545088/backtrace.h#L81). 
> Calling this function allocates resources that cannot be freed.

Does this mean memory use irrevocably grows every time we have a new error? Don't think we can use this library at all if that's the case. Please look into this.

runtime/balrt.h
manuranga: `uintptr_t` is fine, I was commenting on the `unsigned`.

jclark: Oh, I see. Yes, uint32_t would be better here.

jclark:
> I think there might be an issue with the overall approach. I think we want to just call `backtrace_create_state` and keep the `backtrace_state` pointer around, and only call `backtrace_full` when we want to print something.

No, I don't tink so. We need backtrace_full to snapshot the stack. backtrace_state is not the state of one backtrace, it is the state needed for libbacktrace to be able to print the backtrace of anything (i.e. symbol tables).

> I am also concerned about following [libbacktrace comment](https://github.com/ianlancetaylor/libbacktrace/blob/4f57c999716847e45505b3df170150876b545088/backtrace.h#L81).
> 
> > Calling this function allocates resources that cannot be freed.
> 
> Does this mean memory use irrevocably grows every time we have a new error? Don't think we can use this library at all if that's the case. Please look into this.

You call backtrace_create_state once (or once per thread), before you call any of the other functions, not once for each time you want a backtrace.

manuranga:
> No, I don't tink so. We need backtrace_full to snapshot the stack. backtrace_state is not the state of one backtrace, it is the state needed for libbacktrace to be able to print the backtrace of anything (i.e. symbol tables).

Hmm. I was hoping there will be a function that will just give us some pointers to the text section of the program, which we can save and later turn to strings. looks like there is no way to do this, at least not in libbacktrace.

> You call backtrace_create_state once (or once per thread), before you call any of the other functions, not once for each time you want a backtrace.

Ah, got it. @KavinduZoysa lets move `backtrace_state` to a global variable.

runtime/error.c
```diff
@@ -1,9 +1,138 @@
...
+
+static void on_error(void *data, const char *msg, int errnum) { 
+    fprintf(stderr, "Error : %s\n", msg);
+    fflush(stderr);
+    abort();
+}
+
+static int on_frame(void *data, uintptr_t pc, const char *filename, int lineno, const char *function) {
+    // This check is valid if there is a posibility not to have debug info
+    if (filename == NULL || function == NULL) {
```
manuranga: Even when we don't have filenames, it's still useful to have stacktraces with just function names. Please check if that case works.

runtime/error.c
```diff
@@ -1,9 +1,138 @@
...
+#define THREAD 0
+
+typedef struct {
+    uint32_t nFrames;
+    uint32_t cFrames;
+    uint32_t nBytes;
+    uint32_t cBytes;
+    Frame *frames;
+    char *bytes;
+} FramesInfo;
```
manuranga: Isn't this basically the `Trace` struct?
(Minor point: personally don't like the word `Info`.)

jclark: (I'm not keen on `Info` either, although I have been guilty of using it.)

`cBytes` is not comprehensible: `size` (which abbreviates to `sz`) is another word you can use instead of capacity.

runtime/error.c
```diff
@@ -1,9 +1,138 @@
...
+#define SKIP_FROM_END 1 // This number counts from the last stack frame
+#define THREAD 0
+
+typedef struct {
+    uint32_t nFrames;
+    uint32_t cFrames;
+    uint32_t nBytes;
+    uint32_t cBytes;
+    Frame *frames;
+    char *bytes;
```
manuranga: I would have written this to have a single monolithic block of memory.
```C
typedef struct {
    uint32_t nFrames;
    size_t capacity;
    void *bytes;
} Trace;
```

```
+--+------+---------------+----------------+--+------+----------...
|pc|lineNo|func name .../0| file name .../0|pc|lineNo|.............
+--+------+---------------+----------------+--+------+----------...
```
But I am not sure if this is better than what you have done here. You have the advantage of being able access things as struct fields (eg `frame->lineno`) which is nice, but having two block, one referring to other is not looking nice to me.

@jclark any preference ?

jclark: I think what he has done is fine (I suggested it). The blocks are saved together in one allocation in the Error object. Doing it like this makes for significantly simpler code: no need to manually pad out the strings to an appropriate alignment boundary.

jclark: Other big advantage of this approach is that filenames/function names can be shared between frames.  Often many frames in a backtrace will have the same function name and/or filename.

jclark: The code should scan a few frames to try to see if it can reuse a filename/function name.

runtime/error.c

runtime/error.c
KavinduZoysa: I have added this check to catch the library functions which run before C main function.

runtime/error.c
KavinduZoysa: Variable names of the function signature are the same as what is defined in libbacktrace. I have changed the function names to camalCase.

runtime/error.c
KavinduZoysa: Changed.

runtime/error.c
KavinduZoysa: This is the implementation of libbacktrace error callback function. I thought that it will be useful to catch libbacktrace related errors.

runtime/error.c

runtime/error.c
KavinduZoysa: Yes, Changed

runtime/error.c
KavinduZoysa: Fixed in all places.

runtime/error.c
KavinduZoysa: Changed.

runtime/error.c
```diff
@@ -1,9 +1,171 @@
...
+    void *p = malloc(sizeof(Frame) * INITIAL_FRAME_COUNT);
+    if (p == 0)
+        return 1;
+    trace->frames = (Frame *)p;
+    p = malloc(INITIAL_BYTES_COUNT);
+    if (p == 0)
+        return 1;
+    trace->bytes = (char *)p;
+
+    return backtrace_full(state, SKIP_FROM_END, onFrame, onError, trace);
```
jclark: In the last call, we concluded that we should call backtrace_simple when  constructing an error so as to store just the pc. Then we would use backtrace_pcinfo or backtrace_syminfo (according to whether we have debugging info or no), which is the expensive part, when we need to print it out.

jclark (changes requested):

runtime/error.c
KavinduZoysa: @jclark , in the `_bal_error_constructor` we only have to construct the error with pcs and we have to create a new method like `_bal_error_trace_print` in error.c. Is that correct?

jclark: Right.

KavinduZoysa: `<function>(<filepath>:<lineno>)`
@jclark, Is above format looks okay to when printing backtrace?

jclark: You shouldn't need to allocate while printing: just print it out from the callback.

jclark: Looks ok. Have a look at what some other languages do.

Note that you should not need to allocate when printing out: just print directly from the backtrace.  Eventually we will need to demangle the function name, but demangling will need to change to support multiple modules, so let's leave this till later.

manuranga: Something simple is enough for now.

In the long term we'll need to design this properly. On the jBallerina side we have done some changes recently to include cause in the trace, and also to remove duplicate entities when we are doing so (similar to Java). There were some requests from debugger fork asking if we can show entries from parent strand, which we haven't done, but maybe that is a useful thing? Feel like we need a place to document this, is spec the correct place?

@warunalakshitha @gabilang

KavinduZoysa: > Looks ok. Have a look at what some other languages do.
> 
Error trace for java and rust is shown below.
Java:
```
<ERROR_TYPE>
        at <function>(<filename>:<lineno>)
```

Rust:
```
<#number> <function>
       at <filepath>:<lineno>:<columnno>

```
Go has a similar format as shown above. In python, function name appears at the end of the trace line, but lineno always follow the filepath.

jclark: I agree lineno should follow filepath.

I don't care for putting the filename/lineno in parentheses. Implement whatever you prefer and we can adjust later.

KavinduZoysa:
@jclark `Pstackoverflow00` is failing because it reaches the max frames count when constructing error. I do not think increase max frame count is not a good idea. So do we need to introduce a fix to remove backtrace generation when it is not needed?

jclark:
Store just the first MAX_PC_COUNT frames.  The error construction should *never* fail and *never* print anything to stderr.

runtime/balrt.h
```diff
@@ -121,6 +123,8 @@ typedef GC struct Mapping {
 typedef GC struct Error {
     TaggedPtr message;
     int64_t lineNumber;
+    uint32_t npcs;
```
jclark: Not consistent with other names in this file

runtime/balrt.h
```diff
@@ -201,6 +205,7 @@ extern READONLY bool _bal_mapping_eq(TaggedPtr p1, TaggedPtr p2);
 
 extern READNONE UntypedPtr _bal_tagged_to_ptr(TaggedPtr p);
 extern TaggedPtr _bal_error_construct(TaggedPtr message, int64_t lineNumber);
+extern void _bal_error_trace_print(ErrorPtr ep);
```
jclark: _bal_error_backtrace_print

runtime/error.c
```diff
@@ -1,12 +1,103 @@
+#include <string.h>
+#include <stdlib.h>
+#include <stdio.h>
+
 #include "balrt.h"
+#include "third-party/libbacktrace/backtrace.h"
+
+#define INITIAL_PC_COUNT 1
+#define MAX_PC_COUNT 256
+#define SKIP_FROM_END 1 // This number counts from the last stack frame
```
jclark: This should be a parameter at the time you print out

runtime/error.c
```diff
@@ -1,12 +1,103 @@
...
+#define THREAD 0
+
+struct backtrace_state *state;
+
+typedef struct {
+    uint32_t npcs;
+    uint32_t szpcs;
+    PC *pcs;
+} Trace;
```
jclark: This is just a vector of PCs.  So call it PCVector or something like that.

runtime/error.c
```diff
@@ -1,12 +1,103 @@
...
+
+static void onError(void *data, const char *msg, int errnum) {
+    fprintf(stdout, "Error : libbacktrace : %s\n", msg);
```
jclark: Need a name that says what it is doing, not just when it is called. Is this being called when printing the backtrace? If so, then name it accordingly.

runtime/error.c
```diff
@@ -1,12 +1,103 @@
...
+    fflush(stdout);
+}
+
+static int onFrame(void *data, PC pc, const char *filename, int lineno, const char *function) {
+    if (function == NULL || filename == NULL)
+        return 0;
+
+    fprintf(stdout, "%s", function);
```
jclark: stderr not stdout

runtime/error.c
```diff
@@ -1,12 +1,103 @@
...
+    putc('\n', stdout);
+    fflush(stdout);
+    return 0;
+}
+
+static int onPC(void *data, PC pc) {
+    Trace *trace = (Trace *)data;
+    uint32_t npcs = trace->npcs;
+    uint32_t szpcs = trace->szpcs;
+    if (unlikely(npcs == MAX_PC_COUNT))
```
jclark: Enforce the limit on szpcs not npcs.

runtime/error.c
```diff
@@ -1,12 +1,103 @@
...
+        return 1;
+    if (npcs == szpcs) {
+        szpcs = szpcs << 1;
+        trace->szpcs = szpcs;
+        void *p = realloc(trace->pcs, sizeof(PC) * szpcs);
+        if (p == 0)
+            return 1;
```
jclark: Define macros for success/fail instead of using 0 and 1.

runtime/error.c
```diff
@@ -1,12 +1,103 @@
...
+        trace->pcs = (PC *)p;
+    }
+
+    PC *pcs = trace->pcs;
+    *(pcs + npcs) = pc;
```
jclark:
```
trace->pcs[npcs] = pc;
```

runtime/error.c
```diff
@@ -1,12 +1,103 @@
...
+    if (p == 0)
+        return 1;
+    trace->pcs = (PC *)p;
+    return backtrace_simple(state, SKIP_FROM_END, onPC, onError, trace);
+}
 
 TaggedPtr _bal_error_construct(TaggedPtr message, int64_t lineNumber) {
-    ErrorPtr ep = _bal_alloc(sizeof(struct Error));
+    Trace trace = {0, INITIAL_PC_COUNT};
+    int err = getPCs(&trace);
```
jclark: Do malloc here.

runtime/error.c
```diff
@@ -1,12 +1,103 @@
...
+    if (err) {
+        fprintf(stdout, "%s\n", "Error occured when getting backtrace for error value");
```
jclark: No. Recover from errors. Do not print stuff to stderr.

runtime/error.c
```diff
@@ -1,12 +1,103 @@
...
+    memcpy(ep->pcs, pcs, sizeof(PC) * npcs);
+    free(pcs);
+
     return ptrAddFlags(ep, (uint64_t)TAG_ERROR << TAG_SHIFT);
 }
 
+void _bal_error_trace_print(ErrorPtr er) {
+    PC *pcs = (PC *)er->pcs;
+    uint32_t npcs = er->npcs;
+    for (size_t i = 0; i < npcs; i++) {
```
jclark: Why size_t?

runtime/error.c
```diff
@@ -1,12 +1,103 @@
...
+    ep->npcs = npcs;
+
+    PC *pcs = trace.pcs;
+    memcpy(ep->pcs, pcs, sizeof(PC) * npcs);
+    free(pcs);
+
     return ptrAddFlags(ep, (uint64_t)TAG_ERROR << TAG_SHIFT);
 }
 
+void _bal_error_trace_print(ErrorPtr er) {
```
jclark: err not er

runtime/error.c
```diff
@@ -1,12 +1,103 @@
...
+    PC *pcs = (PC *)er->pcs;
```
jclark: No cast needed

runtime/error.c
```diff
@@ -1,12 +1,103 @@
...
+    void *p = malloc(sizeof(PC) * INITIAL_PC_COUNT);
+    if (p == 0)
+        return 1;
+    trace->pcs = (PC *)p;
+    return backtrace_simple(state, SKIP_FROM_END, onPC, onError, trace);
+}
 
 TaggedPtr _bal_error_construct(TaggedPtr message, int64_t lineNumber) {
-    ErrorPtr ep = _bal_alloc(sizeof(struct Error));
+    Trace trace = {0, INITIAL_PC_COUNT};
```
jclark: malloc here

runtime/error.c
```diff
@@ -1,12 +1,103 @@
...
+    return 0;
+}
+
+static int getPCs(Trace *trace) {
+    state = backtrace_create_state(NULL, THREAD, onError, NULL);
+
+    void *p = malloc(sizeof(PC) * INITIAL_PC_COUNT);
+    if (p == 0)
+        return 1;
+    trace->pcs = (PC *)p;
```
jclark: No cast needed

runtime/error.c
```diff
@@ -1,12 +1,103 @@
...
+
+    PC *pcs = trace->pcs;
+    *(pcs + npcs) = pc;
+
+    trace->npcs = npcs + 1;
+    return 0;
+}
+
+static int getPCs(Trace *trace) {
+    state = backtrace_create_state(NULL, THREAD, onError, NULL);
```
jclark: Call backtrace_state once, not every call to getPCs

Can backtrace_create_state return null? Either check or comment

runtime/error.c
```diff
@@ -1,12 +1,103 @@
...
+
 #include "balrt.h"
+#include "third-party/libbacktrace/backtrace.h"
+
+#define INITIAL_PC_COUNT 1
+#define MAX_PC_COUNT 256
+#define SKIP_FROM_END 1 // This number counts from the last stack frame
+#define THREAD 0
+
+struct backtrace_state *state;
```
jclark: Initialize to 0 explicitly

runtime/error.c
```diff
@@ -1,12 +1,103 @@
...
+    PC *pcs;
+} Trace;
+
+static void onError(void *data, const char *msg, int errnum) {
+    fprintf(stdout, "Error : libbacktrace : %s\n", msg);
+    fflush(stdout);
+}
+
+static int onFrame(void *data, PC pc, const char *filename, int lineno, const char *function) {
+    if (function == NULL || filename == NULL)
```
jclark: Be consistent about NULL or 0 for null pointers

jclark (changes requested):
This needs a lot of work.

Also add declarations of static functions at the beginning, and organize functions in a top-down way.

runtime/error.c
```diff
@@ -1,12 +1,103 @@
...
+    Trace *trace = (Trace *)data;
+    uint32_t npcs = trace->npcs;
+    uint32_t szpcs = trace->szpcs;
+    if (unlikely(npcs == MAX_PC_COUNT))
+        return 1;
+    if (npcs == szpcs) {
+        szpcs = szpcs << 1;
+        trace->szpcs = szpcs;
+        void *p = realloc(trace->pcs, sizeof(PC) * szpcs);
+        if (p == 0)
```
jclark: Brace style in this project is to use braces even when there is a single statement.

runtime/error.c
KavinduZoysa: Using NULL

runtime/error.c

runtime/error.c
KavinduZoysa: Fixed.
This can be null if an error occurred and will call error_callback function.

runtime/error.c
KavinduZoysa: Without this cast, I am getting a cast error because ErrorPtr is defined in addrspace(1).

runtime/error.c
KavinduZoysa: Changed to ep

runtime/error.c

runtime/error.c
KavinduZoysa: Is it okay to print a message to stdout if an error occurred?

runtime/error.c
KavinduZoysa: Is it need to malloc Trace struct here, since we memcpy the PCs inside this method?

runtime/error.c

runtime/error.c

runtime/error.c

runtime/error.c

runtime/error.c
KavinduZoysa: This is the error_callback function which is common for backtrace_simple, backtrace_create_state, and backtrace_pcinfo. That is why I followed the same pattern used for other callback functions.

runtime/error.c

runtime/error.c
KavinduZoysa: This is a parameter that we need to pass `backtrace_simple` and without this, libbacktrace generates an entry for `backtrace_simple` function too. So shall we keep it like this?

runtime/balrt.h
KavinduZoysa: Fixed

runtime/balrt.h

runtime/error.c
```diff
@@ -1,12 +1,117 @@
...
+}
+
+static int onFrame(void *data, PC pc, const char *filename, int lineno, const char *function) {
+    if (function == NULL || filename == NULL) {
+        return SUCCESS;
+    }
+
+    fprintf(stderr, "%s", function);
+    fprintf(stderr, " %s", filename);
+    fputs(":", stderr);
```
manuranga: why `fputs` in some cases and `fprintf` in others ?

runtime/error.c
```diff
@@ -1,12 +1,117 @@
...
+
+void _bal_error_backtrace_print(ErrorPtr ep) {
+    PC *pcs = (PC *)ep->pcs;
+    uint32_t nPCs = ep->nPCs;
+    for (uint32_t i = 0; i < nPCs; i++) {
+        backtrace_pcinfo(state, pcs[i], onFrame, onError, NULL);
+    }
+}
+
+static int onFrame(void *data, PC pc, const char *filename, int lineno, const char *function) {
```
manuranga: Wasn't there a suggestion previously to rename this to something like `printBacktraceLine`.
Maybe we need the word `callback` in there as well, or at least a comment saying this is an impl of `backtrace_full_callback`

KavinduZoysa: Changed callback function names and add a comment to indicate this is a callback

runtime/error.c
KavinduZoysa: For print formatted strings I have used `fprintf` and `fputs` is used for print the supplied string.

runtime/error.c
KavinduZoysa: Changed the callback function names.

runtime/error.c
manuranga: ah I just noticed the space, in any case wouldn't a single fprintf work better?

KavinduZoysa: Yes, single `fprintf` looks better. Changed it.

runtime/error.c
manuranga: I don't think so, one option is to keep a flag in `struct Error` and print it when you are printing the error.
I think we'd have a better idea of what to do, if we knew what kind of cases would give us error here.

KavinduZoysa: Yes, we know cases where an error occurs.
1. Memory allocation/reallocation
2. Create backtrace state
3. Exceed MAX_PC_COUNT limit

manuranga: For 1, 3 we at least have a palatially correct stacktrace, maybe even in 2 (when does this happens?). I think we should keep an `internalErrorCode` in `struct Error` and set when creating.
When printing we can consider this, and do something appropriate.
eg:
```
    myfunc1(file1.bal:2)
    myfunc3(file1.bal:6)
    ... (some lines omitted due to exceeding maximum stack trace limit)
```

(maybe this pr is not the correct place to do the full implementation, but avoid printing during construction)

KavinduZoysa: I will change this. The 2nd case happens if an error occurs in `backtrace_create_state`

jclark: Printing a backtrace should be on a best efforts basis. Print out as much of a backtrace as you can (up to MAX_PC_COUNT lines). You can print `...` if you are omitting some.

KavinduZoysa: Okay.
Also, we can do the same thing(set a flag if an error occurs) for error_callback functions without printing the error inside the error_callback function.

KavinduZoysa: Changed.

runtime/error.c
```diff
@@ -1,12 +1,128 @@
...
+static void getPCs(PCVector *pcVector) {
+    if (state == NULL) {
+        state = backtrace_create_state(NULL, THREAD, NULL, NULL);
+        if (state == NULL) {
+            pcVector->errorOccurred = true;
+            return;
+        }
+    }
+
+    void *p = malloc(sizeof(PC) * INITIAL_PC_COUNT);
```
jclark: It's better to do this malloc in _bal_error_construct so that the pcVector is always in a consistent state with the allocated size of pcs equal to szpcs.

runtime/error.c
```diff
@@ -1,12 +1,128 @@
...
+            return FAIL;
+        }
+        szPCs = szPCs << 1;
+        pcVector->szPCs = szPCs;
+        void *p = realloc(pcVector->pcs, sizeof(PC) * szPCs);
+        if (p == NULL) {
+            pcVector->errorOccurred = true;
+            return FAIL;
+        }
+        pcVector->pcs = (PC *)p;
```
jclark: If you have a cast here, you are doing something wrong.

runtime/error.c
```diff
@@ -1,12 +1,128 @@
...
+        pcVector->errorOccurred = true;
+        return;
+    }
+    pcVector->pcs = (PC *)p;
+    backtrace_simple(state, SKIP_FROM_END, updatePCVector, setBackTraceSimpleInternalError, pcVector);
+}
+
+// Implementation of backtrace_simple_callback
+static int updatePCVector(void *data, PC pc) {
+    PCVector *pcVector = (PCVector *)data;
```
jclark: Why do you need a cast here?

runtime/error.c
```diff
@@ -1,12 +1,128 @@
...
+    }
+
+    pcVector->pcs[nPCs] = pc;
+    pcVector->nPCs = nPCs + 1;
+    return SUCCESS;
+}
+
+// Implementation of backtrace_error_callback
+static void setBackTraceSimpleInternalError(void *data, const char *msg, int errnum) {
+    PCVector *pcVector = (PCVector *)data;
```
jclark: Why do you need a cast here?

runtime/error.c
```diff
@@ -1,12 +1,128 @@
...
+#define THREAD 0
+#define SUCCESS 0
+#define FAIL 1
+
+struct backtrace_state *state = NULL;
+
+typedef struct {
+    uint32_t nPCs;
+    uint32_t szPCs;
+    bool errorOccurred;
```
jclark: errorOccurred should separately keep track of whether

- MAX_PC_COUNT was exceeded
- whether an error was returned from libbacktrace; if so, the msg and the errnum (check if msg needs to be copied)

Printing the backtrace should print something different for each case:
- `...` for the former
- the error message from libbacktrace in the latter case

jclark (changes requested):

jclark:
This is much improved.  But you need to pay more attention to type casts.  Each time you have a type cast, please put a comment explaining why a type cast is needed. Mostly, the type casts are not needed and are just covering up the underlying problem (such as not declaring with the right address space).

runtime/error.c
```diff
@@ -1,12 +1,128 @@
...
+    PC *pcs;
+} PCVector;
```
jclark: Now that you have added errorOccurred (which is a good addition), PCVector is no longer a good name. PCVector is now saving the result of  a call to backtrace_simple. So I think I woud rename it to `SimpleBacktrace`.

jclark (changes requested):

runtime/error.c
```diff
@@ -1,12 +1,128 @@
...
+        pcVector->pcs = (PC *)p;
+    }
+
+    pcVector->pcs[nPCs] = pc;
+    pcVector->nPCs = nPCs + 1;
+    return SUCCESS;
+}
+
+// Implementation of backtrace_error_callback
+static void setBackTraceSimpleInternalError(void *data, const char *msg, int errnum) {
```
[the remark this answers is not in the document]

runtime/error.c

runtime/error.c
KavinduZoysa: removed

runtime/error.c
KavinduZoysa: It is removed.

runtime/error.c

runtime/error.c
```diff
@@ -1,12 +1,128 @@
...
+    PCVector *pcVector = (PCVector *)data;
+    pcVector->errorOccurred = true;
+}
+
+void _bal_error_backtrace_print(ErrorPtr ep) {
+    PC *pcs = (PC *)ep->pcs;
```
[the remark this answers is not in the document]

runtime/error.c
jclark: In addition to the two kinds of error described above, there's also the possibility of an allocation failure from malloc or realloc.

KavinduZoysa: @jclark, I have two unclear points here.
1. When looking at libbacktrace source code, mostly they are using 0 and -1 as errnum. But for each errnum, the error message is not unique. So I think it is not useful to keep libbacktrace errnum in our struct. We can keep only the message. (We have to copy the message because buffer is invalid after function returns)
2. Do we need to keep a list of errors in our struct? Since in backtrace_simple or backtrace_pcinfo functions, there is a possibility to call error_callback several times.

If we are not going to print libbacktrace specific error, we have only three types of error
1. MAC_PC_COUNT exceeds
2. allocation failure
3. libbacktrace internal error

runtime/error.c

runtime/error.c
jclark: The header file has this to say about errnum:

>  The ERRNUM argument, if
   greater than 0, holds an errno value.  The MSG buffer may become
   invalid after this function returns.

 >  As a special case, the ERRNUM argument will be passed as -1 if no
   debug info can be found for the executable, or if the debug info
   exists but has an unsupported version, but the function requires
   debug info (e.g., backtrace_full, backtrace_pcinfo).  The MSG in
   this case will be something along the lines of "no debug info".
   Similarly, ERRNUM will be passed as -1 if there is no symbol table,
   but the function requires a symbol table (e.g., backtrace_syminfo).
   This may be used as a signal that some other approach should be
   tried.

So, errnum is useful: it gives you the errno when libbacktrace failed because of a failed system/library call. But we need msg also (which, as you say, should be copied).

I think a list of errors is not necessary.

runtime/error.c
```diff
@@ -1,12 +1,184 @@
...
+
+// Implementation of backtrace_error_callback
+static void setBacktraceInternalError(void *data, const char *msg, int errnum) {
+    SimpleBacktrace *simpleBacktrace = data;
+    // the error that is created first is tracked
+    if (simpleBacktrace->errorCode != NO_ERROR) {
+        return;
+    }
+    simpleBacktrace->errorCode = errnum;
+    uint64_t messageLength = strlen(msg);
```
jclark, line 112: There's a very bad error here (and in the other callback). You are not copying the terminating `\0`. This could cause invalid memory access when the message is accessed.

Furthermore this allocated memory is not being freed.

PR #392, opened 2021-09-01
Build libbacktrace with runtime

### LINKED ISSUE #390: Build libbacktrace as part of runtime build

KavinduZoysa:
## Purpose
Fixes #390

This PR add seperate Makefile for libbacktrace and it will create necessary object files. Then runtime Makefile is changed to archive this object files into balrt.a

### DIFF
--- a/runtime/Makefile
+++ b/runtime/Makefile
@@ -5,8 +5,9 @@ LLC ?= llc$(LLVM_SUFFIX)
 OPT ?= opt$(LLVM_SUFFIX)
 export LLVM_LINK ?= llvm-link$(LLVM_SUFFIX)
 VALGRIND ?= valgrind
-OBJS=alloc.o error.o list.o main.o mapping.o panic.o print.o string.o int.o third-party/dtoa/emyg_dtoa.o tagged_inline.o
+CWARNINGS=-W -Wall -Wwrite-strings -Wstrict-prototypes -Wmissing-prototypes -Wold-style-definition -Wmissing-format-attribute -Wcast-qual
 LIBBACKTRACE_OBJS=atomic.o dwarf.o fileline.o posix.o sort.o state.o backtrace.o simple.o elf.o mmapio.o mmap.o
+OBJS=alloc.o error.o list.o main.o mapping.o panic.o print.o string.o int.o third-party/dtoa/emyg_dtoa.o tagged_inline.o $(addprefix third-party/libbacktrace/, $(LIBBACKTRACE_OBJS))
 C_BCS=eq_inline.bc int_inline.bc list_inline.bc mapping_inline.bc string_inline.bc float_inline.bc
 LL_BCS=float_ir_inline.bc tagged_inline.bc
 BCS=$(C_BCS) $(LL_BCS)
@@ -26,18 +27,17 @@ test: all
 testCoverage: $(LIB) $(BCLIB)
 	$(MAKE) -C tests testCoverage
 
-$(LIB): $(LIBBACKTRACE_OBJS) $(OBJS)
+$(LIB): $(OBJS)
 	$(AR) r $@ $^
-
-$(LIBBACKTRACE_OBJS):
-	$(MAKE) -C third-party/libbacktrace all
-	$(AR) x third-party/libbacktrace/libbacktrace.a
 
 $(BCLIB): $(BCS)
 	set -o pipefail; \
 	$(LLVM_LINK) -o - -S $^ | \
 	sed -e '/define .*@_[Bb][a-zA-Z]/s/^define /define linkonce_odr /' -e '/target datalayout/ s/"$$/-ni:1"/' | \
 	$(LLVM_AS) >$@ 
+
+$(filter third-party/libbacktrace/%.o, $(OBJS)): third-party/libbacktrace/%.o: third-party/libbacktrace/%.c
+	$(CLANG) -DHAVE_CONFIG_H -funwind-tables -frandom-seed=$< $(CWARNINGS) -g -O2 -c -o $@ $<
 
 %.o: %.bc
 	$(CLANG) -O2 -c -o $@ $<
@@ -67,9 +67,8 @@ $(LL_BCS): %.bc: %.ll
 $(OBJS) $(BCS): $(INCLUDES)
 
 clean:
-	-rm -f $(OBJS) $(BCS) $(LIB) $(BCLIB) $(LIBBACKTRACE_OBJS)
+	-rm -f $(OBJS) $(BCS) $(LIB) $(BCLIB)
 	$(MAKE) -C tests clean
-	$(MAKE) -C third-party/libbacktrace clean
 
 .PHONY: test all
 # Ensure bogus files get deleted when there's an error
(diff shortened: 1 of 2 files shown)


KavinduZoysa:
@manuranga, can you check out this PR and try to build the runtime?

runtime/Makefile
```diff
@@ -25,9 +26,13 @@ test: all
 testCoverage: $(LIB) $(BCLIB)
 	$(MAKE) -C tests testCoverage
 
-$(LIB): $(OBJS)
+$(LIB): $(LIBBACKTRACE_OBJS) $(OBJS)
 	$(AR) r $@ $^
 
+$(LIBBACKTRACE_OBJS):
```
jclark: I don't like this at all.

Why do we need a separate Makefile for libbacktrace? Use `$(addsuffix)` to generate the list of objects in the third-party/libbacktrace directory, and then use a static pattern rule to specify the compilation rule for those objects. The objects can then be added to balrt.a directly just as emyg_dtoa.o is.

jclark (changes requested):

runtime/third-party/libbacktrace/backtrace-supported.h
```diff
@@ -0,0 +1,66 @@
...
+/* BACKTRACE_USES_MALLOC will be #define'd as 1 if the backtrace
+   library will call malloc as it works, 0 if it will call mmap
+   instead.  This may be used to determine whether it is safe to call
+   the backtrace functions from a signal handler.  In general this
+   only applies to calls like backtrace and backtrace_pcinfo.  It does
+   not apply to backtrace_simple, which never calls malloc.  It does
+   not apply to backtrace_print, which always calls fprintf and
+   therefore malloc.  */
+
+#define BACKTRACE_USES_MALLOC 0
```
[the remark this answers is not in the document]
KavinduZoysa: I think that it does not affect our callback. All functions defined in alloc.c and mmap.c, are internal functions called by the library.

runtime/third-party/libbacktrace/config.h
```diff
@@ -0,0 +1,180 @@
...
+#endif
+
+/* Define if AIX loadquery is available. */
+/* #undef HAVE_LOADQUERY */
+
+/* Define to 1 if you have the `lstat' function. */
+#define HAVE_LSTAT 1
+
+/* Define to 1 if you have the <mach-o/dyld.h> header file. */
+#ifdef __APPLE__
```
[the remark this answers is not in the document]
KavinduZoysa: I have added this check because we know the difference between Ubuntu and macOS. 
I am currently trying to figure out the exact test by looking at configure file.

runtime/Makefile
KavinduZoysa: Changed to build libbacktrace without seperate MakeFile.

runtime/Makefile
```diff
@@ -34,6 +36,9 @@ $(BCLIB): $(BCS)
 	sed -e '/define .*@_[Bb][a-zA-Z]/s/^define /define linkonce_odr /' -e '/target datalayout/ s/"$$/-ni:1"/' | \
 	$(LLVM_AS) >$@ 
 
+$(filter third-party/libbacktrace/%.o, $(OBJS)): third-party/libbacktrace/%.o: third-party/libbacktrace/%.c
```
jclark: Make a definition for `$(addprefix third-party/libbacktrace/, $(LIBBACKTRACE_OBJS))` and use it here and in the definition of OBJS.

You don't need `third-part/libbacktrace/` in this rule.

runtime/Makefile
```diff
@@ -5,7 +5,9 @@ LLC ?= llc$(LLVM_SUFFIX)
 OPT ?= opt$(LLVM_SUFFIX)
 export LLVM_LINK ?= llvm-link$(LLVM_SUFFIX)
 VALGRIND ?= valgrind
-OBJS=alloc.o error.o list.o main.o mapping.o panic.o print.o string.o int.o third-party/dtoa/emyg_dtoa.o tagged_inline.o
+CWARNINGS=-W -Wall -Wwrite-strings -Wstrict-prototypes -Wmissing-prototypes -Wold-style-definition -Wmissing-format-attribute -Wcast-qual
```
jclark: These warnings are specific to libbacktrace, so use a name like LIBBACKTRACE_WARN_FLAGS

runtime/Makefile
```diff
@@ -5,7 +5,9 @@ LLC ?= llc$(LLVM_SUFFIX)
...
+LIBBACKTRACE_OBJS=atomic.o dwarf.o fileline.o posix.o sort.o state.o backtrace.o simple.o elf.o mmapio.o mmap.o
```
jclark: I would abbreviate LIBBACKTRACE in make variable names to `BT` or

runtime/third-party/libbacktrace/backtrace-supported.h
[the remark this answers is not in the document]
jclark: Why are we configuring libbacktrace not to use malloc?

runtime/third-party/libbacktrace/config.h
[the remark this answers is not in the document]
jclark: You need to read the documentation for Apple/Darwin and see what they document as being the right predefined macro to test.  We want to test for OS X/Darwin not for any random OS made by Apple.

jclark (changes requested):

runtime/Makefile
KavinduZoysa: Changed.

runtime/Makefile
KavinduZoysa: Fixed.

KavinduZoysa: Fixed.

runtime/Makefile
KavinduZoysa: Fixed.

runtime/third-party/libbacktrace/backtrace-supported.h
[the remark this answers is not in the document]
KavinduZoysa: This file is added by configure and it is common for both Ubuntu and macOS. Let me read about this more. What I currently notice is that it will not generate an object for alloc.c only generate the object for mmap.c.

runtime/third-party/libbacktrace/config.h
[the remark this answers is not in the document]
KavinduZoysa: Changed this based on the comment https://github.com/ballerina-platform/nballerina/issues/390#issuecomment-911514917

manuranga:
@KavinduZoysa I have committed auto generated files via the bot, please rebase.
I have selected all files for now. If needed, please remove any files (even the non-generated ones) by changing [mapping.txt](https://github.com/ballerina-platform/nballerina/blob/third-party/libbacktrace/mapping.txt) file.

jclark (approved):
This looks good to me.

Issue #358, opened 2021-08-24
Update test suite to use same format as conformance test suite


jclark:
Development has started an a conformance test suite for the language in the spec repo, which is using a test format similar to us, but with multiple tests in a single `.balt` file:

https://github.com/ballerina-platform/ballerina-spec/blob/ac5724f6bfd36ee76d473b1ccf7064c4e158d664/conformance/lang/values-types-and-variables/array.balt

We should update our test suite to use this.

I would suggest that for each subset, we have just a few balt files (one for each feature added) e.g. subset06-float.balt

jclark:
We should also make this work for semtype tests, so we can share these between @rdhananjaya 's port and here cc @MaryamZi

manuranga:
Q: In the test run output, should it list each `balt` file or each case?
A: It should be each case, so that we can select to run a single case using `--tests` flag.

Q: In that case what should be the name? 
A: I think an incrementing number should work.

Eg:
```
	nballerina
		[pass] testBalt#'float.balt-00'
		[pass] testBalt#'float.balt-01'
...
		[pass] testBalt#'string.balt-00'
```

jclark:
Instead of an incrementing number, I suggest Lnnn where nnn is the line number of the Test-Case line in the .balt file.

manuranga:
I changed it slightly,
```
                [pass] testBalt#'map_types.balt:1'
                [pass] testBalt#'map_types.balt:17'
                [pass] testBalt#'map_types.balt:25'
                [pass] testBalt#'map_types.balt:33'
                [pass] testBalt#'map_types.balt:49'
```
This way when you ctrl+click the name, VSCode opens up the file and jump to correct location.

manuranga:
This is where I am at right now,
https://github.com/manuranga/nballerina/compare/8334939cacdc0621edee35a54414ae553cfb0916..f074cb21d434f90828c9e9b611959cef7deaa8a3

What I am working on
1. I am in  the process of breaking up `subset01.balt` into `loops.balt`, `int.balt`, `boolean.balt`, `conditional.balt`, `function.bal`.
2. Find a way to run balt. Current public api in `main.bal` accepts file path.
   opt 1. write to separate bal files
   opt 2. create an api that accepts lines and compile
3. How to call from Makefile in batch mode
   opt 1. write to separate bal files (reuse the files from above)
   opt 2. cli option to run balt

jclark:
[...]

Issue #365, opened 2021-08-26
Support multiple source files in a module


jclark:
This goes hand in hand with #364.

jclark:
There's a bit of added complexity: it also requires representing imports per file rather than per module.

PR #394, opened 2021-09-02
Implement balt test runner

manuranga:
Add balt tests for subset-01

Part of #358

### DIFF
--- a/compiler/balt.bal
+++ b/compiler/balt.bal
@@ -1,0 +1,169 @@
+import wso2/nballerina.err;
+
+import ballerina/io;
+import ballerina/file;
+
+type TestKind "output" | "panic" | "error";
+
+
+type BaltTestHeader record {|
+    TestKind 'Test\-Case;
+    string...;
+|};
+
+type BaltTestCase record {|
+    int offset;
+    BaltTestHeader header;
+    string[] content;
+|};
+
+enum State {
+    BOF,
+    HEADER,
+    CONTENT
+}
+
+// " " is a good approximation
+// http://www.bitdance.com/blog/2011/04/11_01_Email6_Rewriting_Header_Folding/
+const CONTINUATION_WS = " ";
+
+function parseBalt(string path) returns  BaltTestCase[]|io:Error|file:Error|err:Any {
+    BaltTestCase[] tests = [];
+    string[] lines = check io:fileReadLines(path);
+
+    BaltTestHeader? maybeHeader = ();
+    string? prevFiledBody  = ();
+    string? prevFiledName  = ();
+    string[] content = [];
+    int offset = 0;
+
+    State s = BOF;
+    foreach var [i, l] in lines.enumerate() {
+        if l.startsWith("Test-Case:") {
+            if s != BOF {
+                BaltTestHeader header = <BaltTestHeader>maybeHeader;
+                tests.push({ offset, header, content });
+            }
+            s = HEADER;
+
+            offset = i + 1;
+            content = [];
+            var [fName, fBody] = parseField(l);
+            maybeHeader = {Test\-Case: <TestKind>fBody};
+            prevFiledBody = fBody;
+            prevFiledName = fName;
+        }
+        else if s == HEADER {
+            if l == "" {
+                s = CONTENT;
+            }
+            else {
+                BaltTestHeader header = <BaltTestHeader>maybeHeader;
+                // possible improvement: support other white spaces
+                if l.startsWith(" ") {
+                    if prevFiledName == () ||  prevFiledBody == () {
+                        panic err:impossible("folded field body without field");
+                    }
+                    header[<string>prevFiledName] = <string>prevFiledBody + l;
+                } else {
+                    var [fName, fBody] = parseField(l);
+                    header[fName] = fBody;
+                    prevFiledName = fName;
+                    prevFiledBody = fBody;
+                }
+            }
+        }
+        else if s == CONTENT {
+            content.push(l);
+        }
+        else if s == BOF {
+            // xxx add file path
+            return err:syntax("file should start with 'Test-Case:' header field");
+        }
+        else {
+            panic err:impossible("balt parser illegal state");
+        }
+    }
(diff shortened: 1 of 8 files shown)


jclark:
I suggest we split this up into:

1. changes to the driver to support .balt input files
2. converting tests from current format into .balt

We can merge (1) before we have done (2).

manuranga:
I have change the ll file to have both an incremental number and line number. Makefile's incremental/parallel testing is mostly working, I need few adjustments.

```
out
├── compile.stamp
├── 01-boolean.stamp
├── ...
├── expect
│   ├── 01-boolean
│   │   ├── V01-boolean-0000-0001.txt
│   │   ...
│   │   └── V01-boolean-0010-0179.txt
├── ll
│   ├── 01-boolean
│   │   ├── V01-boolean-0000-0001.txt
```

Should I organize `out` dir to be  kind-major (`out/expect/01-boolean/V01-boolean-0000-0001.txt`) or to be balt-name-major (`out/01-boolean/expect/V01-boolean-0000-0001.txt`) ?

jclark:
Is there any need to repeat eg 01-boolean in the filename as well as the directory name?

jclark:
Should be out/01-boolean

manuranga:
> Is there any need to repeat eg 01-boolean in the filename as well as the directory name?

No need, it's there due to a historical reason, will remove.

>Should be out/01-boolean

Ack

jclark:
Another thing to keep in mind: we will soon want to support multiple files per module; when we do so, the meaning of multiple `.bal` files on the command-line should be to treat them as all being part of the same module.  It would be good if we can test this with `.balt` files.

Also we are soon going to want to add support for multiple modules, which we will mean we need test cases with multiple modules. We will need to be able to test using `.balt` files.

No need to do either of these things right now, but we should design so we can easily evolve to support them.

jclark:
I am going to attempt multiple source files #365 now.

manuranga:
Getting `out/01-boolean/expect/V0000L0001.txt` patten needed some fiddling around because you can't write
`out/%/expect/%.txt` as a rule. Adding `.SECONDEXPANSION` did the trick.

I will test things a bit more and send it soon.

jclark:
We shouldn't need a rule `out/%/expect/%.txt`. We should be doing a recursive make in `out/01-boolean`.

manuranga:
I updated the PR, since it's working and is incremental/parallel, and ready to be reviewed.
I still need to switch to recursive make approach for testing sub dirs.

jclark:
I reworked `main.bal` a lot as part of #365. It had become a complete mess.

compiler/balt.bal
```diff
@@ -0,0 +1,162 @@
...
+        hash = fnv1a32(l.toBytes(), hash);
+    }
+    return hash;
+}
+
+function fnv1a32(byte[] bs, int init = FNV1_32_INIT) returns int {
+    int hash = init;
+    foreach byte b in bs {
+        hash ^= b;
+        hash *= FNV1_32_PRIME;
```
jclark: Isn't this going to overflow?

compiler/balt.bal
```diff
@@ -0,0 +1,162 @@
...
+            expect.push("panic: line " + (i + 1).toString() + ": " + msg);
+        }
+    }
+    return expect;
+}
+
+const FNV1_32_INIT = 0x811c9dc5;
+const FNV1_32_PRIME = 0x1000193;
+
+function hash(string[] lines) returns int {
```
jclark: I thought we were not hashing any more.

compiler/balt.bal
manuranga: No, because hash is at most `0xffffffff`

compiler/balt.bal
manuranga: Ya, will remove.

manuranga:
> I reworked main.bal a lot as part of #365. It had become a complete mess.

Ya, I also noticed and did some refactoring, but unfortunately now it's fully conflicted with current main. Easier to get yours and rework form there.

manuranga:
I temporary move all test bal files to files named by it's subset so we can get this pr merged.
Please use GitHub `Select commit` feature to review.

We may need a way to distinguish `FV` from `FE`, for now I am skipping everything with a `Fail-Issue:`

jclark:
The Makefile can be done more straightforwardly using recursion as follows.

Leave the current Makefile as is for now and start off with two new Makefiles. Let's call them new.mk and sub.mk: new.mk is the one that will replace the current Makefile when we are finished; sub.mk is called by new.mk. For each balt file `NN-foo.balt` there will be an output directory `out/NN-foo/`. The job of new.mk is to run sub.mk on each of these output directories in parallel. The current directory for sub.mk will be the per-balt output directory (this is why it is less confusing to keep it separate).  Here's how you can do the recursion:

new.mk looks like this:

```
TARGETS=test all compile testll

# Construct this list with the GNU make wildcard function
TDIRS=01-boolean 01-conditional 01-function

$(TARGETS): $(TDIRS)

$(TDIRS):
	mkdir -p out/$@
	$(MAKE) -C out/$@ -f ../../sub.mk tdir=$@ $(MAKECMDGOALS)

.PHONY: $(TARGETS) $(TDIRS)
```

I used this sub.mk to test:

```
TARGETS=test all compile testll

$(TARGETS):
	@echo Starting make $@ in `pwd` with tdir=$(tdir)
	@sleep 3
	@echo Ending make $@ in `pwd` with tdir=$(tdir)

.PHONY: $(TARGETS)
```

You can derive sub.mk straightforwardly starting from current Makefile as follows:

1. remove things that are unrelated to a recursive invocation
2. changes paths to work with a current directory being the balt-specific output directory
   - some paths need ../.. added
   - some paths need to change from out/F to F
3. run the compile on a single balt file (whose name you can derive from tdir)

If you don't want to do this, I am happy to do it, but I need the changes to main.bal to support compiling balt files first.

manuranga:
Had to add a sample test due to ballerina-platform/ballerina-lang#32575.

PR #399, opened 2021-09-02
Run GitHub Actions on macOS

manuranga:
Run GitHub Actions on macOS

### DIFF
--- a/test/findJava.sh
+++ b/test/findJava.sh
@@ -1,6 +1,4 @@
-#!/bin/sh
 bal_command=`which bal`
-bal_command=`realpath "$bal_command"`
 bal_dir=`dirname "$bal_command"`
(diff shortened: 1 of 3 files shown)


test/findJava.sh
```diff
@@ -1,6 +1,4 @@
-#!/bin/sh
 bal_command=`which bal`
-bal_command=`realpath "$bal_command"`
```
manuranga, line 3: I removed `realpath` since it's not available on macOS.
It works without it on my machine, do we need to find an alternative ?

jclark: You can just run the make with `JAVA=/foo/bar/bin/java make`, then it shouldn't use findJava.sh.

jclark (changes requested):

manuranga:
Closing this for now, since I am going to split the workflow to test jni separately, will reopen after testing that.

PR #404, opened 2021-09-03
Compile elf.c and macho.c according to OS

### LINKED ISSUE #403: MacOS build is failing for libbacktrace
**Description:**
This build failure occurs because it tries to compile elf.c which is not needed for MacOS. So we have to stop compiling it and we have to compile macho.c

KavinduZoysa:
## Purpose
Fixes #403


jclark:
I thought of a slightly cleaner way to do this https://github.com/ballerina-platform/nballerina/issues/403#issuecomment-912447073

Issue #419, opened 2021-09-06
Organize test suite into directories


manuranga:
Organize test suite into directories.

Alternative to #358

manuranga:
Current proposal is to use a format like:
`testSuite/01-int/subtract-01-v.bal`

jclark:
Just: `testSuite/01-int/subtract1-v.bal`

The number in the filename is not required.

manuranga:
Proposed renamings https://gist.github.com/manuranga/97f3bdcaad2e31cf6c6f9d0bc4a79fdb

This is purely based on the original names, not on labels.

jclark:
There are a couple like this:

> Vnileq03             -> 03-nil/equal4-v.bal

I think you should start at 1 in each created directory rather than adding 1.

So 06-const/19-e.bal  would be 1-e.bal.

In 03. I would distinguish langlib and lib.

I would merge 01-const into 01-int.

I would merge 07-panic into 07-error.

Does it make sense to categorize F cases by subset? Maybe just put all F cases into a fail directory.

There's lots more that could be tweaked, but I think it's easier just to do it and then tweak later.

PR #421, opened 2021-09-07
List renamings of the test suite

manuranga:
List renameings  of the test suite
This a fake PR, **not to be merged**, so you can comment on each line.
Part of #419


compiler/testSuite/renamings.txt
```diff
@@ -0,0 +1,497 @@
...
+Verror01             -> 07-error/4-v.bal
+Verror02             -> 07-error/5-v.bal
+Verror03             -> 07-error/6-v.bal
+Verror04             -> 07-error/7-v.bal
+Verror05             -> 07-error/8-v.bal
+Verror06             -> 07-error/9-v.bal
+Ppanic00             -> 07-error/panic1-p.bal
+Ppanic01             -> 07-error/panic2-p.bal
+Ppanic02             -> 07-error/panic3-p.bal
+Plist05              -> 07-list/1-p.bal
```
jclark: 07-array is better

array goes with map; list goes with mapping

jclark (commented):
There's endless tweaking that could be done. Let's just get this done.

compiler/testSuite/renamings.txt
manuranga: Ok, let's do the individual improvements later, but if you notice any renameings that involve multiple files, it might be easier to modify the script now, please let me know if you notice such cases.

I will work on modifying `test/Makefile`.
I'll rename `new.mk`` and `sub.mk` to `balt.mk` and `balt-sub.mk`.
Then I can introduce a new `sub.mk` and invoke it form `test/Makefile`.

PR #424, opened 2021-09-07
Organize test suite into directories

### LINKED ISSUE #419: Organize test suite into directories

manuranga:
Organize test suite into directories
Fixes #419


manuranga:
Not ready to merge yet, still changing the `testSuite.bal`
Line history of the `Makefile` is persevered in `sub.mk`, even though it appears green here.

test/sub.mk
```diff
@@ -22,26 +32,29 @@ all:
...
-	-../../update.sh expectnew expect txt
-	-../../update.sh llnew ll ll
+	cd llnew; for f in *.ll; do cmp -s $$f ../ll/$$f || mv $$f ../ll/; done
 	-rm -fr llnew
-	-rm -fr expectnew
 	@touch $@
+endif
 
+# This compiles, runs and checks the output of ll/*.ll
 testll: fail.txt
```
jclark: This isn't showing the changes relative to the current Makefile.  In the current Makefile, we have:

```
testll: out/result/fail.txt $(expect_files) $(exe_files)
```

Did you intend to remove the dependency on expect_files and exe_files? If so, why?

manuranga: You may view the correct diff by selecting just the relevant commit ([Modify makefile content into sub.mk](https://github.com/ballerina-platform/nballerina/pull/424/commits/273d57d46e9d473fdd841e2cee82aa5c6ca85d11)).

Yes it was intentional, I though since fail.txt depends on diff_files and each diff is depended on exe and the expect text, explicitly adding it here is redundant.

test/Makefile
```diff
@@ -7,72 +7,73 @@
...
+	../../expect.sh $< >$@
 
 clean:
-	-rm -fr out
-
-update:
-	cd out/ll; for f in *.ll; do cmp -s $$f ../../ll/$$f || cp -p $$f ../../ll/; done
+	-rm -rf actual compile.stamp expect fail.txt ll result
 
-.PHONY: all test clean compile testll update
```
manuranga, line 78: I have accidentally removed `clean` form `.PHONY`, will add it back

test/sub.mk
jclark: I see your reasoning, but I am not confident that's a good change, because there are a lot of complexities about how pattern rules work when dependencies do not exist.

In any case, this is an unrelated change and so should be not be mixed in with this.

manuranga: I'll add it back.

manuranga: [Added](https://github.com/ballerina-platform/nballerina/pull/424/commits/6c7e43661dbf49b8e07b8e5349d51a6992f2e64d).

PR #431, opened 2021-09-10
Add `make diff` and improve `make update`

### LINKED ISSUE #430: Diff `test/ll` against produced ll via Makefile
Remove `test/diff.sh` and do it via Makefile.
Output should just be a list of files instead of full diffs.

manuranga:
An example output of `make diff`:

```
> ll/01-ifelse/2-v.ll
< out/01-int/ll/add3-v.ll
out/cmpll/01-function/1-v.diff
```

`make update` is improved to remove files that are no longer in `out/*/ll`

Fixes #430

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -13,9 +13,20 @@ clean:
 	-rm -fr out
 
 update:
+	mkdir -p out
+	find out -path "*/ll/*.ll" | sed -e 's;/ll/;/;' -e 's;^out;ll;' | sort > out/ll_curr.txt
+	find ll -type f | sort > out/ll_prev.txt
+	comm -13 out/ll_curr.txt out/ll_prev.txt | xargs rm -fv
+	rm out/ll_curr.txt out/ll_prev.txt
 	for f in out/*/ll/*.ll; do \
 		ll=`echo $$f | sed -e 's;/ll/;/;' -e 's;^out;ll;'`; \
 		cmp -s $$f $$ll || (mkdir -p `dirname $$ll` && cp -p $$f $$ll); \
 	done
 
-.PHONY: $(TARGETS) $(TDIRS) clean update
+diff: out/cmpll/ll_curr.txt out/cmpll/ll_prev.txt
+	@for f in out/*/ll/*.ll; do \
+		ll=`echo $$f | sed -e 's;/ll/;/;' -e 's;^out;ll;'`; \
+		[ -e $$ll ] && cmp -s $$f $$ll || echo $$f; \
+	done
+
+.PHONY: $(TARGETS) $(TDIRS) clean update diff
--- a/test/diff.sh
+++ b/test/diff.sh
@@ -1,6 +1,0 @@
-#!/bin/sh
-
-for f in out/*/ll/*.ll
-do
-    diff -u $f $(echo $f | sed -e 's;/ll/;/;' -e 's;^out;ll;')
-done


test/Makefile
```diff
@@ -12,10 +12,33 @@ $(TDIRS):
...
+	comm -13 out/cmpll/ll_curr.txt out/cmpll/ll_prev.txt | xargs rm -fv
+	rm out/cmpll/ll_curr.txt out/cmpll/ll_prev.txt
 	for f in out/*/ll/*.ll; do \
 		ll=`echo $$f | sed -e 's;/ll/;/;' -e 's;^out;ll;'`; \
 		cmp -s $$f $$ll || (mkdir -p `dirname $$ll` && cp -p $$f $$ll); \
 	done
 
-.PHONY: $(TARGETS) $(TDIRS) clean update
+diff: out/cmpll/ll_curr.txt out/cmpll/ll_prev.txt
+	@diff out/cmpll/ll_curr.txt out/cmpll/ll_prev.txt | sed -e 's;^< ll/\(.*\)/;< out/\1/ll/;' -e '/^[<>]/!d'
```
jclark: Can you explain what this is doing?  It's a lot more complicated that I expected.

manuranga: I wanted to produce an output like :
```
> ll/01-ifelse/2-v.ll
< out/01-int/ll/add3-v.ll
out/cmpll/01-function/1-v.diff
```
Here `ll/01-ifelse/2-v.ll` is to be removed, `out/01-int/ll/add3-v.ll` is a new file to be added.

I produce this by listing the two sets of files, massaging it to have the same format and getting the diff between them. Same has been used in `update:` to remove files. Once I get the diff I undo the massaging to get the original file names (so you may click on it to view).

test/Makefile
```diff
@@ -12,10 +12,33 @@ $(TDIRS):
...
+	@rm out/cmpll/ll_curr.txt out/cmpll/ll_prev.txt
+	@for curr in out/*/ll/*.ll; do \
+		( \
+			d=`echo $$curr | sed -e 's;\.ll$$;.diff;' -e 's;/ll/;/;' -e 's;^out;out/cmpll;'`; \
+			mkdir -p `dirname $$d`; \
+			prev=`echo $$curr | sed -e 's;/ll/;/;' -e 's;^out;ll;'`; \
+			[ -e $$prev ] && diff -u $$curr $$prev> $$d; \
+			[ -s $$d ] && echo $$d || rm -f $$d \
+	    )& \
```
jclark: What is the `&` for? Doesn't that run it in the background? I can't see how that works.

manuranga: Yes. It was was taking couple of seconds to run, this speeds it up.

manuranga: Because now it runs in parallel.

jclark: That's not the right way to do things.  You can't have background jobs in a Makefile!

jclark: This is just far too complicated, is doing unnecessary work and so it is slow. All it needs to do is give a list of the files that are changed (files that were present before and after, but are different).

jclark: It does not need to diff them. It does not to list files that were added or removed.

manuranga: In the last meeting I mentioned that I am going to write diffs so one can click and view it, I thought you were OK with it. Let me change to just cmp.

jclark:
I just merged and tried `make diff` and I get:

make: *** No rule to make target 'out/cmpll/ll_curr.txt', needed by 'diff'.  Stop.

PR #432, opened 2021-09-10
Print backtrace on panic

### LINKED ISSUE #6: Enhance runtime panic function to print backtrace before exiting
Eventually panics need to unwind until they are caught with `trap`, but until then our implementation of panic can simply print a backtrace before exiting.

This needs some debug info (#7), and is an enhancement to the strategy in #44, which explicitly passes the location of the offending instruction to the runtime panic function.

KavinduZoysa:
## Purpose
Fixes #6


runtime/error.c
```diff
@@ -150,11 +175,29 @@ void _bal_error_backtrace_print(ErrorPtr ep) {
...
+static void printMissingSymbols(void *data, uintptr_t pc, const char *symname, uintptr_t symval, uintptr_t symsize) {
+    printLine(NULL, 0, symname);
+}
+
+static void printLine(const char *filename, int64_t lineNumber, const char *function) {
     if (function != NULL && filename != NULL) {
-        fprintf(stderr, "%s %s:%" PRIi32 "\n", function, filename, lineno);
-        fflush(stderr);
+        fprintf(stderr, "\t%s %s:%" PRIi64 "\n", function, filename, lineNumber);
+    } else if (function != NULL) {
```
jclark: Brace style.

runtime/error.c
```diff
@@ -116,13 +128,26 @@ static int storePC(void *data, PC pc) {
...
+    if (start < nPCs) {
+        BacktraceStartLine backtraceStartLine = { err, ep->lineNumber };
+        backtrace_pcinfo(state, pcs[start], printBacktraceStartLine, storeBacktraceError, &backtraceStartLine);
+        err.code = backtraceStartLine.error.code;
+        err.message = backtraceStartLine.error.message;
+        if (err.code == NO_DEBUG_INFO) {
+            backtrace_syminfo(state, pcs[start], printMissingSymbols, storeBacktraceError, &err);
+        }
+        for (uint32_t i = start + 1; i < nPCs; i++) {
+            backtrace_pcinfo(state, pcs[i], printBacktraceLine, storeBacktraceError, &err);
```
jclark: If you got NO_DEBUG_INFO error above, do you want to keep calling pcinfo?

runtime/error.c
```diff
@@ -116,13 +128,26 @@ static int storePC(void *data, PC pc) {
...
 
-void _bal_error_backtrace_print(ErrorPtr ep) {
+void _bal_error_backtrace_print(ErrorPtr ep, uint32_t start) {
     GC PC *pcs = ep->pcs;
     uint32_t nPCs = ep->nPCs;
     BacktraceError err = { ep->backtraceErrorCode, ep->backtraceErrorMessage };
-    for (uint32_t i = 0; i < nPCs; i++) {
-        backtrace_pcinfo(state, pcs[i], printBacktraceLine, storeBacktraceError, &err);
+    if (start < nPCs) {
+        BacktraceStartLine backtraceStartLine = { err, ep->lineNumber };
```
jclark: I would separate everything in this if branch into a separate function.

jclark (changes requested):

runtime/error.c
```diff
@@ -150,11 +175,29 @@ void _bal_error_backtrace_print(ErrorPtr ep) {
 
 // Implementation of backtrace_full_callback
 static int printBacktraceLine(void *data, PC pc, const char *filename, int lineno, const char *function) {
+    printLine(filename, lineno, function);
+    return SUCCESS;
+}
+
+// Implementation of backtrace_full_callback
+static int printBacktraceStartLine(void *data, PC pc, const char *filename, int lineno, const char *function) {
+    BacktraceStartLine* backtraceStart = data;
```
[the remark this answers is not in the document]

runtime/error.c

runtime/error.c
KavinduZoysa: It is not needed, changed.

runtime/error.c

runtime/error.c
jclark: Are you sure? What happens if in a frame that contains functions from multiple modules, some modules have debug info and some don't?

jclark:
I am going to work on #434, which should make most of this unnecessary.

KavinduZoysa:
> I am going to work on #434, which should make most of this unnecessary.

Ok, I will change this PR accordingly.

runtime/error.c
KavinduZoysa:
```
   As a special case, the ERRNUM argument will be passed as -1 if no
   debug info can be found for the executable, or if the debug info
   exists but has an unsupported version, but the function requires
   debug info (e.g., backtrace_full, backtrace_pcinfo).  The MSG in
   this case will be something along the lines of "no debug info".
   Similarly, ERRNUM will be passed as -1 if there is no symbol table,
   but the function requires a symbol table (e.g., backtrace_syminfo).
   This may be used as a signal that some other approach should be
   tried. 
```


Based on this libbacktrace comment, what I understood was, errnum -1 comes if the whole executable does not have debug info or symbol table. In the same source file, if some functions do not have debug info, libbacktrace does not call error_callback. Let me check what happens when we have multiple modules.

runtime/error.c
```diff
@@ -148,13 +165,66 @@ void _bal_error_backtrace_print(ErrorPtr ep) {
...
+}
+
+// Implementation of backtrace_syminfo_callback
+static void printInitialMissingSymbols(void *data, uintptr_t pc, const char *symname, uintptr_t symval, uintptr_t symsize) {
+    BacktraceStartLine *backtraceStartLine = data;
+    printLine(NULL, backtraceStartLine->lineNumber, symname);
+}
+
+static void printLine(const char *filename, int64_t lineNumber, const char *function) {
     if (function != NULL && filename != NULL) {
```
jclark: You should check separately for function/filename/lineno being NULL/0, like I did here:

https://github.com/ballerina-platform/nballerina/blob/5f4da8c8ebd0d26f39984ee00faca7c7d13561a5/runtime/error.c#L151

jclark (changes requested):

jclark:
#434 isn't going to work: https://github.com/ballerina-platform/nballerina/issues/434#issuecomment-917346365

We should therefore continue to pass in a line number (and maybe in future a column number) as an argument to _bal_panic_construct and _bal_error_construct. If  the line number passed in as an argument is zero, then that means no information is being provided about the line number; this can happen within a langlib function.

When printing out a backtrace on a panic, the first line should be `panic` plus the message (with no line number), e.g.

```
panic: arithmetic overflow
```

There should be one subsequent line (indented) for each frame in the backtrace that has location information. The line  number should be omitted if it is zero. `runcheck.sh` and `expect.sh` need to be fixed to deal with this format. (They can just check the first line number for now.)

The backtrace printed out should be derived from the *combination* of both

- the line number passed in as an argument
- debugging information from libbacktrace

Specifically the line number for the first frame should be the argument line number if non-zero, and otherwise the line number from libbacktrace.

runtime/error.c
```diff
@@ -148,13 +165,66 @@ void _bal_error_backtrace_print(ErrorPtr ep) {
     fflush(stderr);
 }
 
+static void processPCs(GC PC* pcs, uint32_t nPCs, uint32_t start, int64_t lineNumber, BacktraceError *error) {
+    // Handle the starting pc seperately
+    processInitialPC(pcs[start], lineNumber, error);
+    if (error->code == NO_DEBUG_INFO) {
+        for (uint32_t i = start + 1; i < nPCs; i++) {
```
jclark, line 171: Combine these two loops by having a function pointer that is backtrace_pcinfo or backtrace_syminfo.

runtime/error.c
```diff
@@ -148,13 +165,66 @@ void _bal_error_backtrace_print(ErrorPtr ep) {
...
+        }
+    }
+    else {
+        for (uint32_t i = start + 1; i < nPCs; i++) {
+            backtrace_pcinfo(state, pcs[i], printBacktraceLine, storeBacktraceError, error);
+        }
+    }
+}
+
+static void processInitialPC(PC pc, int64_t lineNumber, BacktraceError *error) {
```
jclark: Don't have two separate callbacks for first line vs other lines. Instead always pass in a line number, with zero meaning no information is being provided.

jclark (changes requested):

runtime/error.c
KavinduZoysa: @jclark, Is it fine to use a function pointer since the function signatures of backtrace_pcinfo and backtrace_syminfo are different(we have to pass two different callbacks)?
Maybe I have misunderstood your suggestion.

runtime/error.c
KavinduZoysa:
```
// Implementation of backtrace_full_callback
static int printBacktraceLine(void *data, PC pc, const char *filename, int lineno, const char *function) {
    if (lineno <= 0) {
        BacktraceStartLine *backtraceStartLine = data;
        lineno = backtraceStartLine->lineNumber;
    }
    printLine(filename, lineno, function);
    return SUCCESS;
}
```


For the `backtrace_pcinfo` case that is true. We can do something like the above for all lines. 

But `backtrace_syminfo` case is a bit different. For the first line, if we get errnum -1, we can get the function name. Since we have the first lineNumber we can generate the first trace line with the function name and lineNumber.  For rest of the lines, lineNumber should be zero. If we pass lineNumber to the callback of `backtrace_syminfo`, we do not have information in that callback to compare the lineNumber as shown above. Therefore `backtrace_syminfo` case we will need two seperate callbacks for first lines vs other lines.

runtime/error.c
jclark: I had not noticed they were different. My suggestion won't work.

runtime/error.c
jclark: Can you pass in a zero line number after the first line?

KavinduZoysa: Please ignore my comment. This can be done.

runtime/error.c
KavinduZoysa: I used this implementation and did a small change. I changed `%d` to `%ld`, because the type of lineNumber passed from _bal_panic_construct is int64_t

jclark: `%ld` is right if you are using `long`. int64_t needs PRi64

KavinduZoysa:
MacOS `-p` tests are failing with below log,

```
panic: arithmetic overflow
    _B_add 9
    (no info about stack frame)
    printLine
libbacktrace: no debug info in Mach-O executable
```

Also, if I replace the `clang -O2` with `llc -O2 && clang -O0`, it starts to work. I am working to fix this issue

jclark:
> MacOS `-p` tests are failing with below log,
> 
> ```
> panic: arithmetic overflow
>     _B_add 9
>     (no info about stack frame)
>     printLine
> libbacktrace: no debug info in Mach-O executable
> ```
> 
> Also, if I replace the `clang -O2` with `llc -O2 && clang -O0`, it starts to work. I am working to fix this issue

That needs to be fixed, but why does it cause the p tests to fail? The test suite tests should be just checking the line number, which you've got.

And why are you printing out a `no debug info` error? That's errnum -1, right? Having no debug info isn't an error.

KavinduZoysa:
> That needs to be fixed, but why does it cause the p tests to fail? The test suite tests should be just checking the line number, which you've got.
> 

In runcheck.sh I am trying to generate something shown below. But in macOS filename is NULL. Therefore diff fails.

```
panic: arithmetic overflow
add2-b.bal:9
```

> And why are you printing out a `no debug info` error? That's errnum -1, right? Having no debug info isn't an error.

Here I have changed the runcheck.sh just to print everything to figure out where the actual error is. That's why `libbacktrace:`  is shown. Yes this is errnum -1.

jclark:
> But in macOS filename is NULL

I see: you're checking the filename too. That's good.

KavinduZoysa:
@jclark, using `dsymutil` does not work. But adding `-g` flag to the final [clang command](https://github.com/KavinduZoysa/nballerina-1/commit/fbc462c59720adda1b64a336d541f405693b9915#diff-89a5dd5857c744afeb3204f39a51524aaa7ec4460293f2a090d3b9202870cf57) causes to generate the correct backtrace. So is that find to add the -g flag after checking the OS in sub.mk?

manuranga:
> using dsymutil does not work. 

Can you please elaborate on how it doesn't work, does it not produce `.dSYM`? does it give an error?

KavinduZoysa:
> > using dsymutil does not work.
> 
> Can you please elaborate on how it doesn't work, does it not produce `.dSYM`? does it give an error?

It produces .dSYM with a warning `warning: no debug symbols in executable (--arch x86_64)`. Then I ran the executable but the backtrace is not correct. Then I change the sub.mk to run with -g and ran the tests

jclark:
> So is that find to add the -g flag after checking the OS in sub.mk?

Can you see whether adding -g to that particular command increases the size of the executable on Linux significantly? If not, then we can always add it.

runtime/error.c
```diff
@@ -148,15 +161,71 @@ void _bal_error_backtrace_print(ErrorPtr ep) {
...
+static void printMissingSymbols(void *data, uintptr_t pc, const char *symname, uintptr_t symval, uintptr_t symsize) {
+    BacktraceStartLine *backtraceStartLine = data;
+    printLine(NULL, backtraceStartLine->lineNumber, symname);
+}
+
+static void printLine(const char *filename, int64_t lineNumber, const char *function) {
+    // API docs say any of filename, lineno and function may be 0 meaning unavailable
+    const char *sep = "    ";
+    if (function != NULL) {
+        fprintf(stderr, "%s%s", sep, function);
```
manuranga: I think it's better to use compile time string concatenation. 
```C
define ERR_SEPR "     "
...

fprintf(stderr, "%s" ERR_SEPR, function);
```

KavinduZoysa:
> > So is that find to add the -g flag after checking the OS in sub.mk?
> 
> Can you see whether adding -g to that particular command increases the size of the executable on Linux significantly? If not, then we can always add it.

I checked, there is not a size change.

runtime/error.c
manuranga: Oh I see you are assigning to `sep` below, hmm still looks wired to me, why concat static strings at runtime?

jclark: I wrote this: it keeps the cases more uniform. There's no performance issue here.

runtime/error.c

runtime/error.c

runtime/error.c
```diff
@@ -148,15 +161,71 @@ void _bal_error_backtrace_print(ErrorPtr ep) {
...
+static void processInitialPC(PC pc, int64_t lineNumber, BacktraceStartLine *backtraceStartLine) {
+    backtrace_pcinfo(state, pc, printBacktraceLine, storeBacktraceError, backtraceStartLine);
+    if (backtraceStartLine->error.code == NO_DEBUG_INFO) {
+        // We do not need to pass error_callback because, already we have an error
+        backtrace_syminfo(state, pc, printMissingSymbols, NULL, backtraceStartLine);
+    }
+}
+
 // Implementation of backtrace_full_callback
 static int printBacktraceLine(void *data, PC pc, const char *filename, int lineno, const char *function) {
```
jclark: printBacktraceLine and printLine are not well named: printLine prints a line of a backtrace.

Maybe add `Callback` or `CB` at the end of the names of every function that is used as a callback for libbacktrace?

Then printBacktraceLine would become printBacktraceLineCB and printLine would become printBacktraceLine.

jclark (changes requested):

runtime/error.c

runtime/tests/error_backtrace1.c
```diff
@@ -25,7 +25,7 @@ void func2() {
 void func1() {
     TaggedPtr taggedErr = _bal_error_construct(makeString("Func1 error"), 26);
     ErrorPtr ep = (ErrorPtr)taggedToPtr(taggedErr);
-    _bal_error_backtrace_print(ep);
+    _bal_error_backtrace_print(ep, 1);
```
jclark: I don't see how these tests work. How are you checking that the output is correct?

runtime/error.c
```diff
@@ -116,13 +129,13 @@ static int storePC(void *data, PC pc) {
     return SUCCESS;
 }
 
-void _bal_error_backtrace_print(ErrorPtr ep) {
-    GC PC *pcs = ep->pcs;
+void _bal_error_backtrace_print(ErrorPtr ep, uint32_t start) {
```
jclark: I would add a FILE * as a parameter here, so you can test it more easily.

runtime/tests/error_backtrace1.c
KavinduZoysa: Fixed

runtime/error.c

runtime/tests/error_backtrace1.c
```diff
@@ -3,32 +3,61 @@
...
 void func3() {
-    TaggedPtr taggedErr = _bal_error_construct(makeString("Func3 error"), 7);
+    TaggedPtr taggedErr = _bal_error_construct(makeString("Func3 error"), 27);
     ErrorPtr ep = (ErrorPtr)taggedToPtr(taggedErr);
-    _bal_error_backtrace_print(ep);
-    PC *pcs = (PC *)ep->pcs;
-    uint32_t nPCs = ep->nPCs;
-    assert(nPCs >= 4);
+
+    FILE *fp = fopen("/tmp/backtrace.txt", "w+");
```
jclark: That's not the right way to create a temp file.

jclark (changes requested):

runtime/tests/error_backtrace1.c
KavinduZoysa: Fixed

jclark:
If this is working, then 07-array/push6-fp.bal, should pass. Please check that it is and rename to without the `f`.

KavinduZoysa:
> If this is working, then 07-array/push6-fp.bal, should pass. Please check that it is and rename to without the `f`.

The test is passing. Renamed the test.

PR #442, opened 2021-09-13
Flip meaning of tagged int's 62th bit to immediate

### LINKED ISSUE #302: Fix string/int heap/immediate inconsistency
The tagged int implementation uses 0x20 bit of upper byte to say the int is on the heap;  the string implementation uses it to say the string is immediate (not on the heap).

manuranga:
`taggedInt & 0x20000000 != 0` used to mean that the int is on heap, with this change it to mean that it's not on the heap, ie immediate.

Fixes #302

### DIFF
--- a/runtime/tests/tagged_ptr.c
+++ b/runtime/tests/tagged_ptr.c
@@ -45,7 +45,7 @@ void testTaggedToInt() {
     // int on stack
     for(int64_t i=0; i <= NTESTS; i ++) {
         int64_t val = rand();
-        TaggedPtr ptr = bitsToTaggedPtr(val);
+        TaggedPtr ptr = bitsToTaggedPtr(val | IMMEDIATE_FLAG);
         int64_t prtVal = taggedToInt(ptr); //? work only with positive values
         assert(prtVal == val);
     }
@@ -56,7 +56,6 @@ void testTaggedToInt() {
 	int expectedValue = rand();
         *val = expectedValue;
         uint64_t addr = (uint64_t) val;
-        addr = addr | ((uint64_t)FLAG_INT_ON_HEAP << TAG_SHIFT);
         TaggedPtr ptr = bitsToTaggedPtr(addr);
         uint64_t ptrVal = taggedToInt(ptr);
         assert(ptrVal == expectedValue);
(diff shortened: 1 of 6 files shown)


runtime/tests/tagged_ptr.c
```diff
@@ -45,7 +45,7 @@ void testTaggedToInt() {
     // int on stack
     for(int64_t i=0; i <= NTESTS; i ++) {
         int64_t val = rand();
-        TaggedPtr ptr = bitsToTaggedPtr(val);
+        TaggedPtr ptr = bitsToTaggedPtr(val | IMMEDIATE_FLAG);
         int64_t prtVal = taggedToInt(ptr); //? work only with positive values
         assert(prtVal == val);
```
manuranga, line 50: I don't think this test is correct, `rand()` doesn't cover full `int64_t` range. We also need to consider intimidate int range when creating the int.

manuranga: I'll fix it.

runtime/tests/tagged_ptr.c
```diff
@@ -56,7 +56,6 @@ void testTaggedToInt() {
 	int expectedValue = rand();
         *val = expectedValue;
         uint64_t addr = (uint64_t) val;
-        addr = addr | ((uint64_t)FLAG_INT_ON_HEAP << TAG_SHIFT);
         TaggedPtr ptr = bitsToTaggedPtr(addr);
         uint64_t ptrVal = taggedToInt(ptr);
```
manuranga, line 60: Here also we are missing `TAG_INT` @heshanpadmasiri 
I'll fix it.

manuranga:
@jclark I am not sure what needs to be changed in `build.bal`. Is it the `buildHasTagInSet` function?

jclark:
> @jclark I am not sure what needs to be changed in `build.bal`. Is it the `buildHasTagInSet` function?

buildHasTagInSet looks OK.

I don't see anything obvious that needs changing. The tests should fail if we have missed anything.

manuranga:
In that case please consider this ready to be reviewed. I'll improve the tests in a separate pr.

jclark (approved):
LGTM

Issue #412, opened 2021-09-04
Allow statements to use full type descriptors


jclark:
Currently we parse type descriptors in two ways:

- full type descriptors are allowed in type definitions
- "inline" type descriptors are allowed in statements

We need to improve this so that statements can contain full type descriptors.  This requires fixing both parsing and type resolution.

Part of the necessary work for type resolution was done in #339.

jclark:
The fix for #411 fixes type resolution. The remaining part is parsing, which requires unbounded lookahead, for example:

```
(((x))).clone();
(((x))) n;
```

or

```
x[10] = 17;
x[10] foo;
```

jclark:
Unbounded lookahead will be easier with #408.

jclark:
With #454, we can do unbounded lookahead:

- make a new Tokenizer
- save the state of the current tokenizer
- restore the state into the new Tokenizer
- use the new Tokenizer to lookahead as much as needed

jclark:
So what we have to figure out is, given this ability, what is the best way to figure out whether we have a local variable declaration or something else.  This requires looking at what follows the type descriptor (or what would be the type descriptor if it's a local variable declaration).  In the local variable declaration case it will be something that can start a binding-pattern (e.g. an identifier); in other cases, it will be something more operator-like (e.g. `.`).  The first task is to analyze the grammar and make this precise.

After that, we can figure out the strategy to actually implement it.

jclark:
Note that for subset 8 (in #471), we really ought to be able to handle

m:T v = E;

manuranga:
Meeting notes:

1) We should try to make do with left factoring when it's reasonable, instead of save/restore.
2) Qualified identifier should be done first to facilitate #438

jclark:
Also consider a peek method on Tokenizer that would turn a FragCode into a SingleCharDelim|MultiCharDelim|Keyword|VariableLengthToken

manuranga:
I plan to add a `string? modPrefix` to `VarRefExpr` and `TypeDescRef`.

jclark:
Just call it `prefix`.

jclark:
I think we're starting to blur the lines between this and #471.

This issue is really about allowing arbitrary type descriptors in local variable declaration statements (by doing lookahead).  So let's make a separate issue (#473) just for parsing of qualified identifier in variable references and type references.

manuranga:
[...]

PR #454, opened 2021-09-15
Provide two-stage interface to front module

### LINKED ISSUE #451: Provide two-stage interface to front module
In order to support #438, we need to provide a two stage interface to the front module:

1. In the first stage, it provides a list of the module’s imports. In this stage, it runs the parser/tokenizer only up till the end of the import declarations. It doesn’t yet build a complete AST.
2. In the second stage, it generates a BIR module; it is provided with the public interface to each of the modules that it imports. In this stage , it builds the complete AST.

manuranga:
Provide two-stage interface to front module
Fixes  #451

### DIFF
--- a/compiler/modules/front.syntax/parse.bal
+++ b/compiler/modules/front.syntax/parse.bal
@@ -6,20 +6,45 @@ public type FilePath record {|
     string? directory = ();
 |};
 
-public function parseModulePart(string[] lines, FilePath path, int partIndex) returns ModulePart|err:Syntax {
+public function scanModulePart(string[] lines, FilePath path, int partIndex) returns ScannedModulePart|err:Syntax {
     SourceFile file = createSourceFile(path, lines);
     Tokenizer tok = new (file);
     check tok.advance();
-    ModulePart part = {
-        file,
-        partIndex,
-        defns: [],
-        importDecls: check parseImportDecls(tok)
-    };
-    while tok.current() != () {
-        part.defns.push(check parseModuleDecl(tok, part));
+    return new(file, check parseImportDecls(tok).cloneReadOnly(), tok.save(), partIndex);
+}
+
+public readonly class ScannedModulePart {
+    private TokenizerState tokState;
+    private ImportDecl[] imports;
+    private SourceFile file;
+    private int partIndex;
+
+    function init(SourceFile file, readonly & ImportDecl[] imports, TokenizerState tokState, int partIndex) {
+        self.file = file;
+        self.imports = imports;
+        self.tokState = tokState;
+        self.partIndex = partIndex;
     }
-    return part;
+
+    public function getImportDecls() returns ImportDecl[]|err:Syntax {
+        return self.imports;
+    }
+
+    public function parse() returns ModulePart|err:Syntax {
+        Tokenizer tok = new(self.file);
+        tok.restore(self.tokState);
+        ModuleLevelDefn[] defns = [];
+        ModulePart part = {
+            file: self.file,
+            partIndex: self.partIndex,
+            defns,
+            importDecls: self.imports
+        };
+        while tok.current() != () {
+            defns.push(check parseModuleDecl(tok, part));
+        }
+        return part;
+    }
 }
 
 public function parseExpression(string[] lines, FilePath path) returns Expr|err:Syntax {
@@ -34,7 +59,8 @@ public function parseExpression(string[] lines, FilePath path) returns Expr|err:
 }
 
 function createSourceFile(FilePath path, string[] lines) returns SourceFile {
-    return new(path, scanLines(lines));
+    ScannedSourceFile file = new(path, scanLines(lines));
+    return file;
 }
 
 function parseImportDecls(Tokenizer tok) returns ImportDecl[]|err:Syntax {
--- a/compiler/modules/front.syntax/token.bal
+++ b/compiler/modules/front.syntax/token.bal
@@ -1,6 +1,6 @@
 import wso2/nballerina.err;
 
-type Token FixedToken|VariableLengthToken;
+type Token FixedToken|readonly & VariableLengthToken;
 type FixedToken SingleCharDelim|MultiCharDelim|Keyword;
 
 const IDENTIFIER = 0;
@@ -92,6 +92,16 @@ type Mode MODE_NORMAL|MODE_TYPE_DESC;
 function toToken(string t) returns Token {
     return <Token>t;
 }
+
+type TokenizerState readonly & record {
+    int lineIndex;
+    int fragCodeIndex;
+    int codePointIndex;
+    int fragmentIndex;
+    int tokenStartCodePointIndex;
+    Mode mode;
+    Token? curTok;
+};
 
 class Tokenizer {
     private final ScannedLine[] lines;
@@ -276,13 +286,48 @@ class Tokenizer {
         return err:syntax(msg, loc=err:location(self.file, self.currentPos()));
     }
 
+    function save() returns TokenizerState {
+        return {
+            lineIndex : self.lineIndex,
+            fragCodeIndex : self.fragCodeIndex,
+            codePointIndex : self.codePointIndex,
+            fragmentIndex : self.fragmentIndex,
+            tokenStartCodePointIndex : self.tokenStartCodePointIndex,
+            mode : self.mode,
+            curTok : self.curTok
+        };
+    }
+
+    function restore(TokenizerState s) {
+        self.lineIndex = s.lineIndex;
+        self.fragCodeIndex = s.fragCodeIndex;
+        self.codePointIndex = s.codePointIndex;
+        self.fragmentIndex = s.fragmentIndex;
+        self.tokenStartCodePointIndex = s.tokenStartCodePointIndex;
+        self.mode = s.mode;
+        self.curTok = s.curTok;
+
+        ScannedLine scannedLine = self.lines[self.lineIndex - 1];
+        self.fragCodes = scannedLine.fragCodes;
+        self.fragments = scannedLine.fragments;
+    }
+
 }
 
 function createPosition(int line, int column) returns Position {
     return (line << 32) | column;
 }
 
-public readonly class SourceFile {
+// JBUG when SourceFile is replaced with the ScannedSourceFile, get `incompatible types`
+public type SourceFile readonly & object {
+    // JUBG should included *err:File here and include this in ScannedSourceFile
+    public function filename() returns string;
+    public function directory() returns string?;
+    public function lineColumn(Position pos) returns err:LineColumn;
+    function scannedLines() returns readonly & ScannedLine[];
+};
+
+readonly class ScannedSourceFile {
     *err:File;
     private string fn;
     private string? dir;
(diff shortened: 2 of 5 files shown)


compiler/modules/front.syntax/parse.bal
```diff
@@ -7,20 +7,44 @@ public type FilePath record {|
...
+        return imports;
+    }
+
+    public function parse() returns ModulePart|err:Syntax {
+        Tokenizer? t = self.tok;
+        if t is () {
+            panic error("parseImports() must be called before parse()");
+        }
+
+        Tokenizer tok = <Tokenizer>t;
```
manuranga: Is there a better way do this?

compiler/modules/front/front.bal
```diff
@@ -133,20 +133,46 @@ type LoadedSourcePart record {|
...
+
+    function init(s:Parser[] parsers,s:ImportDecl[] imports , bir:ModuleId id) {
+        self.parsers = parsers;
+        self.imports = imports;
+        self.id = id;
+    }
+
+    function getImports() returns s:ImportDecl[] => self.imports;
+
+    public function load(t:Env env) returns ResolvedModule|err:Any|io:Error {
```
jclark: Call this `resolve`.

compiler/modules/front.syntax/parse.bal
jclark: Move parseImports back out of Parser. Call parseImports before you construct the parser. Construct the parser with a non-nil Tokenizer and a ModulePart that contains the list of imports. Provide a function getImports() that gets the imports from the module part.

Parser is not quite the right word. It's more like PartiallyParsedModule, but that's too long. Perhaps you can think of a better word.

jclark: I think it's maybe not too difficult to make this readonly.

You can save Tokenizer state just by doing cloneReadOnly().  Then you just need to add a method on Tokenizer `function restore(Tokenizer t)` which resets the non-final fields from `t` (panicking if the final fields are not `===`). I expect we will need something like this for #412.

jclark (changes requested):

compiler/modules/front.syntax/parse.bal
jclark: Actually there's no need for getImports. Just don't make `part` private. We are not trying to encapsulate parsing in this class.

jclark: Should be `getImportDecls()`

manuranga: >You can save Tokenizer state just by doing cloneReadOnly(). 

I am not sure I understand, I though in Ballerina `cloneReadOnly()` doesn't work on objects.

Should I add methods to Tokenizer to dump state `function state() TokenizerState` and restore form it `function restore(TokenizerState s)`?

manuranga: To clarify:

```ballerina
    Tokenizer tok = new (file);
    readonly & Tokenizer t = tok.cloneReadOnly()
```
gives
```
undefined method 'cloneReadOnly' in object 'wso2/nballerina.front.syntax:0.1.0:Tokenizer'
```

manuranga: For the record, we decided to add a method to save state.

manuranga:
I've run into an issue where I can't assign `SourceFile` to a `readonly & SourceFile` type variable. I expect this to be possible given that `SourceFile` class itself is readonly. Is this a jballerina bug?

jclark:
> I've run into an issue where I can't assign `SourceFile` to a `readonly & SourceFile` type variable. I expect this to be possible given that `SourceFile` class itself is readonly. Is this a jballerina bug?

Sounds like it: `SourceFile` and `readonly & SourceFile` are the same type. I've run into this sort of thing before. Sometimes you can workaround by introducing a separate temp variable.

Why do you need `readonly & SourceFile` type variable? If you let me see the code, I can suggest some possible workarounds.

jclark:
If you cannot get the readonly side working, send the pull request without this and I will have a go.

compiler/modules/front.syntax/parse.bal
```diff
@@ -7,20 +7,49 @@ public type FilePath record {|
...
+// xxx alternative names ModulePartPrecursor ModulePrecursor ResolvableModulePart ScannedModulePart
+public class PreModulePart {
+    private TokenizerState tokState;
+    private ImportDecl[] imports;
+    private SourceFile file;
+    private int partIndex;
+
+    function init(SourceFile file, readonly & ImportDecl[] imports, TokenizerState tokState, int partIndex) {
+        // xxx JBUG when class is readonly, gives error `incompatible types`
+        self.file = file;
```
manuranga, line 23: @jclark This is where I face the issue.

compiler/modules/front.syntax/token.bal
```diff
@@ -92,6 +92,16 @@ function toToken(string t) returns Token {
...
 }
 
+type TokenizerState readonly & record {
+    int lineIndex;
+    int fragCodeIndex;
+    int codePointIndex;
+    int fragmentIndex;
+    int tokenStartCodePointIndex;
+    Mode mode;
+    Token? curTok;
```
manuranga, line 103: Another options is to have `Token curTok;` and change `TokenizerState tokState` to `TokenizerState? tokState` in `PreModulePart`'s `init`.

compiler/modules/front.syntax/parse.bal
```diff
@@ -7,20 +7,49 @@ public type FilePath record {|
...
-        importDecl: check parseImportDecl(tok)
-    };
-    while tok.current() != () {
-        part.defns.push(check parseModuleDecl(tok, part));
+    // xxx inline after #443, and check if cloneReadOnly can be avoided
+    ImportDecl? im = check parseImportDecl(tok);
+    return new(file, im == () ? [] : [im.cloneReadOnly()], tok.save(), partIndex);
+}
+
+// xxx alternative names ModulePartPrecursor ModulePrecursor ResolvableModulePart ScannedModulePart
```
jclark: I think ScannedModulePart is the best. Then you can call the function scanModulePart.

compiler/modules/front.syntax/token.bal
```diff
@@ -275,6 +285,30 @@ class Tokenizer {
         return err:syntax(msg, loc=err:location(self.file, self.currentPos()));
     }
 
+    function save() returns TokenizerState {
+        return { lineIndex : self.lineIndex,
```
jclark: start lineIndex on a separate line

compiler/modules/front.syntax/token.bal
```diff
@@ -275,6 +285,30 @@ class Tokenizer {
...
+                 fragCodeIndex : self.fragCodeIndex,
+                 codePointIndex : self.codePointIndex,
+                 fragmentIndex : self.fragmentIndex,
+                 tokenStartCodePointIndex : self.tokenStartCodePointIndex,
+                 mode : self.mode,
+                 curTok : self.curTok.cloneReadOnly() };
```
jclark: Better to make Token be readonly, which requires changing VariableLengthToken.

compiler/modules/front.syntax/token.bal
jclark: What you have seems fine to me.

compiler/modules/front/front.bal
```diff
@@ -133,20 +133,46 @@ type LoadedSourcePart record {|
...
+        s:ScannedModulePart p = check s:scanModulePart(loaded.lines, loaded.path, i);
+        parts.push(p);
+        imports.push(... check p.getImportDecls());
+    }
+    ScannedModule mod = new(parts, imports, id);
+    return mod;
+}
+
+public class ScannedModule {
+    private s:ScannedModulePart[] parsers;
```
manuranga: Need to rename this.

manuranga: fixed

compiler/modules/front.syntax/parse.bal
jclark: Try declaring `lines` in `SourceFile` as `readonly & ScannedLine[]`.

manuranga: This didn't work, but found a different workaround below, please take a look.

compiler/modules/front.syntax/token.bal
manuranga: fixed

compiler/modules/front.syntax/token.bal
```diff
@@ -275,14 +285,50 @@ class Tokenizer {
...
+    }
+
 }
 
 function createPosition(int line, int column) returns Position {
     return (line << 32) | column;
 }
 
-public readonly class SourceFile {
+// JBUG when SourceFile is replaced with the ScannedSourceFile, get `incompatible types`
```
manuranga, line 321: This is the workaround, but I got some errors when I tried to use type inclusions. But it still works without that.

compiler/modules/front.syntax/parse.bal
```diff
@@ -7,20 +7,49 @@ public type FilePath record {|
...
+    public function getImportDecls() returns ImportDecl[]|err:Syntax {
+        return self.imports;
+    }
+
+    public function parse() returns ModulePart|err:Syntax {
+        Tokenizer tok = new(self.file);
+        tok.restore(self.tokState);
+        ModuleLevelDefn[] defns = [];
+        // inline after #443
+        ImportDecl? im = self.imports.length() > 0 ? self.imports[0] : ();
```
jclark: Prefer `decl` to `im`

compiler/modules/front.syntax/token.bal
```diff
@@ -275,14 +285,50 @@ class Tokenizer {
...
+public type SourceFile readonly & object {
+    // xxx should err:File be included here ?
```
jclark: Yes, you should include err:File

compiler/modules/front/front.bal
```diff
@@ -133,20 +133,46 @@ type LoadedSourcePart record {|
...
+    ScannedModule mod = new(parts, imports, id);
+    return mod;
+}
+
+public class ScannedModule {
+    private s:ScannedModulePart[] scannedParts;
+    private s:ImportDecl[] imports;
+    private bir:ModuleId id;
+
+    function init(s:ScannedModulePart[] scannedParts,s:ImportDecl[] imports , bir:ModuleId id) {
```
jclark: Couple of spaces wrong

jclark (changes requested):

compiler/modules/front.syntax/token.bal
manuranga: it's giving me `object type inclusion cannot be used with a 'readonly' type descriptor in an 'object' type descriptor` error. JBUG?

jclark: Probably. Let's get this in and I will try to find a better workaround.

compiler/modules/front.syntax/parse.bal
```diff
@@ -34,7 +59,8 @@ public function parseExpression(string[] lines, FilePath path) returns Expr|err:
 }
 
 function createSourceFile(FilePath path, string[] lines) returns SourceFile {
-    return new(path, scanLines(lines));
+    ScannedSourceFile x = new(path, scanLines(lines));
```
manuranga: not a good name, will change

PR #455, opened 2021-09-15
Optimize const small string equality

### LINKED ISSUE #276: Generate better LLVM IR for equality when one operand is constant
We can generate much better for code equality when we know one operand is constant.

In particular, for small strings we can generate a pointer comparison.

Relates to #267 and #264

KavinduZoysa:
## Purpose
Fixes #276

### DIFF
--- a/compiler/modules/nback/build.bal
+++ b/compiler/modules/nback/build.bal
@@ -1281,7 +1281,7 @@ function buildEquality(llvm:Builder builder, Scaffold scaffold, bir:EqualityInsn
                 return buildStoreBoolean(builder, scaffold, builder.iCmp(op, lhsValue, rhsValue), result);
             }
             else if reprIsString(lhsRepr) && reprIsString(rhsRepr) {
-                return buildEqualStringString(builder, scaffold, op, <llvm:PointerValue>lhsValue, <llvm:PointerValue>rhsValue, result);
+                return generateStringEquality(builder, scaffold, op, insn, lhsValue, rhsValue, result);
             }
             else {
                 return buildEqualTaggedTagged(builder, scaffold, exact, op, <llvm:PointerValue>lhsValue, <llvm:PointerValue>rhsValue, result);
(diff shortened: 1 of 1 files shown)


compiler/modules/nback/build.bal
```diff
@@ -1361,6 +1374,10 @@ function reprIsImmediate(Repr repr) returns boolean {
     return !(repr is TaggedRepr) || (repr.subtype & ~(t:NIL|t:BOOLEAN)) == 0;
 }
 
+function isSmallString(int nCodePoints, byte[] bytes, int nBytes) returns boolean {
```
jclark: Move isSmallString to after addStringDefn.

compiler/modules/nback/build.bal
```diff
@@ -1281,7 +1281,7 @@ function buildEquality(llvm:Builder builder, Scaffold scaffold, bir:EqualityInsn
                 return buildStoreBoolean(builder, scaffold, builder.iCmp(op, lhsValue, rhsValue), result);
             }
             else if reprIsString(lhsRepr) && reprIsString(rhsRepr) {
-                return buildEqualStringString(builder, scaffold, op, <llvm:PointerValue>lhsValue, <llvm:PointerValue>rhsValue, result);
+                return generateStringEquality(builder, scaffold, op, insn, lhsValue, rhsValue, result);
```
jclark: Instead of `generateStringEquality`, separate out a `function isAnyOperandSmallString(bir:Operand[]) returns boolean`.

compiler/modules/nback/build.bal
```diff
@@ -1343,6 +1343,19 @@ function buildEqualFloat(llvm:Builder builder, Scaffold scaffold, boolean exact,
     return buildStoreBoolean(builder, scaffold, b, reg);
 }
 
+function generateStringEquality(llvm:Builder builder, Scaffold scaffold, CmpEqOp op, bir:EqualityInsn insn, llvm:Value lhsValue, llvm:Value rhsValue, bir:Register result) {
+    foreach var operand in insn.operands {
+        if !(operand is string) {
```
jclark: Do not use `continue` so you can avoid the cast later on.

compiler/modules/nback/build.bal
```diff
@@ -1343,6 +1343,19 @@ function buildEqualFloat(llvm:Builder builder, Scaffold scaffold, boolean exact,
...
+            continue;
+        }
+        string str = <string>operand;
+        if isSmallString(str.length(), str.toBytes(), str.toBytes().length()) {
```
jclark: Don't call str.toBytes twice: each call does the conversion to UTF-8.

jclark (changes requested):

compiler/modules/nback/build.bal
KavinduZoysa: Changed

compiler/modules/nback/build.bal
KavinduZoysa: Changed

compiler/modules/nback/build.bal
KavinduZoysa: Fixed

compiler/modules/nback/build.bal
KavinduZoysa: Fixed

PR #467, opened 2021-09-17
Support `check` and `checkpanic` in call stmts

manuranga:
Support `check` and `checkpanic` in call stmts

### DIFF
--- a/compiler/modules/front.syntax/ast.bal
+++ b/compiler/modules/front.syntax/ast.bal
@@ -47,7 +47,7 @@ public type ConstDefn record {|
 |};
 
 public type Stmt VarDeclStmt|AssignStmt|CallStmt|ReturnStmt|IfElseStmt|MatchStmt|WhileStmt|ForeachStmt|BreakStmt|ContinueStmt|CompoundAssignStmt|PanicStmt;
-public type CallStmt FunctionCallExpr|MethodCallExpr;
+public type CallStmt FunctionCallExpr|MethodCallExpr|CheckingStmt;
 public type Expr NumericLiteralExpr|ConstValueExpr|FloatZeroExpr|BinaryExpr|UnaryExpr|CheckingExpr|FunctionCallExpr|MethodCallExpr|VarRefExpr|TypeCastExpr|TypeTestExpr|ConstructorExpr|MemberAccessExpr;
 public type ConstructorExpr ListConstructorExpr|MappingConstructorExpr|ErrorConstructorExpr;
 public type SimpleConstExpr ConstValueExpr|VarRefExpr|IntLiteralExpr|SimpleConstNegateExpr;
@@ -201,6 +201,13 @@ public type CheckingKeyword "check"|"checkpanic";
 public type CheckingExpr record {|
     CheckingKeyword checkingKeyword;
     Expr operand;
+|};
+
+public type CheckingStmt record {|
+    // JBUG can't include CheckingExpr
+    // *CheckingExpr;
+    CheckingKeyword checkingKeyword;
+    CallStmt operand;
 |};
 
 public type ListConstructorExpr record {|
--- a/compiler/modules/front.syntax/parseStmt.bal
+++ b/compiler/modules/front.syntax/parseStmt.bal
@@ -65,6 +65,11 @@ function parseStmt(Tokenizer tok) returns Stmt|err:Syntax {
         "error" => {
             return parseErrorStmt(tok);
         }
+        "check"|"checkpanic" => {
+            check tok.advance();
+            // JBUG cast
+            return finishCheckingCallStmt(tok, <CheckingKeyword>cur);
+        }
         var td if td is InlineBuiltinTypeDesc|"map" => {
             return parseVarDeclStmt(tok);
         }
@@ -152,7 +157,7 @@ function parseErrorStmt(Tokenizer tok) returns Stmt|err:Syntax {
     }
 }
 
-function parseMethodCallStmt(Tokenizer tok) returns Stmt|err:Syntax {
+function parseMethodCallStmt(Tokenizer tok) returns MethodCallExpr|err:Syntax {
     Expr expr = check parsePrimaryExpr(tok);
     if expr is MethodCallExpr {
         check tok.expect(";");
@@ -175,6 +180,23 @@ function finishCallStmt(Tokenizer tok, CallStmt expr) returns Stmt|err:Syntax {
     }
     check tok.expect(";");
     return stmt;
+}
+
+function finishCheckingCallStmt(Tokenizer tok, CheckingKeyword checkingKeyword) returns CallStmt|err:Syntax {
+    Token? t = tok.current();
+    if t is "check"|"checkpanic" {
+        check tok.advance();
+        return { checkingKeyword, operand: check finishCheckingCallStmt(tok, t) };
+    }
+    else if t == "(" {
+        return { checkingKeyword, operand: check parseMethodCallStmt(tok) };
+    }
+    Expr operand = check parsePrimaryExpr(tok);
+    if operand is FunctionCallExpr|MethodCallExpr {
+        check tok.expect(";");
+        return { checkingKeyword, operand };
+    }
+    return parseError(tok, "function call, method call or checking expression expected");
 }
 
 function finishAssignStmt(Tokenizer tok, LExpr lValue) returns AssignStmt|err:Syntax {
--- a/compiler/modules/front/codeGen.bal
+++ b/compiler/modules/front/codeGen.bal
@@ -896,14 +896,17 @@ function codeGenCompoundAssignToMember(CodeGenContext cx, bir:BasicBlock bb, Env
 }
 
 function codeGenCallStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environment env, s:CallStmt stmt) returns CodeGenError|StmtEffect {
-    // stmt is FunctionCallExpr or s:MethodCallExpr
+    // stmt is FunctionCallExpr, s:MethodCallExpr or s:CheckingStmt
     bir:Register reg;
     bir:BasicBlock nextBlock;
     if stmt is s:FunctionCallExpr {
-        { result: reg, block: nextBlock } = check codeGenFunctionCall(cx, <bir:BasicBlock>startBlock, env, stmt);
+        { result: reg, block: nextBlock } = check codeGenFunctionCall(cx, startBlock, env, stmt);
+    }
+    else if stmt is s:MethodCallExpr {
+        { result: reg, block: nextBlock } = check codeGenMethodCall(cx, startBlock, env, stmt);
     }
     else {
-        { result: reg, block: nextBlock } = check codeGenMethodCall(cx, <bir:BasicBlock>startBlock, env, stmt);
+        return check codeGenCheckingStmt(cx, startBlock, env, stmt.checkingKeyword, stmt.operand);
     }
     if reg.semType !== t:NIL {
         return cx.semanticErr("return type of function or method in call statement must be nil");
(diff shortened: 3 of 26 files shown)


compiler/testSuite/08-error/check7-v.bal
```diff
@@ -0,0 +1,7 @@
+public function main() {
+    checkpanic newError();
+}
+
+function newError() returns error {
+    return error("failed successfully");
+}
```
manuranga: This is PR is not done yet, I need to fix case.
From the spec, this is a valid case, even though it's failing in jBallerina beta2.

manuranga: I plan to fix this by passing a boolean to `codeGenCheckingExpr`

compiler/modules/front/codeGen.bal
```diff
@@ -991,7 +1000,7 @@ function codeGenExpr(CodeGenContext cx, bir:BasicBlock bb, Environment env, s:Ex
             return { result, block: nextBlock, narrowing };
         }
         var { checkingKeyword, operand } => {
-            return codeGenCheckingExpr(cx, bb, env, checkingKeyword, operand);
+            return <CodeGenError|RegExprEffect> codeGenCheckingExpr(cx, bb, env, checkingKeyword, operand, false);
```
manuranga: Possible refactoring: rename current `codeGenCheckingExpr` to `codeGenChecking`, have separate `codeGenCheckingExpr() returns RegExprEffect` and `codeGenCheckingStmt() returns RegExprEffect|StmtEffect`, and move above cast to former.

compiler/modules/front.syntax/ast.bal
```diff
@@ -46,7 +46,7 @@ public type ConstDefn record {|
 |};
 
 public type Stmt VarDeclStmt|AssignStmt|CallStmt|ReturnStmt|IfElseStmt|MatchStmt|WhileStmt|ForeachStmt|BreakStmt|ContinueStmt|CompoundAssignStmt|PanicStmt;
-public type CallStmt FunctionCallExpr|MethodCallExpr;
+public type CallStmt FunctionCallExpr|MethodCallExpr|CheckingExpr;
```
jclark: Can we subtype of CheckingExpr so as to capture the kind of CheckingExpr allowed in CallStmt?

compiler/modules/front/codeGen.bal
```diff
@@ -896,15 +896,24 @@ function codeGenCompoundAssignToMember(CodeGenContext cx, bir:BasicBlock bb, Env
...
     if stmt is s:FunctionCallExpr {
         { result: reg, block: nextBlock } = check codeGenFunctionCall(cx, <bir:BasicBlock>startBlock, env, stmt);
     }
-    else {
+    else if stmt is s:MethodCallExpr {
         { result: reg, block: nextBlock } = check codeGenMethodCall(cx, <bir:BasicBlock>startBlock, env, stmt);
     }
+    else {
+        var effect = check codeGenCheckingExpr(cx, <bir:BasicBlock>startBlock, env, stmt.checkingKeyword, stmt.operand, true);
+        if effect is StmtEffect {
```
jclark: This doesn't make much sense to me: expression effects and statement effects are different things. This will be resolved by the change I suggest below to codeGenCheckingExpr.

compiler/modules/front.syntax/parseStmt.bal
```diff
@@ -65,6 +65,10 @@ function parseStmt(Tokenizer tok) returns Stmt|err:Syntax {
         "error" => {
             return parseErrorStmt(tok);
         }
+        var kw if kw is CheckingKeyword => {
```
jclark: I prefer "check"|"checkpanic" in this case

compiler/modules/front/codeGen.bal
```diff
@@ -991,7 +1000,7 @@ function codeGenExpr(CodeGenContext cx, bir:BasicBlock bb, Environment env, s:Ex
             return { result, block: nextBlock, narrowing };
         }
         var { checkingKeyword, operand } => {
-            return codeGenCheckingExpr(cx, bb, env, checkingKeyword, operand);
+            return <CodeGenError|RegExprEffect> codeGenCheckingExpr(cx, bb, env, checkingKeyword, operand, false);
         }
```
jclark: Why do we need a cast here?

compiler/modules/front/codeGen.bal
```diff
@@ -1311,7 +1320,7 @@ function codeGenTypeTest(CodeGenContext cx, bir:BasicBlock bb, Environment env,
     return { result, block: nextBlock, narrowing };   
 }
 
-function codeGenCheckingExpr(CodeGenContext cx, bir:BasicBlock bb, Environment env, s:CheckingKeyword checkingKeyword, s:Expr expr) returns CodeGenError|ExprEffect {
+function codeGenCheckingExpr(CodeGenContext cx, bir:BasicBlock bb, Environment env, s:CheckingKeyword checkingKeyword, s:Expr expr, boolean isStmt) returns CodeGenError|RegExprEffect|StmtEffect {
```
jclark: I think we can improve this. I don't like have something called codeGenCheckingExpr that handles both expressions and statements. I suggest we have both codeGenCheckingStmt and a codeGenCheckingExpr that both codeGen the subExpr and then feed into a common function. The codeGenCheckingStmt will check whether the operand is a subtype of t:ERROR and if so handle that case, before feeding into the common function.

compiler/modules/front/codeGen.bal
```diff
@@ -1360,6 +1364,17 @@ function codeGenCheckingExpr(CodeGenContext cx, bir:BasicBlock bb, Environment e
     }
 }
 
+function codeGenCheckingReturn(bir:BasicBlock bb, s:CheckingKeyword checkingKeyword, bir:Register operand) {
```
jclark: I don't think Return is a good word here, because only one case is a return.

jclark (changes requested):

compiler/modules/front.syntax/ast.bal
```diff
@@ -202,6 +202,13 @@ public type CheckingExpr record {|
     Expr operand;
 |};
 
+public type CheckingStmt record {|
+    // JBUG can't include CheckingExpr
+    // *CheckingExpr;
+    CheckingKeyword checkingKeyword;
+    FunctionCallExpr|MethodCallExpr|CheckingExpr operand;
```
jclark: Shouldn't the type be `CallStmt operand;`? This allows a CheckingExpr that contains anything.

compiler/modules/front.syntax/parseStmt.bal
```diff
@@ -181,6 +186,23 @@ function finishCallStmt(Tokenizer tok, CallStmt expr) returns Stmt|err:Syntax {
...
+    Token? t = tok.current();
+    if t is "check"|"checkpanic" {
+        check tok.advance();
+        return { checkingKeyword, operand: check finishCheckingCallStmt(tok, t) };
+    }
+    else if t == "(" {
+        return { checkingKeyword, operand: check parseMethodCallStmt(tok) };
+    }
+    Expr operand = check parsePrimaryExpr(tok);
+    if operand is CallStmt {
```
jclark: Clearer here to say FunctionCallExpr|MethodCallExpr.

compiler/modules/front/codeGen.bal
```diff
@@ -1311,52 +1314,89 @@ function codeGenTypeTest(CodeGenContext cx, bir:BasicBlock bb, Environment env,
...
+    bir:CondNarrowInsn narrowToOk = {
+        result,
+        operand,
+        basis: { insn: testInsnRef, result: false }
+    };
+    okBlock.insns.push(narrowToOk);
+    return { result, block: okBlock };
+}
+
+function codeGenCheckingTerminal(bir:BasicBlock bb, s:CheckingKeyword checkingKeyword, bir:Register operand) {
```
jclark: SInce this is adding a terminator insn, I suggest s/Terminal/Terminator/.

compiler/modules/front/codeGen.bal
```diff
@@ -1311,52 +1314,89 @@ function codeGenTypeTest(CodeGenContext cx, bir:BasicBlock bb, Environment env,
...
-        t:SemType okType = t:diff(operand.semType, t:ERROR);
-        if t:isNever(okType) {
-            // this has to be an error because otherwise the register would be of type never
-            // SUBSET fix this if we allow checking keyword as a call statement
+        // JBUG build fails when the order of binding pattern is reversed
+        { block, result: _ } = check codeGenCheckingCond(cx, nextBlock, operand, errorType, checkingKeyword, resultType);
+    }
+    // resultType === NEVER case is already handled
+    if resultType !== t:NIL {
+        return cx.semanticErr(`type of ${checkingKeyword} statement must be nil or never`);
```
jclark: It's better to phrase this in terms of the type of the operand i.e. operand of ${checkingKeyword} must be a subtype of error?

jclark (commented):
Looking good. Just a few tweaks.

compiler/modules/front.syntax/ast.bal
manuranga: Yes, that is more in line with the spec. I did it this way thinking then it wouldn't be a subtype of `CheckingExpr` because jBallerina gave an error, I guess due to structural typing it has to be a subtype.

Let me change it.

compiler/modules/front.syntax/parseStmt.bal
```diff
@@ -181,6 +186,23 @@ function finishCallStmt(Tokenizer tok, CallStmt expr) returns Stmt|err:Syntax {
...
+    }
+    else if t == "(" {
+        return { checkingKeyword, operand: check parseMethodCallStmt(tok) };
+    }
+    Expr operand = check parsePrimaryExpr(tok);
+    if operand is FunctionCallExpr|MethodCallExpr {
+        check tok.expect(";");
+        return { checkingKeyword, operand };
+    }
+    return parseError(tok, "function call, method call or checking expression expected");
```
manuranga, line 199: If we are being precise, this has to be `s/checking expression/checking statment`, but I though it would be confusing.

compiler/modules/front/codeGen.bal
```diff
@@ -896,15 +896,18 @@ function codeGenCompoundAssignToMember(CodeGenContext cx, bir:BasicBlock bb, Env
 }
 
 function codeGenCallStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environment env, s:CallStmt stmt) returns CodeGenError|StmtEffect {
-    // stmt is FunctionCallExpr or s:MethodCallExpr
+    // stmt is FunctionCallExpr, s:MethodCallExpr or s:CheckingStmt
     bir:Register reg;
     bir:BasicBlock nextBlock;
     if stmt is s:FunctionCallExpr {
         { result: reg, block: nextBlock } = check codeGenFunctionCall(cx, <bir:BasicBlock>startBlock, env, stmt);
```
manuranga: @jclark this cast is unnecessary.

jclark: Then remove it...

compiler/modules/front.syntax/parseStmt.bal
manuranga: Even `"function call or method call expected"` might be enough, since it's unlikely one would want a checking statement with checking expression in it.

compiler/modules/front.syntax/ast.bal
```diff
@@ -203,6 +203,13 @@ public type CheckingExpr record {|
     Expr operand;
 |};
 
+public type CheckingStmt record {|
+    // JBUG can't include CheckingExpr
+    // *CheckingExpr;
+    CheckingKeyword checkingKeyword;
+    FunctionCallExpr|MethodCallExpr|CheckingStmt operand;
```
jclark: This type is equivalent to CallStmt

jclark (changes requested):

Issue #448, opened 2021-09-14
Enhance mangling to handle module names


jclark:
For subset 8, we need to be able to handle public names with

- an org that is either empty or `ballerina`
- arbitrary list of module names

We need to distinguish these from non-public names.

This is part of #438.

jclark:
C++ mangling rules here: https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangling

jclark:
Unquoted identifiers in LLVM must match `[-a-zA-Z$._][-a-zA-Z$._0-9]*` 

https://llvm.org/docs/LangRef.html#identifiers

jclark:
We will rely on the proposed spec restrictions here: https://github.com/ballerina-platform/ballerina-spec/issues/791#issuecomment-858593020

jclark:
What we are currently doing is described here: https://github.com/ballerina-platform/nballerina/issues/43#issuecomment-859163191

jclark:
GNU as syntax for names is described here: https://sourceware.org/binutils/docs/as/Symbol-Names.html#Symbol-Names

> Symbol names begin with a letter or with one of ‘._’. On most machines, you can also use $ in symbol names; exceptions are noted in Machine Dependencies. That character may be followed by any string of digits, letters, dollar signs (unless otherwise noted for a particular target machine), and underscores.

jclark:
Public name looks like:
```
"_B" digit+ (org-or-mod-name "__")+ local-name
```
The digits are a decimal number equals  to (number of org-or-mod-name - 1)*4 + K, where K is

 * 0 if there is a non-empty org name other than "ballerina" (so first org-or-mod-name is organization name)
 * 1 if there is an empty org name (so first org-or-mod-name is a module-name)
 * 2 if the org name is "ballerina" (so first org-or-mod-name is a module-name)
 * 3 if the org name is "ballerina", there are two or more module names and the first module name is "lang"

So following imports would start with following prefixes

```
ballerina/io => _B2io__
ballerina/lang.value => _B3value__
wso2/nballerina.err => _B8wso2__nballerina__err__
foo_bar.baz => _B5foo_bar__baz__
```

We will keep `$` for escaping individual characters.

jclark:
Alternative scheme:

```
"_B" digit* letter  (org-or-mod-name "__")* local-name
```

In this scheme, "well-known" org/module name sequences are assigned a lower-case ASCII letter
- `b` ballerina/
[...]

PR #468, opened 2021-09-18
Demangle function name at runtime

### LINKED ISSUE #458: Demangle function names in backtraces
When a panic prints out a function name, it should demangle it.

Mangling is described here: https://github.com/ballerina-platform/nballerina/issues/448#issuecomment-919844131

KavinduZoysa:
## Purpose
Fixes #458

### DIFF
--- a/runtime/error.c
+++ b/runtime/error.c
@@ -51,6 +51,12 @@ static void getSimpleBacktrace(SimpleBacktrace *p);
 static char *saveMessage(const char *msg);
 static void processPCs(GC PC* pcs, uint32_t nPCs, uint32_t start, int64_t lineNumber, BacktraceError *error);
 static void processInitialPC(PC pc, int64_t lineNumber, BacktraceStartLine *backtraceStartLine);
+static int getOrgNameOffset(const char *mangledName);
+static int getzQualifierOffset(const char *mangledName);
+static void printModuleName(const char *mangledName, bool *isFirstQualifier, FILE *fp);
+static void printLocalName(const char *mangledName, FILE *fp);
+static int printnQualifier(const char *mangledName, bool *isFirstQualifier, FILE *fp);
+static int getDecimalNumberLength(int n);
 
 TaggedPtr _bal_error_construct(TaggedPtr message, int64_t lineNumber) {
     SimpleBacktrace simpleBacktrace;
@@ -249,4 +255,84 @@ static char *saveMessage(const char *msg) {
 TaggedPtr BAL_LANG_ERROR_NAME(message)(TaggedPtr error) {
     ErrorPtr ep = taggedToPtr(error);
     return ep->message;
+}
+
+void _bal_print_mangled_name(const char *mangledName, FILE *fp) {
+    // If mangled name starts from _B_, the identifier is non-public
+    if (mangledName[2] == '_') {
+        fprintf(fp, "%s", mangledName + 3);
+        return;
+    }
+    // The org-name is not considered as a part of function name in backtrace.
+    int initialOffset = getOrgNameOffset(mangledName);
+
+    // Make sure parsing starts from the first character of
+    // nqual, local-name or mod-name
+    bool isFirstQualifier = true;
+    printModuleName(mangledName + initialOffset, &isFirstQualifier, fp);
+}
+
+static int getOrgNameOffset(const char *mangledName) {
+    // Offset starts from '_B'
+    int offset = 2;
+    if (mangledName[offset] == 'b') {
+        offset++;
+    }
+    return offset + getzQualifierOffset(mangledName + offset);
+}
+
+static int getzQualifierOffset(const char *mangledName) {
+    int offset = 0;
+    if (mangledName[offset] == '0') {
+        return ++offset;
+    }
+    int decimalNumber = atoi(mangledName + offset);
+    offset = offset + getDecimalNumberLength(decimalNumber);
+    if (mangledName[offset] == '_') {
+        offset++;
+    }
+    return offset + decimalNumber;
+}
+
+static void printModuleName(const char *mangledName, bool *isFirstQualifier, FILE *fp) {
+    if (mangledName[0] != 'm') {
+        int offset = printnQualifier(mangledName, isFirstQualifier, fp);
+        printLocalName(mangledName + offset, fp);
+        return;
+    }
+    // Adding '+ 1' because 'm' is available
+    int offset = printnQualifier(mangledName + 1, isFirstQualifier, fp) + 1;
+    printModuleName(mangledName + offset, isFirstQualifier, fp);
+}
+
+static void printLocalName(const char *mangledName, FILE *fp) {
+    fprintf(fp, ":%s", mangledName);
+}
+
+static int printnQualifier(const char *mangledName, bool *isFirstQualifier, FILE *fp) {
+    if (*isFirstQualifier == false) {
+        fputs(".", fp);
+    }
+    *isFirstQualifier = false;
+    int offset = 0;
+    int decimalNumber = atoi(mangledName);
+    offset = offset + getDecimalNumberLength(decimalNumber);
+    if (mangledName[offset] == '_') {
+        offset++;
+    }
+    fprintf(fp, "%.*s", decimalNumber, mangledName + offset);
+    offset = offset + decimalNumber;
+    return offset;
+}
+
+static int getDecimalNumberLength(int n) {
+    if (n == 0) {
+        return 1;
+    }
+    int res = 0;
+    while (n > 0) {
+        n = n / 10;
+        res++;
+    }
+    return res;
 }
--- a/runtime/error.c
+++ b/runtime/error.c
@@ -20,8 +20,10 @@
 
 #define PUBLIC_BALLERINA_NAME 1
 #define NON_PUBLIC_BALLERINA_NAME 2
-#define NON_BALLERINA_NAME 3
-#define _B_NON_BALLERINA_NAME 4
+#define NON_BALLERINA_NAME -1
+#define _B_NON_BALLERINA_NAME -2
+
+#define MAX_DEMANGLING_DECIMAL_LENGTH 4
 
 typedef int DemangleResult;
 
@@ -61,7 +63,6 @@ static void processInitialPC(PC pc, int64_t lineNumber, BacktraceStartLine *back
 static DemangleResult demangle(const char *mangledName, const char **localName, FILE *fp);
 static bool demangleModules(const char **mangledModules, FILE *fp);
 static bool demangleCountedName(const char **pp, const char **name);
-static void printUpto(const char *end, const char *start, FILE *fp);
 
 TaggedPtr _bal_error_construct(TaggedPtr message, int64_t lineNumber) {
     SimpleBacktrace simpleBacktrace;
@@ -290,7 +291,7 @@ static DemangleResult demangle(const char *mangledName, const char **localName, 
     }
 
     bool ballerinaOrg = false; 
-    if (mangledName[0] == 'b')  {
+    if (mangledName[0] == 'b') {
         ballerinaOrg = true;
         mangledName++;
     }
@@ -312,7 +313,7 @@ static DemangleResult demangle(const char *mangledName, const char **localName, 
             haveOrg = true;
         }
         if (org != NULL) {
-            printUpto(mangledName, org, fp);
+            fwrite(org, 1, mangledName - org, fp);
             haveOrg = true;
         }
         if (haveOrg) {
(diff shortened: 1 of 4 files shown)


KavinduZoysa:
Calling _bal_print_mangled_name is not implemented yet, because it depends on #432

runtime/tests/demangle.c
```diff
@@ -0,0 +1,93 @@
...
+    "_.io:println\n",
+    "2.io:println\n",
+    "foo_bar.io:println\n",
+    "foo_bar._bar.io:println\n",
+    "foo_bar.2bar.io:println\n",
+    "foo_bar.2bar._io:println\n",
+    "foo_bar.2bar.io:_println\n",
+    "foo_bar.2bar.io:_3println\n",
+    "foo_bar.2bar.new_nballerina_module:_3println\n",
+    "foo_bar.2bar.new_ballerina_module:_3println\n"
```
manuranga: Is it possible to avoid `\n`on each case, and to just check it in the test?

runtime/error.c
```diff
@@ -181,3 +187,83 @@ TaggedPtr BAL_LANG_ERROR_NAME(message)(TaggedPtr error) {
...
+static int getOrgNameOffset(const char *mangledName) {
+    // Offset starts from '_B'
+    int offset = 2;
+    if (mangledName[offset] == 'b') {
+        offset++;
+    }
+    return offset + getzQualifierOffset(mangledName + offset);
+}
+
+static int getzQualifierOffset(const char *mangledName) {
```
manuranga: Why `z` in `getz` ?

runtime/tests/demangle.c
KavinduZoysa: Yes, we can do that. The reason to use '\n' is to indicate the end of the demangled name.

runtime/error.c
KavinduZoysa: Actually I am using the keywords used in https://github.com/ballerina-platform/nballerina/issues/448#issuecomment-919844131

runtime/error.c
```diff
@@ -181,3 +187,83 @@ TaggedPtr BAL_LANG_ERROR_NAME(message)(TaggedPtr error) {
     ErrorPtr ep = taggedToPtr(error);
     return ep->message;
 }
+
+void _bal_print_mangled_name(const char *mangledName, FILE *fp) {
```
jclark, line 265: This needs to written so that it has well-defined behaviour when passed arbitrary function names.

runtime/error.c
```diff
@@ -181,3 +187,83 @@ TaggedPtr BAL_LANG_ERROR_NAME(message)(TaggedPtr error) {
...
+    // If mangled name starts from _B_, the identifier is non-public
+    if (mangledName[2] == '_') {
```
jclark: This code makes me sad. First, you must not assume that the first two chars are `_B`. More importantly, you must not assume that mangledName has length > 1.

runtime/error.c
```diff
@@ -181,3 +187,83 @@ TaggedPtr BAL_LANG_ERROR_NAME(message)(TaggedPtr error) {
...
+    // The org-name is not considered as a part of function name in backtrace.
+    int initialOffset = getOrgNameOffset(mangledName);
+
+    // Make sure parsing starts from the first character of
+    // nqual, local-name or mod-name
+    bool isFirstQualifier = true;
+    printModuleName(mangledName + initialOffset, &isFirstQualifier, fp);
+}
+
+static int getOrgNameOffset(const char *mangledName) {
```
jclark: This is all soooo ugly.

runtime/tests/demangle.c
manuranga: C strings do end in `\0` automatically, so why do we need another maker ?

runtime/error.c
manuranga: Correct camlCase encoding of that would be `getZQualifierOffset`.

runtime/error.c
KavinduZoysa: Changed

runtime/error.c

runtime/tests/demangle.c
KavinduZoysa: Yes, that is correct. When printing to an output file, I am adding a new line. That is why I am checking the new line. But removed the new line and checking the end of line in the test now.
Fixed

runtime/tests/demangle.c
```diff
@@ -0,0 +1,100 @@
+#include "../balrt.h"
+#include "../hash.h"
+#include <stdio.h>
+#include <assert.h>
+
+#define NAMES_COUNT 28
+
+HASH_DEFINE_KEY;
+
+const char input[NAMES_COUNT][60] = {
```
manuranga: I think it's better to have
```C
typedef struct {
    char subject[60];
    char expected[60];
} DemangleTestCase;

static const struct DemangleTestCase tests[] = {
    { "_Bb02ioprintln", "io:println" },
    ...
};
```

runtime/error.c
```diff
@@ -17,6 +17,13 @@
 #define EXCEEDS_MAX_PC_COUNT -3
 #define OUT_OF_MEMORY -4
 
+#define NON_PUBLIC_SYMBOL_PREFIX "_B_"
+#define NON_PUBLIC_SYMBOL_PREFIX_LENGTH strlen(NON_PUBLIC_SYMBOL_PREFIX)
```
manuranga: Hopefully clang optimize this to a constant. @KavinduZoysa please verify.

runtime/tests/demangle.c
KavinduZoysa: Changed

runtime/error.c
KavinduZoysa: Yes, it is optimized to a constant.

runtime/error.c
```diff
@@ -17,6 +17,13 @@
 #define EXCEEDS_MAX_PC_COUNT -3
 #define OUT_OF_MEMORY -4
 
+#define PUBLIC_BALLERINA_NAME 1
```
jclark: Use a sensible set of numbers. Those were just the numbers in my list of possible cases.

runtime/error.c
```diff
@@ -181,3 +193,113 @@ TaggedPtr BAL_LANG_ERROR_NAME(message)(TaggedPtr error) {
...
+    if (*name[0] == '_') {
+        *name = *name + 1;
+        nChars++;
+    }
+    *pp = *pp + nRead + nChars;
+    return true;
+}
+
+static void printUpto(const char *end, const char *start, FILE *fp) {
+    while (start < end) {
```
jclark: This can be a single call to fwrite.

runtime/error.c
```diff
@@ -181,3 +193,113 @@ TaggedPtr BAL_LANG_ERROR_NAME(message)(TaggedPtr error) {
...
+    int nRead;
+    if (sscanf(*pp, "%ld%n", &nChars, &nRead) <= 0) {
+        return false;
+    }
+    *name = *pp + nRead;
+    if (*name[0] == '_') {
+        *name = *name + 1;
+        nChars++;
+    }
+    *pp = *pp + nRead + nChars;
```
jclark: You need to check that pp + nRead does not go past the terminating '\0' of the string. There is also a potential integer overflow

runtime/error.c
```diff
@@ -181,3 +193,113 @@ TaggedPtr BAL_LANG_ERROR_NAME(message)(TaggedPtr error) {
...
+    return true;
+}
+
+static void printUpto(const char *end, const char *start, FILE *fp) {
```
jclark: First argument should be start; second argument should be a size_t length.  Call it printChars or something like that.

jclark (changes requested):

runtime/error.c
KavinduZoysa: Removed this method because this can be done a single line of code as suggested.

runtime/error.c

runtime/error.c
KavinduZoysa: Used `fwrite` instead of this whole method

runtime/error.c
KavinduZoysa: Changed.

runtime/error.c
```diff
@@ -181,3 +194,115 @@ TaggedPtr BAL_LANG_ERROR_NAME(message)(TaggedPtr error) {
...
+    }
+    if (nRead > MAX_DEMANGLING_DECIMAL_LENGTH) {
+        return false;
+    }
+    *name = *pp + nRead;
+    if (*name[0] == '_') {
+        *name = *name + 1;
+        nChars++;
+    }
+    for (int i = 0; i < nRead + nChars; i++) {
```
jclark: const char *p = name;

Then increment p and decrement nChars.

runtime/error.c
```diff
@@ -181,3 +194,115 @@ TaggedPtr BAL_LANG_ERROR_NAME(message)(TaggedPtr error) {
...
+    return demangleModules(mangledModules, fp);
+}
+
+static bool demangleCountedName(const char **pp, const char **name) {
+    long nChars;
+    int nRead;
+    if (sscanf(*pp, "%ld%n", &nChars, &nRead) <= 0) {
+        return false;
+    }
+    if (nRead > MAX_DEMANGLING_DECIMAL_LENGTH) {
```
jclark: Why do we need this?

runtime/error.c
```diff
@@ -181,3 +194,115 @@ TaggedPtr BAL_LANG_ERROR_NAME(message)(TaggedPtr error) {
...
+    }
+
+    bool ballerinaOrg = false; 
+    if (mangledName[0] == 'b') {
+        ballerinaOrg = true;
+        mangledName++;
+    }
+
+    const char *org;
+    if (mangledName[0] == '0') {
```
jclark, line 299: Don't need to special case this. You will just end up with an empty org (i.e. *org == '\0')

runtime/error.c
```diff
@@ -181,3 +194,115 @@ TaggedPtr BAL_LANG_ERROR_NAME(message)(TaggedPtr error) {
...
+            return _B_NON_BALLERINA_NAME;
+        }
+    }
+    if (fp != NULL) {
+        bool haveOrg = false;
+        if (ballerinaOrg) {
+            fputs("ballerina", fp);
+            haveOrg = true;
+        }
+        if (org != NULL) {
```
jclark, line 314: if (*org != '\0)

jclark (changes requested):

runtime/error.c
KavinduZoysa: We need this for a case like `_Bb02ioprintln`. If we pass without checking '0' to `demangleCountedName` function, it reads `02` together.

runtime/error.c
KavinduZoysa: I did this, because there was a potential integer overflow as mentioned above. Now we are not doing an addition like `nRead + nChars` we can remove this.

runtime/error.c

runtime/error.c
jclark: OK. Good point!

runtime/error.c
```diff
@@ -250,3 +263,116 @@ TaggedPtr BAL_LANG_ERROR_NAME(message)(TaggedPtr error) {
...
+    return demangleModules(mangledModules, fp);
+}
+
+static bool demangleCountedName(const char **pp, const char **name) {
+    long nChars;
+    int nRead;
+    if (sscanf(*pp, "%ld%n", &nChars, &nRead) <= 0) {
+        return false;
+    }
+    if (nRead > MAX_DEMANGLING_DECIMAL_LENGTH) {
```
jclark: You don't need this.

runtime/error.c
```diff
@@ -18,6 +18,15 @@
 #define EXCEEDS_MAX_PC_COUNT -3
 #define OUT_OF_MEMORY -4
 
+#define PUBLIC_BALLERINA_NAME 1
+#define NON_PUBLIC_BALLERINA_NAME 2
+#define NON_BALLERINA_NAME -1
+#define _B_NON_BALLERINA_NAME -2
+
+#define MAX_DEMANGLING_DECIMAL_LENGTH 4
```
jclark: Don't need this.

runtime/error.c
```diff
@@ -18,6 +18,15 @@
 #define EXCEEDS_MAX_PC_COUNT -3
 #define OUT_OF_MEMORY -4
 
+#define PUBLIC_BALLERINA_NAME 1
```
jclark: These numbers seem fairly random to me. Simpler just to make DemangleResult be an enum.

jclark (changes requested):

runtime/error.c
KavinduZoysa: Removed

runtime/error.c
KavinduZoysa: Removed

runtime/error.c

PR #469, opened 2021-09-19
Allow statements to use full type descriptors

### LINKED ISSUE #412: Allow statements to use full type descriptors

manuranga:
Allow statements to use full type descriptors
Fixes #412


jclark:
@manuranga Can we get rid of this?

PR #472, opened 2021-09-21
Support qualified identifiers

### LINKED ISSUE #473: Parse qualified identifiers in variable ref expr and type desc ref
Parsing part of #471 is for a variable-reference-expr and type-ref should allow a qualified-identifier.

This means extends the AST to allow to `VarRefExpr` and `TypeDescRef` to support identifiers with a prefix.

This has been separated out of #412.

manuranga:
Used to be part of #412
Fixes #473

### DIFF
--- a/compiler/modules/front.syntax/parseExpr.bal
+++ b/compiler/modules/front.syntax/parseExpr.bal
@@ -223,15 +223,13 @@ function startPrimaryExpr(Tokenizer tok) returns Expr|err:Syntax {
     Token? t = tok.current();
     if t is [IDENTIFIER, string] {
         Position pos = tok.currentPos();
-        string identifier = t[1];
-        VarRefExpr expr = { varName: identifier };
         check tok.advance();
-        t = tok.current();
-        if t == "(" {
+        var [prefix, varName] = check parseOptQualIdentifier(tok, t[1]);
+        if tok.current() == "(" {
             check tok.advance();
-            return finishFunctionCallExpr(tok, (), identifier, pos);
+            return finishFunctionCallExpr(tok, prefix, varName, pos);
         }
-        return expr;
+        return { prefix, varName };
     }
     else if t is [DECIMAL_NUMBER, string] {
         IntLiteralExpr expr = { base: 10, digits: t[1], pos: tok.currentPos() };
@@ -406,10 +404,10 @@ function parseSimpleConstExpr(Tokenizer tok) returns SimpleConstExpr|err:Syntax 
         return expr;
     }
     match t {
-        [IDENTIFIER, var varName] => {
-            VarRefExpr expr = { varName };
+        [IDENTIFIER, var identifier] => {
             check tok.advance();
-            return expr;
+            var [prefix, varName] = check parseOptQualIdentifier(tok, identifier);
+            return { prefix, varName };
         }
         [STRING_LITERAL, var value] => {
             ConstValueExpr expr = { value };
@@ -433,6 +431,16 @@ function parseSimpleConstExpr(Tokenizer tok) returns SimpleConstExpr|err:Syntax 
         }
     }
     return parseError(tok);
+}
+
+function parseOptQualIdentifier(Tokenizer tok, string identifier) returns [string?, string]|err:Syntax {
+    if tok.current() == ":" {
+        check tok.advance();
+        return [identifier, check tok.expectIdentifier()];
+    }
+    else {
+        return [(), identifier];
+    }
 }
 
 function parseNumericLiteralExpr(Tokenizer tok) returns NumericLiteralExpr|err:Syntax {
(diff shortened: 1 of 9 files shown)


compiler/modules/front/codeGen.bal
```diff
@@ -1025,8 +1025,9 @@ function codeGenExpr(CodeGenContext cx, bir:BasicBlock bb, Environment env, s:Ex
             return codeGenTypeTest(cx, bb, env, td, left, negated);
         }
         // Variable reference
-        var { varName } => {
-            return codeGenVarRefExpr(cx, varName, env, bb);
+        var { varName, prefix } => {
```
manuranga: @jclark This doesn't match after adding prefix, also getting below at compile time
`WARNING [codeGen.bal:(1028:9,1028:32)] pattern will not be matched`
Is this a bug or am I missing something ?

jclark: `match` seems to have lots of bugs

manuranga: ` var ref if ref is s:VarRefExpr => {` works as a workaround.

compiler/modules/front.syntax/parseStmt.bal
```diff
@@ -76,28 +76,29 @@ function parseStmt(Tokenizer tok) returns Stmt|err:Syntax {
 }
 
 
-function finishIdentifierStmt(Tokenizer tok, string identifier, Position pos) returns Stmt|err:Syntax {
+function finishIdentifierStmt(Tokenizer tok, string identifier, Position pos, string? prefix = ()) returns Stmt|err:Syntax {
```
jclark: The subset grammar won't allow most of these cases because they don't make sense semantically in the subset, because we don't have module-level variables and constants are not structured. The handling for `:` also makes sense only where there's not already a prefix. I think a better approach is to have two separate functions and then factor out the logic for common cases. We should really have the grammar first and then write the code to match the grammar.

jclark:
I updated the grammar. I believe the new cases that can start with a qualified identifier are only method-call-expr and local-var-decl-stmt

manuranga:
I think we'll need `const-reference-expr` to be a  qualified identifier as well. I'll add it.

jclark:
BTW, for this issue please leave the handling of int:* in parsePrimaryTypeDesc as is. I will fix that separately.

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -199,9 +199,17 @@ function parsePrimaryTypeDesc(Tokenizer tok) returns TypeDesc|err:Syntax {
...
             check tok.advance();
+            TypeDescRef r;
+            if tok.current() == ":" {
+                check tok.advance();
+                r = { prefix: identifier, ref: check tok.expectIdentifier(), pos };
+            }
+            else {
+                r = { ref: identifier, pos };
+            }
             return r;
```
manuranga: I wasn't sure I should just `return` inline.

compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -224,14 +224,23 @@ function startPrimaryExpr(Tokenizer tok) returns Expr|err:Syntax {
     if t is [IDENTIFIER, string] {
         Position pos = tok.currentPos();
         string identifier = t[1];
-        VarRefExpr expr = { varName: identifier };
         check tok.advance();
-        t = tok.current();
-        if t == "(" {
+        string varName;
+        string? prefix;
+        if tok.current() == ":" {
```
jclark: Something similar to this bit of code is repeated in several places. Try factoring out a function:

`function parseOptQualIdentifier(Tokenizer tok, string identifier) returns [string?, string]|err:Syntax;`

Then this code would become

`var [prefix, varName] = check parseOptQualIdentifier(tok, t[1]);`

compiler/modules/front.syntax/parseStmt.bal
```diff
@@ -121,16 +111,27 @@ function finishIdentifierStmt(Tokenizer tok, string identifier, Position pos) re
...
+        return finishQualIdentifierStmt(tok, identifier, check tok.expectIdentifier(), pos);
     }
     else if cur is [IDENTIFIER, string] {
         TypeDescRef ref = { ref: identifier, pos };
         return finishVarDeclStmt(tok, ref);
     }
+    return finishQualIdentifierStmt(tok, (), identifier, pos);
+}
+
+function finishQualIdentifierStmt(Tokenizer tok, string? prefix, string identifier, Position pos) returns Stmt|err:Syntax {
```
jclark: This is a neat solution. But I would call it finishOptQualIdentifierStmt since it is called for identifiers that are not qualified.

jclark (changes requested):

PR #475, opened 2021-09-22
Enhance testSuite driver for error tests to support multiple modules

### LINKED ISSUE #470: Enhance testSuite driver for error tests to support multiple modules
Specifically:

- allow @error in submodules
- check that the filename in error messages is present and correct (test should fail if filename is missing)

Part of #438

heshanpadmasiri:
## Purpose

- Enhance testSuite driver for error tests to support multiple modules
- Add additional tests

Resolves #470

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/compile.bal
```diff
@@ -72,7 +72,13 @@ function compileBalFile(string filename, string basename, string? outputBasename
 
 function processModule(CompileContext cx, bir:ModuleId id, front:SourcePart[] sourceParts, string? outFilename) returns front:ResolvedModule|CompileError {
     front:ScannedModule scanned = check front:scanModule(sourceParts, id);
-    ResolvedImport[] resolvedImports = from var mod in scanned.getImports() select check resolveImport(cx, mod);
+    // Fallowing doesn't properly pass the error back to the calling function if we get an error the import
+    // ResolvedImport[] resolvedImports = from var mod in scanned.getImports() select check resolveImport(cx, mod);
+    ResolvedImport[] resolvedImports = [];
```
jclark, line 77: The new code looks equivalent to me. When you say `doesn't properly pass back the error` do you mean the code is wrong or that there's a JBUG?

compiler/tests/testSuite.bal
```diff
@@ -27,6 +28,12 @@ function testCompileVPO(string path, string kind) returns io:Error? {
     }
 }
 
+function verifyErrorFileName(err:Any err) {
```
jclark: Filename not FileName

compiler/tests/testSuite.bal
```diff
@@ -27,6 +28,12 @@ function testCompileVPO(string path, string kind) returns io:Error? {
...
+    err:Detail detail = <err:Detail>err.detail();
+    test:assertTrue(detail.location is err:Location, "error without location");
+    test:assertNotEquals((<err:Location>detail.location).filename.length(), 0, "error with an empty filename");
```
jclark: We need to check that the filename is right, not just that it is non-empty.

compiler/tests/testSuite.bal
```diff
@@ -69,22 +78,29 @@ function compileErrorLineNumber(CompileError err) returns int? {
...
+    foreach var dir in check file:readDir(path) {
         if !check file:test(dir.absPath, file:IS_DIR) {
             continue;
         }
         string category = check file:basename(dir.absPath);
         foreach var test in check file:readDir(dir.absPath) {
             string name = check file:basename(test.absPath);
             var [base, ext] = basenameExtension(name);
-            if ext != ".bal" {
+            if check file:test(test.absPath, file:IS_DIR) && ext == ".modules" {
```
jclark: You can just look at test.dir instead of calling file:text

But I don't understand why you are looking at .modules in this function

compiler/tests/testSuite.bal
```diff
@@ -37,6 +44,8 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
             test:assertNotExactEquals(err, (), "expected an error " + path);
         }
         else {
+            // JBUG #31334 cast needed
+            verifyErrorFileName(err);
```
jclark: You need to look at the .modules directory here and find the file and line number of the @error line

jclark (changes requested):

compiler/compile.bal
heshanpadmasiri: What I meant was check doesn't seems to return the error so that I am not getting the errors in imports back to `testCompileEU`

compiler/compile.bal
```diff
@@ -72,7 +72,13 @@ function compileBalFile(string filename, string basename, string? outputBasename
...
+    foreach var mod in scanned.getImports() {
+        ResolvedImport im = check resolveImport(cx, mod);
```
jclark: Please don't abbreviate `import` to `im`. If the type is ResolvedImport then use `ri`.

compiler/compile.bal
jclark: That should have a JBUG comment then.  But I'm surprised because in other cases of this kind of jBallerina bug, I found the generated code crashed. When resolveImport is returning an error, what is happening?

jclark:
Let's start the name of tests relating to imported functions with `function`. I did this for existing tests here af25c3b7e39bd1682f876ad1a5b9e1025cbc1481

compiler/compile.bal
heshanpadmasiri: If we have it in this way when we hit something like an semantic error since we have checks in all the function calls it get passed all the way back to `testCompileEU` where we handle the error to make sure it is what we are expecting. But if we have it in the previous way it get test failure saying 
```
 '+' not supported for operand types {"location":{"filename":"bar-e.bal","startPos":null,"endPos":null},"functionName":"b"}
at wso2.nballerina.err.0_1_0:semantic(err.bal:61)
```
and stack trace ends at 
```
wso2.nballerina.0_1_0.tests.testSuite:testCompileEU(tests/testSuite.bal:42)
```
(that line was  `CompileError? err = testCompileFile(path);`)

But I am not certain it's a JBUG because so far I when I tried to create an isolated test case it has worked correctly

compiler/tests/testSuite.bal
heshanpadmasiri: fixed

compiler/tests/testSuite.bal
heshanpadmasiri: fixed

compiler/tests/testSuite.bal
heshanpadmasiri: Changed such that it'll check if there is a file in .modules directory matching the error's filename and optimistically accept that is to be the file that caused the error. This is because if A imports B we will get the error in A as well with B's name (but not path)

compiler/compile.bal
heshanpadmasiri: fixed

compiler/tests/testSuite.bal
heshanpadmasiri: What I am trying to do here is to see if the "test" is actually the modules directory for a test and if so call `listSources` recursively on that directory as well.

compiler/compile.bal
manuranga: Hmm, I tried to see if there is a difference in behavior, I didn't come across any. This is what I tried https://gist.github.com/manuranga/7ff76b633dd38c0e23b5fb61c5701c96 

Please see if you can modify above to give different behaviors under `iter` vs `query`.

compiler/tests/testSuite.bal
```diff
@@ -57,6 +76,26 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
     }
 }
 
+function importFilename(string path, string errFilename) returns string? {
+    int? index = path.lastIndexOf(".bal");
+    if index is int {
+        string subModPath = path.substring(0, index) + ".modules";
+        if checkpanic file:test(subModPath, file:EXISTS) && checkpanic file:test(subModPath, file:IS_DIR) {
```
jclark: All these checkpanics aren't very nice.  Can we make importFilename return string|file:Error? Then call checkpanic once in verifyErrorFilename.

compiler/tests/testSuite.bal
```diff
@@ -27,6 +28,22 @@ function testCompileVPO(string path, string kind) returns io:Error? {
     }
 }
 
+function verifyErrorFilename(err:Any err, string path) returns string {
```
jclark: I would make err:Detail the argument here.

compiler/tests/testSuite.bal
```diff
@@ -17,6 +17,7 @@ function testCompileVPO(string path, string kind) returns io:Error? {
     else {
         string msg = "compilation error";
         if err is err:Any {
+            _ = verifyErrorFilename(err, path);
```
jclark: Why are you calling verifyErrorFilename here? This is just goint to get a failed assert in a few lines.

compiler/tests/testSuite.bal
```diff
@@ -57,6 +76,26 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
     }
 }
 
+function importFilename(string path, string errFilename) returns string? {
+    int? index = path.lastIndexOf(".bal");
+    if index is int {
+        string subModPath = path.substring(0, index) + ".modules";
+        if checkpanic file:test(subModPath, file:EXISTS) && checkpanic file:test(subModPath, file:IS_DIR) {
+            foreach var subMod in checkpanic file:readDir(subModPath) {
+                if !checkpanic file:test(subMod.absPath, file:IS_DIR) {
```
jclark: Just use subMod.dir

compiler/tests/testSuite.bal
```diff
@@ -37,6 +54,8 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
             test:assertNotExactEquals(err, (), "expected an error " + path);
         }
         else {
+            // JBUG #31334 cast needed
+            string errorFilepath = verifyErrorFilename(err, path);
```
jclark: Why isn't this being called within the `if` just before it is used in the line 70?

compiler/tests/testSuite.bal
```diff
@@ -69,22 +108,29 @@ function compileErrorLineNumber(CompileError err) returns int? {
...
         if !check file:test(dir.absPath, file:IS_DIR) {
             continue;
         }
         string category = check file:basename(dir.absPath);
         foreach var test in check file:readDir(dir.absPath) {
             string name = check file:basename(test.absPath);
             var [base, ext] = basenameExtension(name);
-            if ext != ".bal" {
+            if check file:test(test.absPath, file:IS_DIR) && ext == ".modules" {
+                TestSuiteCases subCases = check listSources(initialChars, test.absPath);
```
jclark: Why do you think you need to listSources in *.module?

compiler/tests/testSuite.bal
jclark: Why?

jclark (changes requested):

compiler/tests/testSuite.bal
jclark: > This is because if A imports B we will get the error in A as well with B's name (but not path)

Can you explain more? That seems strange. Which test case gives you exactly what error?

compiler/tests/testSuite.bal
manuranga: Also move this function below `testCompileEU`

compiler/tests/testSuite.bal
jclark: There's a bug with how we are using s:defnLocation. It gives us the line number of definition, but we are using it where we only want the filename. Maybe this is related.

compiler/tests/testSuite.bal
```diff
@@ -37,6 +54,8 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
             test:assertNotExactEquals(err, (), "expected an error " + path);
         }
         else {
+            // JBUG #31334 cast needed
```
manuranga: I don't see a cast.

compiler/tests/testSuite.bal
```diff
@@ -69,22 +108,29 @@ function compileErrorLineNumber(CompileError err) returns int? {
...
 }
 
-function listSourcesVPO() returns TestSuiteCases|error => listSources("vpo");
+function listSourcesVPO() returns TestSuiteCases|error => listSources("vpo", "./testSuite");
 
-function listSourcesEU() returns TestSuiteCases|error => listSources("eu");
+function listSourcesEU() returns TestSuiteCases|error => listSources("eu", "./testSuite");
 
-function listSources(string initialChars) returns TestSuiteCases|io:Error|file:Error {
+function listSources(string initialChars, string path) returns TestSuiteCases|io:Error|file:Error {
```
manuranga: This might be already apparent with below comments by James; you only have to list the root bal files from this function, since the import lookup will discover the files inside modules. So you will not need the path to be a parameter.

compiler/tests/testSuite.bal
```diff
@@ -57,6 +76,26 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
     }
 }
 
+function importFilename(string path, string errFilename) returns string? {
+    int? index = path.lastIndexOf(".bal");
```
manuranga: use `path.endsWith(".bal")` and return if it doesn't match, this will avoid an extra level of indentation.

compiler/tests/testSuite.bal
```diff
@@ -57,6 +76,26 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
...
+                }
+                foreach var test in checkpanic file:readDir(subMod.absPath) {
+                    if  checkpanic file:basename(test.absPath) == errFilename {
+                        return test.absPath;
+                    }
+                }
+            }
+        }
+    }
+    return ();
```
manuranga: return nil is implicit, no need to type it.

jclark:
The logic I would expect would be to read all the modules to find the file that contains the '@error' line and then check that the filename in the error message was the same.

compiler/tests/testSuite.bal
heshanpadmasiri: What I meant was If lets say `foo.bal` import `bar.bal` and `bar.bal` has an error, error's filename is just `bar.bal` not the path to `bar.bal` (unlike if `foo.bal` had the error in which case you get a path not just the filename). So then what I am doing is going over all the files in foo.bal's modules to see if I can find a `bar.bal` (this is because I don't know to what sub module `bar.bal` belongs to from the error). A problem will arise if we have multiple `bar.bal` files because we can't determine which `bar.bal` with the current error information.

jclark: > error's filename is just bar.bal 

That doesn't sound right.  Which test case in your pull request is this?

compiler/tests/testSuite.bal
```diff
@@ -57,6 +58,40 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
...
+        test:assertEquals(filename, checkpanic file:basename(importName), "invalid error filename");
+        return importName;
+    }
+    else {
+        test:assertEquals(filename, path, "invalid error filename");
+        return path;
+    }
+}
+
+function importFilename(string path, string errFilename) returns string|file:Error? {
```
manuranga: Current name doesn't give a good idea about what it is doing.
How about something like:
```suggestion
function searchSubModuleFile(string rootModule, string filename) returns string|file:Error? {
```

I personally think it's better if the error trace had the sub-module name, then we wouldn't need this function. Let's discuss that option separately.

compiler/tests/testSuite.bal
heshanpadmasiri: I thought we need to test we need to test both the imported files separately and then test them after importing. Now removed

compiler/tests/testSuite.bal
heshanpadmasiri: fixed by making imported files return path relative to modules directory

compiler/tests/testSuite.bal
heshanpadmasiri: function removed

compiler/tests/testSuite.bal
heshanpadmasiri: function removed

compiler/tests/testSuite.bal
heshanpadmasiri: I thought we need to verify error has a filename here as well even though we are not suppose to get an error here. Removed

compiler/tests/testSuite.bal
heshanpadmasiri: function removed

compiler/tests/testSuite.bal
heshanpadmasiri: fixed

compiler/tests/testSuite.bal
heshanpadmasiri: I thought we need to add them to validate whenever we get a test failure it was cause by import not an issue in the imported file itself (because it was tested separately) . Removed

compiler/tests/testSuite.bal
heshanpadmasiri: comment removed

compiler/tests/testSuite.bal
heshanpadmasiri: fixed

compiler/tests/testSuite.bal
heshanpadmasiri: function removed

compiler/tests/testSuite.bal
heshanpadmasiri: fixed

compiler/tests/testSuite.bal
heshanpadmasiri: function removed. All errors should show the path (either absolute or relative) of the file that caused the error

compiler/compile.bal
```diff
@@ -135,6 +141,11 @@ function subModuleSourceParts(string basename, bir:ModuleId id) returns front:So
     return sourceParts;
 }
 
+function getPathRelativeToModuleDir(string modulesDirPath, string filePath) returns string|file:Error {
```
jclark: We want the filename in the same form as the filename for the root. So I think we just do `file:joinPath(directory, file:basename(md,absPath))`

compiler/tests/testSuite.bal
```diff
@@ -57,6 +57,66 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
...
 }
 
+function moduleDir(string filePath) returns string|file:Error? {
+    string subModPath = filePath.substring(0, filePath.length()-4) + ".modules";
+    if check file:test(subModPath, file:IS_DIR) {
+        return check file:normalizePath(subModPath, file:CLEAN);
+    }
+}
+
+function verifyErrorLineNumber(CompileError err, string path) returns file:Error? {
```
jclark: This is all so complicated. Write a function

```
function expectedErrorLocation(string testCasePath) returns [string, int]?
```

Then have do asserts that the filename in the error matches the expected filename (probably trim some directories off first), and if there's a line number in the error, that it matches the expected line number.

jclark (changes requested):

compiler/tests/testSuite.bal
heshanpadmasiri: fixed

compiler/compile.bal
heshanpadmasiri: fixed. Add normalize to get rid of the `/./` joinPath adds

compiler/tests/testSuite.bal
```diff
@@ -57,15 +68,41 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
...
+            }
+        }
+    }
+    [errorLineNumber, errorFileName] = findErrorLine(path, errorLineNumber);
+    if errorFileName is string && errorLineNumber is int {
+        return [errorLineNumber, errorFileName];
+    }
+}
+
+function findErrorLine(string filePath, int? currentErrorLine) returns [int?, string?] {
```
jclark: Make the type be `[string, int]|io:Error?`

compiler/tests/testSuite.bal
```diff
@@ -57,15 +68,41 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
...
+    int? fileErrorLine = checkpanic errorLine(filePath);
```
jclark: Use `check`.

compiler/tests/testSuite.bal
```diff
@@ -57,15 +68,41 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
...
+        return check file:normalizePath(subModPath, file:CLEAN);
+    }
+}
+
+function expectedErrorLocation(CompileError err, string path) returns [int, string]|file:Error|io:Error? {
+    string? modulePath = check moduleDir(path);
+    string? errorFileName = ();
+    int? errorLineNumber = ();
+    if modulePath is string {
+        foreach var dir in check file:readDir(modulePath) {
```
jclark: dir has type MetaData, so use `md` not dir.

compiler/tests/testSuite.bal
```diff
@@ -57,15 +68,41 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
...
+            if !check file:test(dir.absPath, file:IS_DIR) {
```
jclark: MetaData has a `dir` field. Use that not file:test.

compiler/tests/testSuite.bal
```diff
@@ -57,15 +68,41 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
...
+function moduleDir(string filePath) returns string|file:Error? {
+    string subModPath = filePath.substring(0, filePath.length()-4) + ".modules";
+    if check file:test(subModPath, file:IS_DIR) {
+        return check file:normalizePath(subModPath, file:CLEAN);
+    }
+}
+
+function expectedErrorLocation(CompileError err, string path) returns [int, string]|file:Error|io:Error? {
+    string? modulePath = check moduleDir(path);
+    string? errorFileName = ();
```
jclark: [string, int]? errorLocation = ();

compiler/tests/testSuite.bal
```diff
@@ -57,15 +68,41 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
...
-    if err is io:Error {
-        return ();
+function moduleDir(string filePath) returns string|file:Error? {
+    string subModPath = filePath.substring(0, filePath.length()-4) + ".modules";
+    if check file:test(subModPath, file:IS_DIR) {
+        return check file:normalizePath(subModPath, file:CLEAN);
+    }
+}
+
+function expectedErrorLocation(CompileError err, string path) returns [int, string]|file:Error|io:Error? {
```
jclark: [string, int] is more natural than [int, string] here

compiler/tests/testSuite.bal
```diff
@@ -57,15 +68,41 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
...
+    string? modulePath = check moduleDir(path);
+    string? errorFileName = ();
+    int? errorLineNumber = ();
+    if modulePath is string {
+        foreach var dir in check file:readDir(modulePath) {
+            if !check file:test(dir.absPath, file:IS_DIR) {
+                continue;
+            }
+            foreach var file in check file:readDir(dir.absPath) {
+                [errorLineNumber, errorFileName] = findErrorLine(file.absPath, errorLineNumber);
```
jclark: This isn't detecting duplicates.

compiler/tests/testSuite.bal
```diff
@@ -57,15 +68,41 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
     }
 }
 
-function compileErrorLineNumber(CompileError err) returns int? {
-    if err is io:Error {
-        return ();
+function moduleDir(string filePath) returns string|file:Error? {
```
jclark: Move this below expectedErrorLocation

jclark (changes requested):

compiler/compile.bal
manuranga: I couldn't fine a reported issue in jBallerina. I tried to replicate the issue by returning a fake error in `resolveImport`, but an error got return as expected.
So still unclear on how to replicate, or even if there is an issue in the first place.

@gimantha can you please have a call with @heshanpadmasiri and figure out if something wrong is going on here, and create an issue if there is.

gimantha: Sure will have a look

compiler/tests/testSuite.bal
heshanpadmasiri: fixed

compiler/tests/testSuite.bal
heshanpadmasiri: fixed

compiler/tests/testSuite.bal
heshanpadmasiri: fixed

compiler/tests/testSuite.bal
heshanpadmasiri: fixed

compiler/tests/testSuite.bal
heshanpadmasiri: `findErrorLine` will check if the given `currentErrorLocation` to be null if it find an error line (this is suppose to detect if multiple files have errors). `errorLine` will check if it already have found an error line every time it come across an error line (this is suppose to detect if there are multiple errors in the same file)

compiler/tests/testSuite.bal
heshanpadmasiri: fixed

compiler/tests/testSuite.bal
heshanpadmasiri: fixed

compiler/tests/testSuite.bal
heshanpadmasiri: fixed

compiler/tests/testSuite.bal
```diff
@@ -46,9 +46,20 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
             else if !err.message().includes("'io:println'") {
                 test:assertFalse(err is err:Semantic, "semantic error on U test" + path);
             }
-            int? lineNumber = compileErrorLineNumber(err);
-            if lineNumber != () && (isE || kind[1] == "e") {
-                test:assertEquals(lineNumber, check errorLine(path), "wrong line number in error " + path);
+            if kind == "e" || kind == "ue" {
+                var expectedErrorLocation = checkpanic expectedErrorLocation(err, path);
+                if !(expectedErrorLocation is ()) {
+                    var [expectedFilename, expectedLineNo] = expectedErrorLocation;
```
manuranga: We could avoid this `if` and extra variable, by `expectedErrorLocation` returning and error instead `()` when it can't find a location. Something like `return error("@error annotation not found");` should work. 

Do we need a `checkpanic` ? wouldn't just `check` work?

heshanpadmasiri: There are test cases without @error annotaions such as `testSuite/05-narrow/13-e.bal` is this expected or is this an error?

compiler/tests/testSuite.bal
```diff
@@ -57,15 +68,34 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
...
+    string subModPath = filePath.substring(0, filePath.length()-4) + ".modules";
+    if check file:test(subModPath, file:IS_DIR) {
+        return check file:normalizePath(subModPath, file:CLEAN);
+    }
+}
+
+function findErrorLine(string filePath, any? currentErrorLocation) returns [string, int]|io:Error? {
+    int? fileErrorLine = check errorLine(filePath);
+    if fileErrorLine is int {
+        test:assertTrue(currentErrorLocation is (), "Multiple files with error annotations found");
```
jclark: Don't capitalize error messages

compiler/tests/testSuite.bal
```diff
@@ -57,15 +68,34 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
...
+}
+
+function moduleDir(string filePath) returns string|file:Error? {
+    string subModPath = filePath.substring(0, filePath.length()-4) + ".modules";
+    if check file:test(subModPath, file:IS_DIR) {
+        return check file:normalizePath(subModPath, file:CLEAN);
+    }
+}
+
+function findErrorLine(string filePath, any? currentErrorLocation) returns [string, int]|io:Error? {
```
jclark: `any` is not the right type. I suggest you add a type definition 

type FilenameLine [string, int];

compiler/tests/testSuite.bal
manuranga: It's a bug, it should have `@error` on line 6.

compiler/tests/testSuite.bal
```diff
@@ -46,9 +46,20 @@ function testCompileEU(string path, string kind) returns file:Error|io:Error? {
...
-                test:assertEquals(lineNumber, check errorLine(path), "wrong line number in error " + path);
+            if kind == "e" || kind == "ue" {
+                var expectedErrorLocation = checkpanic expectedErrorLocation(err, path);
+                if !(expectedErrorLocation is ()) {
+                    var [expectedFilename, expectedLineNo] = expectedErrorLocation;
+                    // JBUG #31334 cast needed
+                    err:Detail detail = <err:Detail> err.detail();
+                    test:assertTrue(detail.location is err:Location, "error without location");
+                    string filename =(<err:Location>detail.location).filename;
+                    test:assertEquals(file:getAbsolutePath(filename), expectedFilename, "Invalid error filename" + filename);
```
jclark: Don't capitalize error message

jclark (changes requested):
Nearly there.

compiler/tests/testSuite.bal
manuranga: How was that passing ? Let me check.

compiler/tests/testSuite.bal
heshanpadmasiri: fixed

compiler/tests/testSuite.bal
heshanpadmasiri: fixed

compiler/tests/testSuite.bal
heshanpadmasiri: fixed

compiler/tests/testSuite.bal
manuranga: Looks like the previous logic didn't to the line check if `@error` wasn't there, but please feel free to add it to any file with `-e.bal` suffix.

heshanpadmasiri: made #484 to fix this

compiler/compile.bal
manuranga: @gimantha did we create a issue for this in ballerina-lang repo? If so what is the number?

PR #476, opened 2021-09-22
Add `expectIdentifier` and refactor

### LINKED ISSUE #474: Add an expect method for `VariableLengthToken`s to Tokenizer
Currently there is an `expect` method for fixed tokens. Adding similar method for `VariableLengthToken` will

- Simplify usage code by removing one level of indentation
- Improve the parser error messages

manuranga:
Fixes #474

### DIFF
--- a/compiler/modules/front.syntax/token.bal
+++ b/compiler/modules/front.syntax/token.bal
@@ -268,6 +268,24 @@ class Tokenizer {
         return true;
     }
 
+    function expectIdentifier() returns string|err:Syntax {
+        Token? t = self.curTok;
+        if t is [IDENTIFIER, string] {
+            check self.advance();
+            return t[1];
+        }
+        else {
+            err:Template msg;
+            if t is string {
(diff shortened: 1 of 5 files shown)


compiler/modules/front.syntax/token.bal
```diff
@@ -268,6 +268,24 @@ class Tokenizer {
...
+    function expectIdentifier() returns string|err:Syntax {
+        Token? t = self.curTok;
+        if t is [IDENTIFIER, string] {
+            check self.advance();
+            return t[1];
+        }
+        else {
+            err:Template msg;
+            if t is string {
+                msg = `expected identifier; got ${t}`;
```
jclark: expected an identifer

jclark (changes requested):
This is a very nice improvement!

compiler/modules/front.syntax/token.bal
manuranga: Fixed

PR #488, opened 2021-09-26
Fix #487

### LINKED ISSUE #487: Empty source files give `IndexOutOfRange` error
Compiling an empty source file gives 
```
error: {ballerina/lang.array}IndexOutOfRange {"message":"array index out of range: index: -1, size: 0"}
        at wso2.nballerina.front.syntax.0_1_0.Tokenizer:restore(token.bal:336)
```

manuranga:
Fix passing arse empty bal files.
Fixes #487

### DIFF
--- a/compiler/modules/front.syntax/token.bal
+++ b/compiler/modules/front.syntax/token.bal
@@ -330,6 +330,12 @@ class Tokenizer {
         self.mode = s.mode;
         self.curTok = s.curTok;
 
+        if self.lineIndex == 0 {
+            // advance() hasn't been called or lines are empty
+            self.fragCodes = [];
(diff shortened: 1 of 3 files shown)


compiler/modules/front.syntax/token.bal
```diff
@@ -330,6 +330,10 @@ class Tokenizer {
         self.mode = s.mode;
         self.curTok = s.curTok;
 
+        if self.lineIndex == 0 {
```
jclark, line 333: In this case I would set fragCodes and fragments to [], so it's the same as at initialization.

jclark (changes requested):

PR #496, opened 2021-09-27
Support for multiple debug levels from command line

### LINKED ISSUE #486: Replace `--debug` option with `--debugLevel=n` option
With #6, we should by default produce enough debugging info for backtraces. This is corresponds to nback:Options debugLevel of 1.  So we should make the default debugLevel be 1, and then have an option that can change it to 0 (no debugging) or 2 (full debugging, which is not yet implemented).  Easiest solution is a `--debugLevel` option that allows an argument of 0/1/2.

KavinduZoysa:
## Purpose
Fixes #486

### DIFF
--- a/compiler/main.bal
+++ b/compiler/main.bal
@@ -12,7 +12,7 @@ type CompileError err:Any|io:Error|file:Error;
 public type Options record {
     boolean testJsonTypes = false;
     boolean showTypes = false;
-    int debugLevel = nback:DEBUG_BACKTRACE;
+    int? debugLevel;
     // outDir also implies treating each file as a separate module
     string? outDir = ();
     string? expectOutDir = ();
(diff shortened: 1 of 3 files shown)


KavinduZoysa:
@jclark, regarding the `debugLevel=0` case, when panic happens it will print the backtrace without line number and filename because debuginfo is not available. Is that expected behaviour?

jclark:
Yes. That's why this is not the default.

compiler/main.bal
```diff
@@ -43,9 +43,10 @@ public function main(string[] filenames, *Options opts) returns error? {
         check showTypes([{ filename: filenames[0] }]);
         return;
     }
+    int debugLevel = opts.debugLevel;
     nback:Options nbackOptions = {
         gcName: check nback:validGcName(opts.gc),
-        debugLevel: opts.debug ? nback:DEBUG_BACKTRACE : nback:DEBUG_NONE
+        debugLevel: debugLevel < nback:DEBUG_NONE || debugLevel > nback:DEBUG_FULL ? nback:DEBUG_BACKTRACE : debugLevel
```
jclark: Write a function nback:validDebugLevel like nback:validGcName.

jclark (changes requested):

compiler/main.bal
manuranga: AFAIU, if it's valid, this function should return debugLevel as an `(1|2|3)`, instead returning a boolean that says it's valid.

This is what I try to capture as the [parse, don't validate](https://docs.google.com/document/d/1Vj_UErPC1DaNnHY92nmq6TlwA5WwmZYkr4xPqP8Lur0/edit#heading=h.oxug7334y2gf) rule in style guide.

manuranga: I would also assume any other input will result in an error.

KavinduZoysa: I have created this method to return DEBUG_BACKTRACE for any integer value which does not equal 0 or 2 (Rust also following a similar thing). If the input is 0 or 2, it will return the debugLevel accordingly.

PR #516, opened 2021-09-29
Introduce string:Char type

### LINKED ISSUE #477: Implement `string:Char` type
**Description:**
For the moment, we should be able to parse `string:Char` as we have done for `int` subtypes.
Need to figure out how to represent this within `StringSubtype`.

On top of my mind we should be able to do:
```ballerina
type StringSubtype StringSubtypeData|STRING_CHAR;

const string STRING_CHAR = "char";

type StringSubtypeData readonly & record {|
    boolean allowed;
    string[] values;
|};
```

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**Steps to reproduce:**

**Related Issues:**

### LINKED ISSUE #493: Redo front-end handling of int subtypes
At the moment, these are special-cased in front.syntax:parseTypeDesc.

We should change this as follows:

* parseTypeDesc should turn them into a TypeDescRef with a prefix of `int`
* lib.bal should have a map that maps each of the names onto a SemType (subtype of ModuleExports)
* the prefix should be mapped onto this map

Follow on from #438 and #471.

rdhananjaya:
## Purpose
Fixes https://github.com/ballerina-platform/nballerina/issues/477
Fixes https://github.com/ballerina-platform/nballerina/issues/493

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -166,6 +165,23 @@ function parsePrimaryTypeDesc(Tokenizer tok) returns TypeDesc|err:Syntax {
...
+            check tok.advance();
+            Token? t = tok.current();
+            if t is [IDENTIFIER, string] {
+                var name = t[1];
+                if name == "Char" {
+                    check tok.advance();
+                    return "char";
+                }
+                return tok.err("unrecognized string subtype '" + name + "'");
+            }
```
manuranga:
```suggestion
            if tok.expectIdentifier() == "Char" {
                return "char";
            }
            else {
                return tok.err(`unrecognized string subtype ${name}`);
            }
```

compiler/modules/types/string.bal
```diff
@@ -1,65 +1,152 @@
...
+public function stringChar() returns SemType {
+    StringSubtype st = {
+        char: { allowed: false, values: [] }, 
+        nonChar: { allowed: true, values: [] }
+    };
     return uniformSubtype(UT_STRING, st);
 }
 
+function isAllChar(StringSubtype st) returns boolean {
+    return !st.char.allowed && st.char.values.length() == 0;
```
manuranga: There is a very rare case where one can list all the chars one by one, is that a possibility?

compiler/modules/types/string.bal
```diff
@@ -1,65 +1,152 @@
...
+}
+
 function stringSubtypeSingleValue(SubtypeData d) returns string? {
     if d is boolean {
         return ();
     }
-    StringSubtype s = <StringSubtype>d;
-    if !s.allowed {
+    var st =  <StringSubtype>d;
```
manuranga:
```suggestion
    StringSubtype st =  <StringSubtype>d;
```

manuranga: I'll add a style rule.

compiler/tests/data/string-singleton.bal
```diff
@@ -14,8 +14,8 @@
 // Z<:XYZ
 // Z<:XZ
 type X "x";
-type Y "y";
-type Z "z";
+type Y "y-y";
+type Z "z-z";
```
manuranga, line 18: Need a test with all four of string singleton, char singleton, string type, char type, but I would add a new one and leave this one as is.

compiler/modules/types/string.bal
rdhananjaya: I think we can remove the call to this function.

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -166,6 +165,23 @@ function parsePrimaryTypeDesc(Tokenizer tok) returns TypeDesc|err:Syntax {
             // JBUG should not need cast #30191
             return <LeafTypeDesc>cur;
         }
+        "string" => {
```
jclark: This is not the right way to do it now we have multiple modules. See https://github.com/ballerina-platform/nballerina/issues/493

compiler/modules/types/string.bal
```diff
@@ -1,65 +1,152 @@
...
+
+public function stringChar() returns SemType {
+    StringSubtype st = {
+        char: { allowed: false, values: [] }, 
+        nonChar: { allowed: true, values: [] }
+    };
     return uniformSubtype(UT_STRING, st);
 }
 
+function isAllChar(StringSubtype st) returns boolean {
```
jclark: This should be called something like stringSubtypeIsAllChar

compiler/modules/types/string.bal
```diff
@@ -1,65 +1,152 @@
 // Implementation specific to basic type string.
 
 type StringSubtype readonly & record {|
+    CharStringSubtype char;
+    NonCharStringSubtype nonChar;
+|};
+
+// allowed = false with empty `values` means `String:char`
```
jclark, line 8: I know I used `boolean allowed;`, but coming back to the code I am finding it pretty confusing. Maybe `boolean valuesInOrOut;` then `const VALUES_IN = true; const VALUES_OUT = false;` ?

compiler/modules/types/string.bal
```diff
@@ -1,65 +1,152 @@
...
+        nonChar: { allowed: true, values: [] }
+    };
     return uniformSubtype(UT_STRING, st);
 }
 
+function isAllChar(StringSubtype st) returns boolean {
+    return !st.char.allowed && st.char.values.length() == 0;
+}
+
 function stringSubtypeSingleValue(SubtypeData d) returns string? {
```
jclark, line 43: This function should be handling char and nonChar uniformly. I think it can be written more clearly, so that it is obviously correct.

compiler/modules/types/string.bal
jclark: One can check for that case by checking that the sum of the lengths of the two arrays is equal to 0x11000 less the number of surrogate code points.

jclark (changes requested):

compiler/modules/types/string.bal
manuranga: how about `boolean modifier;` or `boolean adjective;`

with `const ALLOW = true;` `const DENY = false;` or `const REJECT = false;`

jclark: The problem with these (as with `allowed`) is that the name applies to `values` but they are a field on the subtype.

Some other ideas:

```
   INCLUSIVE|EXCLUSIVE spec;
```

It also might be better to use +1 and -1 (as a boolean) to say respectively that the values are included or excluded.

compiler/modules/types/string.bal
```diff
@@ -1,65 +1,153 @@
...
+        if char.allowed && nonChar.allowed {
+            return true;
+        }
+        else if !char.allowed && !nonChar.allowed {
+            return false;
+        }
+    }
+    return createStringSubtype(
+        { allowed: !char.allowed, values: char.values.cloneReadOnly() },
+        { allowed: !nonChar.allowed, values: nonChar.values.cloneReadOnly() }
```
manuranga: Aren't they already read-only?

compiler/tests/data/string-singleton.bal
rdhananjaya: Added compiler/tests/data/string-all-subtypes.bal

compiler/modules/types/string.bal
rdhananjaya: I tried `enum Spec { INCLUDE, EXCLUDE }`
It just bloated the downstream code. In my opinion, changing the name from `boolean allowed` to anything else didn't make it that much better either, do we need this change?

compiler/modules/front/symbols.bal
```diff
@@ -25,6 +25,8 @@ public type ExportedDefn bir:FunctionSignature|t:SemType|s:ResolvedConst;
 
 public type ModuleExports readonly & map<ExportedDefn>;
 
+final map<Import> implicitImports = {};
```
jclark: Please avoid mutable module-level variables. Bad for concurrency.

compiler/modules/front/symbols.bal
```diff
@@ -109,3 +115,22 @@ function lookupImportedConst(ModuleSymbols mod, s:ModuleLevelDefn modDefn, strin
     }
     return err:semantic(`${prefix + ":" + varName} is not defined as a public const`, loc=s:locationInDefn(modDefn), functionName=modDefn.name);
 }
+
+function lookupImplicitImport(string prefix) returns Import? {
+    Import? 'import = implicitImports[prefix];
```
jclark: Better not to use a keyword as a variable name. So many others to choose from.

compiler/modules/front/symbols.bal
```diff
@@ -109,3 +115,22 @@ function lookupImportedConst(ModuleSymbols mod, s:ModuleLevelDefn modDefn, strin
...
+    if 'import == () {
+        ModuleExportSemtypes? defns = predefinedSubtypes[prefix];
+        if defns != () {
+            bir:ModuleId moduleId = { org: "ballerina", names: ["lang", prefix] };
+            Import im = {
+                decl: { ...moduleId, prefix, pos: 0, partIndex: 0 },
```
jclark: It's rather ugly having to fake up a a ImportDecl.

compiler/modules/types/string.bal
```diff
@@ -1,65 +1,153 @@
...
+        nonChar = { allowed: true, values: [] };
+    }
+    else {
+        char = { allowed: true, values: [] };
+        nonChar = { allowed: true, values: [value] };
+    }
+    return uniformSubtype(UT_STRING, { char, nonChar });
+}
+
+public function stringChar() returns SemType {
```
jclark: We don't need this to be `public`. We can just use CHAR.

compiler/modules/types/core.bal
```diff
@@ -363,6 +363,7 @@ public final UniformTypeBitSet READONLY = uniformTypeUnion(UT_READONLY);
 public final UniformTypeBitSet SIMPLE_OR_STRING = uniformTypeUnion((1 << UT_NIL) | (1 << UT_BOOLEAN) | (1 << UT_INT)| (1 << UT_FLOAT)| (1 << UT_DECIMAL)| (1 << UT_STRING));
 public final UniformTypeBitSet NUMBER = uniformTypeUnion((1 << UT_INT) | (1 << UT_FLOAT) | (1 << UT_DECIMAL));
 public final SemType BYTE = intWidthUnsigned(8);
+public final SemType CHAR = stringChar();
```
jclark: I prefer to call this `STRING_CHAR`.

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -166,9 +156,18 @@ function parsePrimaryTypeDesc(Tokenizer tok) returns TypeDesc|err:Syntax {
             // JBUG should not need cast #30191
             return <LeafTypeDesc>cur;
         }
+        var prefix if prefix == "string" || prefix == "int" => {
```
jclark: Can't I just do `"string"|"int"` here?

compiler/modules/front/symbols.bal
jclark: Maybe create a supertype of Import in which decl is optional and return that.

compiler/modules/front/symbols.bal
```diff
@@ -109,3 +115,22 @@ function lookupImportedConst(ModuleSymbols mod, s:ModuleLevelDefn modDefn, strin
     }
     return err:semantic(`${prefix + ":" + varName} is not defined as a public const`, loc=s:locationInDefn(modDefn), functionName=modDefn.name);
 }
+
+function lookupImplicitImport(string prefix) returns Import? {
+    Import? 'import = implicitImports[prefix];
+    if 'import == () {
+        ModuleExportSemtypes? defns = predefinedSubtypes[prefix];
+        if defns != () {
+            bir:ModuleId moduleId = { org: "ballerina", names: ["lang", prefix] };
```
jclark: We create the moduleId here and also in front.bal which suggests we don't have quite the right structure.

compiler/modules/front/front.bal
```diff
@@ -176,6 +178,14 @@ function importPartPrefixes(ScannedModule scanned, (ModuleExports|string?)[] res
             resolved = ioLibFunctions;
             partial = true;
         }
+        else if moduleId == BALLERINA_LANG_INT {
```
jclark: This doesn't seem quite right. This is processing import declarations, but we don't have import declarations for these prefixes.  Why do we need this?

compiler/modules/front/symbols.bal
jclark: I think we need a readonly table containing records with fields for (1) prefix (2) module id and (3) `readonly & map<SemType>`

jclark (commented):
This almost all looks good, but I think the auto-import part can be improved.

compiler/modules/front/front.bal
rdhananjaya: My bad, will remove these.

compiler/modules/front.syntax/parseTypeDesc.bal
rdhananjaya: Type narrowing doesn't quite work, we can do:
```ballerina
        "string"
        |"int" => {
            "int"|"string" prefix = <"int"|"string"> cur;
            Position pos = tok.currentPos();
            check tok.advance();
            if tok.current() != ":" {
                return prefix;
            }
            check tok.advance();
            return { prefix, typeName: check tok.expectIdentifier(), pos };
        }
```

jclark:
> Type narrowing doesn't quite work, we can do:

I would probably use `LeafTypeDesc` instead of `"int"|"string"`: there's some other code in the function that does something similar.

compiler/modules/front.syntax/parseTypeDesc.bal
manuranga:
```suggestion
        var prefix if prefix is "string"|"int" => {
```

rdhananjaya: Went with `LeafTypeDesc`

compiler/modules/types/string.bal
```diff
@@ -1,65 +1,151 @@
...
 
 function stringSubtypeSingleValue(SubtypeData d) returns string? {
     if d is boolean {
         return ();
     }
-    StringSubtype s = <StringSubtype>d;
-    if !s.allowed {
-        return ();
+    var { char, nonChar } = <StringSubtype>d;
+    int charCount = char.values.length();
```
jclark: This still seems overly complicated to me. Can something like this work?

```
int charCount = char.allowed ? char.values.length() : 2;
int nonCharCount = nonChar.allowed ? nonChar.values.length() : 2;
if charCount + nonCharCount == 1 {
    return charCount != 0 ? char.values[0] : nonChar.values[0];
}
```

jclark (changes requested):

jclark:
The auto-import is now very neat!

compiler/modules/types/string.bal
rdhananjaya: Wow, that's clever.

manuranga: Little bit too clever. We could have written it better, if we had nil lifting `+`.

jclark:
> Little bit too clever. We could have written it better, if we had nil lifting `+`. 

Don't see how nil lifting would help.

jclark (approved):

manuranga:
> Don't see how nil lifting would help.

Unassuming reader might assume `2` has some special meaning, when it works equally well with `42`, we can avoid that confusion by using nil instead.

```ballerina
int? charCount = char.allowed ? char.values.length() : ();
int? nonCharCount = nonChar.allowed ? nonChar.values.length() : ();
if charCount + nonCharCount == 1 {
    return charCount != 0 ? char.values[0] : nonChar.values[0];
}
```

jclark:
Or more straightforwardly `const MANY = 2;`

PR #517, opened 2021-09-29
Generate init module

### LINKED ISSUE #511: Create an initialization module for each program
Currently, we create an LLVM module for each Ballerina module. With this issue, we will create an additional initialization module.

This module will be responsible for defining the program's main function, which will have a fixed name like `_bal_main`, and is called by C `main` function. This will call the user-defined main function in the root module, whose name depends on the name of root module (call this the "user main"). This will allow us to fix #456.

The initialization module will expand to deal with other things (which will be separate issues):
 * initialize every module in the program #512
 * process command-line arguments, turning the OS-provided arguments into the form expected by the user main
 * deal with possibility that LLVM return type of user main may be void or a tagged ptr
 * provide data structures the the program modules need for performing runtime type-checking #402 
 
For a root module `foo.bal` we will generate the initialization module into `foo._init.ll`. (We should implement #444 to ensure this cannot clash with a user module name.) The testing Makefile needs to deal with this.

nback will export a function that builds an initialization module, which will be called by compile.bal. The parameters to that will include
- a list of the module ids of all the programs modules, in the order in which the modules should be initialized (or the reverse of that)
- the ModuleExports for the root module (so that we know how to call main)

This is a prerequisite for #495.

manuranga:
Fixes #511

### DIFF
--- a/compiler/compile.bal
+++ b/compiler/compile.bal
@@ -41,6 +41,10 @@ class CompileContext {
         return nback:buildModule(birMod, self.llContext, self.nbackOptions);
     }
 
+    function buildInitModule(bir:ModuleId entryModId, map<bir:FunctionSignature> publicFuncs) returns LlvmModule|CompileError {
+        return nback:buildInitModule(entryModId, publicFuncs, self.llContext);
+    }
+
     function job(bir:ModuleId id) returns Job {
         Job? j = self.jobs[id];
         if j is () {
@@ -69,6 +73,7 @@ function compileBalFile(string filename, string basename, string? outputBasename
     CompileContext cx = new(basename, outputBasename, nbackOptions, outOptions);
     front:ResolvedModule mod = check processModule(cx, DEFAULT_ROOT_MODULE_ID, [ {filename} ], cx.outputFilename());
     check mod.validMain();
+    check generateInitModule(cx, mod);
 }
 
 function processModule(CompileContext cx, bir:ModuleId id, front:SourcePart[] sourceParts, string? outFilename) returns front:ResolvedModule|CompileError {
@@ -148,4 +153,22 @@ function subModuleSourceParts(string basename, bir:ModuleId id) returns front:So
 
 function subModuleSuffix(bir:ModuleId id) returns string {
     return ".".'join(...id.names.slice(1));
+}
+
+function generateInitModule(CompileContext cx, front:ResolvedModule entryMod) returns CompileError? {
+    LlvmModule initMod = check cx.buildInitModule(entryMod.getId(), filterFuncs(entryMod.getExports()));
+    string? initOutFilename = cx.outputFilename(".init");
+    if initOutFilename != () {
+        check outputModule(initMod, initOutFilename, cx.outputOptions);
+    }
+}
+
+function filterFuncs(front:ModuleExports defns) returns map<bir:FunctionSignature> {
+    map<bir:FunctionSignature> result = {};
+    foreach var [name, defn] in defns.entries() {
+        if defn is bir:FunctionSignature {
+            result[name] = defn;
+        }
+    }
+    return result;
 }
--- a/compiler/modules/nback/basic.bal
+++ b/compiler/modules/nback/basic.bal
@@ -1,4 +1,3 @@
-import wso2/nballerina.err;
 import wso2/nballerina.bir;
 import wso2/nballerina.types as t;
 import wso2/nballerina.print.llvm;
@@ -213,23 +212,12 @@ function buildFunctionDecl(Scaffold scaffold, bir:ExternalSymbol symbol, bir:Fun
         return decl;
     }
     else {
-        // TODO: fix this: scaffold.location(0)
-        llvm:FunctionType ty = check buildFunctionSignature(sig, scaffold.location(0));
+        llvm:FunctionType ty = buildFunctionSignature(sig);
         llvm:Module mod = scaffold.getModule();
         llvm:FunctionDecl d = mod.addFunctionDecl(mangleExternalSymbol(symbol), ty);
         scaffold.addImportedFunction(symbol, d);
         return d;
     }
-}
-
-function buildFunctionSignature(bir:FunctionSignature signature, err:Location loc) returns llvm:FunctionType|BuildError {
-    llvm:Type[] paramTypes = from var ty in signature.paramTypes select (semTypeRepr(ty)).llvm;
-    RetRepr repr = semTypeRetRepr(signature.returnType);
-    llvm:FunctionType ty = {
-        returnType: repr.llvm,
-        paramTypes: paramTypes.cloneReadOnly()
-    };
-    return ty;
 }
 
 function buildErrorConstruct(llvm:Builder builder, Scaffold scaffold, bir:ErrorConstructInsn insn) returns BuildError? {
--- a/compiler/modules/nback/build.bal
+++ b/compiler/modules/nback/build.bal
@@ -393,3 +393,13 @@ function buildConstBoolean(boolean b) returns llvm:Value {
 function heapPointerType(llvm:Type ty) returns llvm:PointerType {
     return llvm:pointerType(ty, HEAP_ADDR_SPACE);
 }
+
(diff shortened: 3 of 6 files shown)


compiler/modules/nback/basic.bal
```diff
@@ -222,7 +222,8 @@ function buildFunctionDecl(Scaffold scaffold, bir:ExternalSymbol symbol, bir:Fun
     }
 }
 
-function buildFunctionSignature(bir:FunctionSignature signature, err:Location loc) returns llvm:FunctionType|BuildError {
+// pr-todo: remove loc param
+function buildFunctionSignature(bir:FunctionSignature signature, err:Location? loc = ()) returns llvm:FunctionType|BuildError {
```
manuranga: I am planing to remove  `err:Location loc`.
@heshanpadmasiri

compiler/modules/nback/init.bal
```diff
@@ -0,0 +1,24 @@
...
+    check buildMain(entryModId, userMainName, publicFuncs.get(userMainName), llMod, builder);
+    return llMod;
+}
+
+public function buildMain(bir:ModuleId entryModId, string userMainName, bir:FunctionSignature userMainSig, llvm:Module llMod, llvm:Builder builder) returns BuildError? {
+    llvm:FunctionType ty = { returnType: "void", paramTypes: [] };
+    llvm:FunctionDefn func = llMod.addFunctionDefn("_bal_main", ty);
+    builder.positionAtEnd(func.appendBasicBlock());
+
+    string userMainMangledName = mangleInternalSymbol(entryModId, {identifier: userMainName, isPublic: true});
```
jclark: Mapping constructor missing spaces

compiler/modules/nback/init.bal
```diff
@@ -0,0 +1,24 @@
+import wso2/nballerina.bir;
+import wso2/nballerina.print.llvm;
+
+public function buildInitModule(bir:ModuleId entryModId, map<bir:FunctionSignature> publicFuncs, llvm:Context llContext) returns llvm:Module|BuildError {
+    llvm:Module llMod = llContext.createModule();
+    llvm:Builder builder = llContext.createBuilder();
+    // SUBSET support cases with no/different entry function
+    string userMainName = "main";
+    check buildMain(entryModId, userMainName, publicFuncs.get(userMainName), llMod, builder);
```
jclark: front-end isn't requiring main: it just requires that if main exists, then it has to have a particular signature and be public. Not sure we want to require main for E test cases.  A Ballerina program doesn't have to have a main (think something that just starts a server). So I think for now our _bal_main should just not call anything if there is no user main. Maybe we can add an error/warning down the road.

compiler/modules/nback/basic.bal
jclark: Move this into build.bal since it's going to be used by init.bal.

compiler/compile.bal
```diff
@@ -68,7 +72,9 @@ class CompileContext {
 function compileBalFile(string filename, string basename, string? outputBasename, nback:Options nbackOptions, OutputOptions outOptions) returns CompileError? {
     CompileContext cx = new(basename, outputBasename, nbackOptions, outOptions);
     front:ResolvedModule mod = check processModule(cx, DEFAULT_ROOT_MODULE_ID, [ {filename} ], cx.outputFilename());
+    // pr-todo: find if this should be moved to init.bal
```
jclark: No, because nback doesn't depend on front.

jclark (changes requested):
Does sub.mk need changes or does it work as is?

compiler/compile.bal
manuranga: I didn't mean just the method call, I mean move the whole logic to `init.bal`.

runtime/main.c
```diff
@@ -2,12 +2,12 @@
 #include "hash.h"
 
 char *_bal_stack_guard;
-extern void BAL_ROOT_NAME(main)();
```
manuranga, line 5: This is no longer used, but I didn't remove the definition.

compiler/modules/nback/init.bal
manuranga: Ahg fixed, only if we had a style guide.

manuranga:
It works as is, didn't need to change `sub.mk`.

compiler/compile.bal
manuranga: Discussed in the meeting, this logic belongs in the frontend.

compiler/modules/nback/init.bal
```diff
@@ -0,0 +1,25 @@
...
+
+public function buildInitModule(bir:ModuleId entryModId, map<bir:FunctionSignature> publicFuncs, llvm:Context llContext) returns llvm:Module|BuildError {
+    llvm:Module llMod = llContext.createModule();
+    llvm:Builder builder = llContext.createBuilder();
+    string userMainName = "main";
+    check buildMain(entryModId, userMainName, publicFuncs[userMainName], llMod, builder);
+    return llMod;
+}
+
+public function buildMain(bir:ModuleId entryModId, string userMainName, bir:FunctionSignature? userMainSig, llvm:Module llMod, llvm:Builder builder) returns BuildError? {
```
jclark: Shouldn't be public.

compiler/modules/nback/init.bal
```diff
@@ -0,0 +1,25 @@
+import wso2/nballerina.bir;
+import wso2/nballerina.print.llvm;
+
+public function buildInitModule(bir:ModuleId entryModId, map<bir:FunctionSignature> publicFuncs, llvm:Context llContext) returns llvm:Module|BuildError {
+    llvm:Module llMod = llContext.createModule();
+    llvm:Builder builder = llContext.createBuilder();
+    string userMainName = "main";
```
jclark: I would make this a const for now.

compiler/modules/nback/init.bal
```diff
@@ -0,0 +1,25 @@
...
+public function buildMain(bir:ModuleId entryModId, string userMainName, bir:FunctionSignature? userMainSig, llvm:Module llMod, llvm:Builder builder) returns BuildError? {
+    llvm:FunctionType ty = { returnType: "void", paramTypes: [] };
+    llvm:FunctionDefn func = llMod.addFunctionDefn("_bal_main", ty);
+    builder.positionAtEnd(func.appendBasicBlock());
+
+    if userMainSig != () {
+        string userMainMangledName = mangleInternalSymbol(entryModId, { identifier: userMainName, isPublic: true });
+        llvm:FunctionType userMainTy = check buildFunctionSignature(userMainSig);
+        llvm:FunctionDecl userMainDecl = llMod.addFunctionDecl(userMainMangledName, userMainTy);
+        llvm:Value? retValue = builder.call(userMainDecl, []);
```
jclark: Not using retValue.

jclark (changes requested):

compiler/modules/nback/build.bal
```diff
@@ -393,3 +393,13 @@ function buildConstBoolean(boolean b) returns llvm:Value {
 function heapPointerType(llvm:Type ty) returns llvm:PointerType {
     return llvm:pointerType(ty, HEAP_ADDR_SPACE);
 }
+
+function buildFunctionSignature(bir:FunctionSignature signature) returns llvm:FunctionType|BuildError {
```
jclark: This doesn't need to return BuildError any more does it?

compiler/compile.bal
```diff
@@ -149,3 +154,21 @@ function subModuleSourceParts(string basename, bir:ModuleId id) returns front:So
 function subModuleSuffix(bir:ModuleId id) returns string {
     return ".".'join(...id.names.slice(1));
 }
+
+function buildInitModule(CompileContext cx, front:ResolvedModule entryMod) returns CompileError? {
```
jclark: We shouldn't use the same name for this as for the method on CompileContext when they do different things. buildModule is returning a llvm:MOdule, so I would call this something different, perhaps generateInitModule.

jclark (changes requested):

Issue #513, opened 2021-09-29
Add positions to AST nodes


jclark:
Every AST node should have a Position for its start and its end. The start position should be the start of the first token that is part of what it represents; the end position should be the end of the last token that is part of what it represents.

This can be done with startPos, endPos fields in the AST nodes.

jclark:
We can split AST nodes into groups
1. Top-level declarations: ModuleLevelDefn and ImportDecl
2. Statements
3. Expressions
4. Type descriptors

Let's do each of these in a separate pull request, starting with ModuleLevelDefns.

jclark:
For ModuleLevelDefn, the existing pos should be renamed to namePos and should correspond to the start of the name being defined by the defn.

jclark:
Create a record type

```
type PositionFields record {|
   Position startPos;
   Position endPos;
|};
```

And then use `*PositionFields;` to include it in AST nodes.

jclark:
We also need to come up with a good test strategy.  For group 1: everything other than whitespace or comments should be within exactly one top-level defn.  I think you will need to write code that maps from a Position onto an index of a fragment within a ScannedLine (you have to count the length of each fragment).

Do top-level defns first; then come up the testing strategy.

jclark:
I think we will need a test function: 

```
function testIsWhitespace(SourceFile file, Position startPos, Position endPos) returns boolean
```

that tests whether all fragments between startPos and endPos are whitespace (e.g. FRAG_WHITESPACE or FRAG_COMMENT).

heshanpadmasiri:
I have a suggestion for a possible test strategy for general case

1. We extend the tokenizer so that we can move forward it up to a given position
2. We'll use 1 to seek the tokenizer forward up to start position 
3. Then call the parser function for the given node (ex:`TypeDefn` -> `parseTypeDefinition`)
4. If the start position is wrong parser will fail
5. Else after the parser function we look at the tokenizer's current position and see if that is same as the end position for the ast node

heshanpadmasiri:
@jclark for the line ([testcase](https://github.com/ballerina-platform/nballerina/blob/main/compiler/testSuite/05-const/2-v.bal))
```
const N1 = 100;
```
I am getting fragcodes as `[135, 0, 3, 0, 61, 0, 4, 59]` (with fragments being `[" ",  "N1",  " ", " ", "100"]`) is this the expected behavior. (problem being fragcodes having FRAG_WHITESPACE(0))

[...]

PR #519, opened 2021-10-01
Add positions to top level declaration AST nodes

heshanpadmasiri:
## Purpose
Add positions to top level declaration (ModuleLevelDefn and ImportDecl)  AST nodes
Related to #513

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/front.syntax/token.bal
```diff
@@ -249,6 +251,16 @@ class Tokenizer {
         return createPosition(self.lineIndex, self.tokenStartCodePointIndex);
     }
 
+    function previousPos() returns Position {
```
jclark: I don't like this at all.  Too stateful. Instead we should have a function that returns the Position of the end of the current token (currentPos() returns the start of the current token).  We should probably rename currentPos() to currentStartPos(); then the new function would be currentEndPos(). An endPos should always come from calling currentEndPos() and a startPos should always come from calling currentStartPos() (otherwise whitespace will be in the wrong place).

compiler/modules/front.syntax/parse.bal
```diff
@@ -133,38 +137,53 @@ function parseModuleDecl(Tokenizer tok, ModulePart part) returns ModuleLevelDefn
...
     InlineBuiltinTypeDesc? td = ();
     if t is InlineBuiltinTypeDesc {
         check tok.advance();
         td = t;
     }
+    Position namePos = tok.currentPos();
     string name = check tok.expectIdentifier();
     check tok.expect("=");
     Expr expr = check parseInnerExpr(tok);
+    Position endPos = tok.currentPos();
```
jclark: We want the endPos to be the end of the `;` token. The way to do that is to have a function like `expect` that returns the endPos of the expected token.

compiler/modules/front.syntax/parse.bal
```diff
@@ -133,38 +137,53 @@ function parseModuleDecl(Tokenizer tok, ModulePart part) returns ModuleLevelDefn
 }
 
 function parseTypeDefinition(Tokenizer tok, ModulePart part, Visibility vis) returns TypeDefn|err:Syntax {
```
jclark, line 135: You should parse in the startPos of the declaration as a parameter.

jclark (changes requested):

compiler/modules/front.syntax/token.bal

compiler/modules/front.syntax/parse.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/parse.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/token.bal
```diff
@@ -245,10 +245,25 @@ class Tokenizer {
         self.mode = m;
     }
 
-    function currentPos() returns Position {
+    function currentStartPos() returns Position {
         return createPosition(self.lineIndex, self.tokenStartCodePointIndex);
     }
 
+    function currentEndPos() returns Position {
+        Token? tok = self.current();
```
jclark: This should be done using self.codePointIndex. Not sure if anything else is needed (don't think there is). If there is, current() should set it up and it should be returned here.

jclark (changes requested):

compiler/modules/front.syntax/token.bal

compiler/modules/front.syntax/tests/testParserOnTestSuite.bal
```diff
@@ -22,6 +22,25 @@ function testParserOnTestSuite() returns err:Syntax|io:Error|file:Error? {
...
+                topLevelDefnPos = topLevelDefnPos.sort();
+                err:Position lastEnd = 1<<32;
+                foreach var [startPos, endPos] in topLevelDefnPos {
+                    test:assertTrue(startPos >= lastEnd, "overlapping top level definitions");
+                    test:assertTrue(startPos <= endPos, "invalid start and end positions");
+                    SourceFile file = part.file;
+                    var [stLine, stCol] = file.lineColumn(startPos);
+                    var [endLine, endCol] = file.lineColumn(lastEnd);
+                    string errorBody = string ` filename: ${file.filename()} between (${endLine}, ${endCol})  and (${stLine}, ${stCol})`;
+                    test:assertTrue(testIsWhitespace(part.file, lastEnd, startPos), "none white space tokens between top level definition" + errorBody);
```
jclark, line 41: You don't have to have whitespace between top-level definitions:

const X = 1;const Y = 2;

The point is that if there is any between, it should not be between the startPos/endPos of any of the definitions.

compiler/modules/front.syntax/tests/testParserOnTestSuite.bal
```diff
@@ -44,3 +63,85 @@ function partToLines(ModulePart part) returns string[] {
...
+            }
+            i += 1;
+        }
+        i = 0;
+        lineIndex += 1;
+    }
+    return true; // start == end
+}
+
+function fragIndex(err:LineColumn lineColumn, ScannedLine line) returns int {
```
jclark: I would hope to have a method on sourceFile that takes a Position and returns the FragCode at that position. This is a generally useful function that would not be just in test.

compiler/modules/front.syntax/tests/testParserOnTestSuite.bal
heshanpadmasiri: I  am considering the the range from the last top level definition's endposition (or begining of the file for the first definition) and start position of current definition.  In a case like this where this range is empty `testIsWhitespace` will return true.

compiler/modules/front.syntax/tests/testParserOnTestSuite.bal
heshanpadmasiri: Moved the function to `SourceFile`

heshanpadmasiri:
@jclark most top level parsers call another parser function (I'll call these "internal parsers") (for example `parseFunctionDefinition` calls `parseStmtBlock`) and these "internal parsers" will advance to next token at the end of parsing. This means when we return to the top level parser tokenizer can be at the beginning of the definition if we call this internal parser at the end of top level parser(ex:`parseFunctionDefinition` -> `parseStmtBlock`). Currently I am using the current token's (after returning) start position as the end position for the AST nodes created by the top level definition. Is this correct? If not I can think of two solutions

1. Make all such internal parsers return their end positions
2. Stop them from advancing at the end

We are going to have this problem only until we have implemeted end position for all AST nodes, at which point I can use the endPos of the last child node (in our example last `Stmt`) as the endPos of the parent and we can validate this happend correctly by removing the equality from greater than or equal in overlapping test
```
test:assertTrue((startPos == (1<<32)) || (startPos >= lastEnd), "overlapping top level definitions");
```

jclark:
Definitely not OK to use next token's start position as the end position. That will include whitespace before the end position that should not be included. Will need to think about best solution.

jclark:
I think the least painful way to  do this is to do something similar to what you did before. Add state to Tokenizer so that it can return the position of the end of the previous token. This will need two extra fields in Tokenizer: prevTokenEndCodePointIndex and prevTokenEndLineIndex. You set these fields at the beginning of advance() to codePointIndex and lineIndex respectively. You will need to handle saving and restoring.

heshanpadmasiri:
> I think the least painful way to do this is to do something similar to what you did before. Add state to Tokenizer so that it can return the position of the end of the previous token. This will need two extra fields in Tokenizer: prevTokenEndCodePointIndex and prevTokenEndLineIndex. You set these fields at the beginning of advance() to codePointIndex and lineIndex respectively. You will need to handle saving and restoring.

fixed

compiler/modules/front.syntax/token.bal
```diff
@@ -369,7 +393,42 @@ public readonly class SourceFile {
         return [pos >> 32, pos & 0xFFFFFFFF];
     }
 
+    public function fragIndex(Position pos) returns int {
```
jclark: Accessing the internals of a ScannedLine should be done by a function in scan.bal.  This function isn't quite doing the right thing.  What we want is to from a SourceFile+Position to a `[ScannedLine, fragIndex]` tuple. Then `testIsWhitespace` function should become much simpler.  Using lineColumn isn't quite right. You should have a function unpackPosition (next to createPosition) that returns `[int,int]`; then lineColumn should call unpackPosition.

compiler/modules/front.syntax/token.bal
```diff
@@ -245,10 +251,18 @@ class Tokenizer {
...
-    function currentPos() returns Position {
+    function currentStartPos() returns Position {
         return createPosition(self.lineIndex, self.tokenStartCodePointIndex);
     }
 
+    function currentEndPos() returns Position {
+        return createPosition(self.lineIndex, self.codePointIndex);
+    }
+
+    function prevEndPos() returns Position {
```
jclark: To be consistent, this should be `previousEndPos` (or we could have `curStartPos` and `curEndPos`)

jclark (changes requested):

compiler/modules/front.syntax/token.bal

compiler/modules/front.syntax/token.bal
heshanpadmasiri: Added two functions one to give `[ScannedLineIndex, fragIndex]` and `[ScannedLine, fragIndex]` (which wraps the first one). First one is there for functions such as `testIsWhitespace` that iterate over a range which include multiple `ScannedLines` so they can start with given line index and increment. Change `lineColumn` as suggested  and added functions to get `fragement` and `fragCode` for a given line,index combination to `scan.bal`

compiler/modules/front.syntax/token.bal
```diff
@@ -366,10 +394,48 @@ public readonly class SourceFile {
...
-        return [pos >> 32, pos & 0xFFFFFFFF];
+        return unpackPosition(pos);
+    }
+
+    function fragLine(Position pos) returns [readonly & ScannedLine, int] {
+        var [lineIndex, fragIndex] = self.fragIndex(pos);
+        return [self.line(lineIndex), fragIndex];
+    }
+
+    function fragIndex(Position pos) returns [int, int] {
```
jclark, line 405:
```
function fragment(Position pos) returns [ScannedLine, int] {
    var [lineIndex, codePointIndex] = unpackPosition(pos);
    ScannedLine line = self.lines[lineIndex];
    return [line, scannedLineFragIndex(line, codePointIndex)];
}
```

jclark (changes requested):

compiler/modules/front.syntax/tests/testParserOnTestSuite.bal
```diff
@@ -44,3 +63,26 @@ function partToLines(ModulePart part) returns string[] {
 function scanAndParseModulePart(string[] lines, FilePath path, int partIndex) returns ModulePart|err:Syntax {
     return parseModulePart(check scanModulePart(createSourceFile(lines, path), partIndex));
 }
+
+function testIsWhitespace(SourceFile file, Position startPos, Position endPos) returns boolean {
+    var [startLineIndex, startFragIndex] = file.fragIndex(startPos);
```
jclark, line 68: I was suggesting to use the function that returns a [ScannedLine, int], but I now see that doesn't work because you need the index to get the lines between the startPos and endPos.

jclark (changes requested):

compiler/modules/front.syntax/tests/testParserOnTestSuite.bal
jclark: Let me merge it and see if I can improve it.

PR #520, opened 2021-10-01
Generate error for unused variables

### LINKED ISSUE #357: Make unused local variables be an error
The spec currently says nothing about unused variables, but my expectation is that this will become an error.

We should implement this now to avoid accumulating test cases with unused variables.

To make this convenient, we should first implement #356 (so you can so `_ = x;` to "use" `x`).

KavinduZoysa:
## Purpose
Fixes #357

### DIFF
--- a/compiler/modules/front/codeGen.bal
+++ b/compiler/modules/front/codeGen.bal
@@ -367,32 +367,18 @@ function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialE
 }
 
 function unusedLocalVariables(CodeGenContext cx, Environment env, Binding? bindingLimit) returns CodeGenError? {
-    Binding? tem = env.bindings;
+    Binding? binding = env.bindings;
     while true {
-        if bindingLimit is () {
-            if tem is () {
-                break;
-            }
-            else {
-                if isUnused(tem) {
-                    return cx.semanticErr(`unused local variable ${tem.name}`);
-                }
-                tem = tem.prev;
-            }
+        if binding === bindingLimit {
+            break;
         }
         else {
-            if tem is () {
-                break;
+            // binding is non-nil
+            Binding tem = <Binding>binding;
+            if isUnused(tem) {
+                return cx.semanticErr(`unused local variable ${tem.name}`);
             }
-            else if tem.name == bindingLimit.name {
-                break;
-            }
-            else {
-                if isUnused(tem) {
-                    return cx.semanticErr(`unused local variable ${tem.name}`);
-                }              
-                tem = tem.prev;
-            }
+            binding = tem.prev;
         }
     }
     return ();
--- a/compiler/modules/front/codeGen.bal
+++ b/compiler/modules/front/codeGen.bal
@@ -365,7 +365,7 @@ function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialE
     check unusedLocalVariables(cx, env);
     int[] assignments = [];
     addAssignments(assignments, env.assignments, startRegister);
-    return { block: curBlock, bindings: env.bindings, assignments };
+    return { block: curBlock, assignments };
 }
 
 function unusedLocalVariables(CodeGenContext cx, Environment env) returns CodeGenError? {
--- a/compiler/modules/front/codeGen.bal
+++ b/compiler/modules/front/codeGen.bal
@@ -8,8 +8,6 @@ type Environment record {|
     // A list of registers that were narrowed but have been assigned to
     // Holds the number of the original, unnarrowed register
     int[] assignments = [];
-    // A list of local variable names created for given environment
-    string[] localVarNames = [];
 |};
 
 type Binding record {|
@@ -362,40 +360,52 @@ function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialE
             env.assignments.push(...effect.assignments);
         }
     }                
-    check unusedLocalVariables(cx, env);
+    check unusedLocalVariables(cx, env, initialEnv.bindings);
     int[] assignments = [];
     addAssignments(assignments, env.assignments, startRegister);
     return { block: curBlock, assignments };
 }
 
-function unusedLocalVariables(CodeGenContext cx, Environment env) returns CodeGenError? {
-    Binding? bindings = env.bindings;
-    foreach var varName in env.localVarNames {
-        if isUnused(varName, bindings) {
-            return cx.semanticErr(`unused local variable ${varName}`);
+function unusedLocalVariables(CodeGenContext cx, Environment env, Binding? bindingLimit) returns CodeGenError? {
+    Binding? tem = env.bindings;
+    while true {
+        if bindingLimit is () {
+            if tem is () {
+                break;
+            }
+            else {
+                if isUnused(tem) {
+                    return cx.semanticErr(`unused local variable ${tem.name}`);
+                }
+                tem = tem.prev;
+            }
+        }
+        else {
+            if tem is () {
+                break;
+            }
+            else if tem.name == bindingLimit.name {
+                break;
+            }
+            else {
+                if isUnused(tem) {
+                    return cx.semanticErr(`unused local variable ${tem.name}`);
+                }              
+                tem = tem.prev;
+            }
         }
     }
     return ();
 }
 
-function isUnused(string varName, Binding? bindings) returns boolean {
-    Binding? tem = bindings;
-    while true {
-        if tem is () {
-            break;
-        }
-        else {
-            Binding? narrowed = tem.unnarrowed;
-            if !(narrowed is ()) {
-                tem = narrowed;
-            }
-            if tem.name == varName && !tem.used {
-                return true;
-            }
-            tem = tem.prev;
-        }
+function isUnused(Binding binding) returns boolean {
+    Binding? unnarrowed = binding.unnarrowed;
+    if unnarrowed is () {
+        return !binding.used;
     }
-    return false;   
+    else {
+        return !unnarrowed.used;
+    }
 }
 
 function environmentCopy(Environment env) returns Environment {
--- a/compiler/modules/front/codeGen.bal
+++ b/compiler/modules/front/codeGen.bal
@@ -376,7 +376,6 @@ function unusedLocalVariables(CodeGenContext cx, Environment env, Binding? bindi
         }
         binding = tem.prev;
     }
-    return ();
 }
 
 function environmentCopy(Environment env) returns Environment {
--- a/compiler/modules/front/codeGen.bal
+++ b/compiler/modules/front/codeGen.bal
@@ -8,12 +8,15 @@ type Environment record {|
     // A list of registers that were narrowed but have been assigned to
     // Holds the number of the original, unnarrowed register
     int[] assignments = [];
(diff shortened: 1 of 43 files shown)


compiler/modules/front/codeGen.bal
```diff
@@ -359,11 +362,42 @@ function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialE
             env.assignments.push(...effect.assignments);
         }
     }                
+    check unusedLocalVariables(cx, env);
     int[] assignments = [];
     addAssignments(assignments, env.assignments, startRegister);
     return { block: curBlock, assignments };
 }
 
+function unusedLocalVariables(CodeGenContext cx, Environment env) returns CodeGenError? {
```
jclark: There's a simpler way to do this. Pass in a parameter `Binding? limit` which is initialEnv.bindings in codeGenStmts. Then check the used field in each binding in env up to limit that is not a narrowed binding. You don't need localVarNames.

jclark (changes requested):

compiler/modules/front/codeGen.bal
KavinduZoysa: Changed

compiler/modules/front/codeGen.bal
```diff
@@ -359,11 +360,54 @@ function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialE
...
     }                
+    check unusedLocalVariables(cx, env, initialEnv.bindings);
     int[] assignments = [];
     addAssignments(assignments, env.assignments, startRegister);
     return { block: curBlock, assignments };
 }
 
+function unusedLocalVariables(CodeGenContext cx, Environment env, Binding? bindingLimit) returns CodeGenError? {
+    Binding? tem = env.bindings;
+    while true {
```
jclark: This looks too complicated. You can stop when `tem === bindingLimit`. If we haven't reached that, then tem must be non-nil.

jclark (changes requested):

compiler/modules/front/codeGen.bal
KavinduZoysa: Fixed

compiler/modules/front/codeGen.bal
```diff
@@ -359,11 +360,40 @@ function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialE
...
     }                
+    check unusedLocalVariables(cx, env, initialEnv.bindings);
     int[] assignments = [];
     addAssignments(assignments, env.assignments, startRegister);
     return { block: curBlock, assignments };
 }
 
+function unusedLocalVariables(CodeGenContext cx, Environment env, Binding? bindingLimit) returns CodeGenError? {
+    Binding? binding = env.bindings;
+    while true {
```
jclark: while binding !== bindingLimit

compiler/modules/front/codeGen.bal
```diff
@@ -359,11 +360,40 @@ function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialE
...
+        if binding === bindingLimit {
+            break;
+        }
+        else {
+            // binding is non-nil
+            Binding tem = <Binding>binding;
+            if isUnused(tem) {
```
jclark: I think you can just do `if tem.unnarrowed == () && !tem.used`. The unnarrowed binding must have been added before bindingLimit; you can just ignore the narrowed one.

jclark (changes requested):

jclark:
In order to make sure we identify all unused variables in existing tests, I suggest
- make unusedLocalVariables print out a message for each unused variable (instead of returning an error)
- run it on all tests cases other than 10-unused to identify all unused variables in tests
- fix these
- then make it return an error
- maybe make the test runner report an error if it gets an unused variable error and neither the directory nor file contain "unused"

compiler/testSuite/01-function/call13-e.bal
```diff
@@ -1,6 +1,6 @@
 // @productions return-stmt local-var-decl-stmt int-literal
 public function main() {
-    int x = 1;
+    _ = 1;
```
jclark: This is not an OK change!!!!! Fix like this:

```
int x = 1;
_ = x;
```

jclark (changes requested):

compiler/testSuite/01-function/call13-e.bal
KavinduZoysa: @jclark, since we are testing `unreachable code` error in this test case, can you please explain the reason for the above particular change? 
What I thought was, since we are testing `unreachable code`, it does not matter the statements we have before the return-stmt.

jclark: You are changing what the test is testing.  One of the things that the original test is checking is that `1` can be assigned to a variable of type `int`. With your change, that is no longer being tested. You should only ever be adding lines that are assignments to `_`.

jclark (changes requested):

compiler/modules/front/codeGen.bal
KavinduZoysa: Fixed

compiler/modules/front/codeGen.bal
KavinduZoysa: Fixed

compiler/testSuite/01-function/call13-e.bal
KavinduZoysa: Thank you, I understood.
Fixed

manuranga: Five exclamation marks, the sure sign of an insane mind. ― Sir Terry Pratchett

jclark (changes requested):
I did a few cases that use an `ignore` function for the same purpose as `_`. Can you also fix them to use `_`?

```
compiler/testSuite/03-list/01-e.bal:public function ignore(any x) {
compiler/testSuite/04-list/1-e.bal:function ignore(any x) { }
compiler/testSuite/04-map/01-e.bal:function ignore(any x) { }
compiler/testSuite/04-map/03-e.bal:function ignore(any x) { }
compiler/testSuite/07-array/int1-e.bal:function ignore(any x) {}
compiler/testSuite/07-array/int4-e.bal:function ignore(any x) {}
compiler/testSuite/07-array/int7-e.bal:function ignore(any x) {}
compiler/testSuite/07-map/float2-v.bal:function ignore(any x) {}
compiler/testSuite/07-map/int1-e.bal:function ignore(any x) {}
compiler/testSuite/07-map/int4-e.bal:function ignore(any x) {}
compiler/testSuite/07-map/int5-p.bal:function ignore(any x) {}
compiler/testSuite/07-map/int7-e.bal:function ignore(any x) {}
```

There are some others with a parameter type other than `any`. Let's leave them alone for now.

compiler/modules/front/codeGen.bal
```diff
@@ -359,11 +360,25 @@ function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialE
...
+    Binding? binding = env.bindings;
+    while binding !== bindingLimit {
+        // binding is non-nil
+        Binding tem = <Binding>binding;
+        if tem.unnarrowed == () && !tem.used {
+            return cx.semanticErr(`unused local variable ${tem.name}`);
+        }
+        binding = tem.prev;
+    }
+    return ();
```
jclark: Don't need `return ();` here.

compiler/testSuite/04-misc/evalorder-v.bal
```diff
@@ -14,55 +14,55 @@ function two()  returns int{
 
 public function main() {
    int mul = one() * two();        // @output 1
-                                   // @output 2
+   _ = mul;                        // @output 2
```
jclark: This is confusing because the line is doing the output. Please add your line after both comments.

jclark (changes requested):
Also need to add some tests that show that variable shadowing from type narrowing is interacting correctly with this.

compiler/testSuite/04-misc/evalorder-v.bal
KavinduZoysa: Fixed

compiler/modules/front/codeGen.bal
KavinduZoysa: Fixed

KavinduZoysa:
> Also need to add some tests that show that variable shadowing from type narrowing is interacting correctly with this.

Added

KavinduZoysa:
> I did a few cases that use an `ignore` function for the same purpose as `_`. Can you also fix them to use `_`?
> 
> ```
> compiler/testSuite/03-list/01-e.bal:public function ignore(any x) {
> compiler/testSuite/04-list/1-e.bal:function ignore(any x) { }
> compiler/testSuite/04-map/01-e.bal:function ignore(any x) { }
> compiler/testSuite/04-map/03-e.bal:function ignore(any x) { }
> compiler/testSuite/07-array/int1-e.bal:function ignore(any x) {}
> compiler/testSuite/07-array/int4-e.bal:function ignore(any x) {}
> compiler/testSuite/07-array/int7-e.bal:function ignore(any x) {}
> compiler/testSuite/07-map/float2-v.bal:function ignore(any x) {}
> compiler/testSuite/07-map/int1-e.bal:function ignore(any x) {}
> compiler/testSuite/07-map/int4-e.bal:function ignore(any x) {}
> compiler/testSuite/07-map/int5-p.bal:function ignore(any x) {}
> compiler/testSuite/07-map/int7-e.bal:function ignore(any x) {}
> ```
> 
> There are some others with a parameter type other than `any`. Let's leave them alone for now.

Changed

compiler/testSuite/01-function/assign12-v.bal
```diff
@@ -0,0 +1,12 @@
+// @productions local-var-decl-stmt if-else-stmt string-literal
```
jclark: Why is this added?

compiler/testSuite/04-map/01-e.bal
```diff
@@ -1,8 +1,6 @@
-// @productions map-type-descriptor mapping-constructor-expr any function-call-expr local-var-decl-stmt int-literal
+// @productions map-type-descriptor mapping-constructor-expr any local-var-decl-stmt int-literal
 public function main() {
     map<any> m = {};
     int i = 2;
-    ignore(m[i]); // @error
+    _ = m[i]; // @error
```
jclark: Leave this one as is.

jclark (commented):
We need some `-v` tests. For example: unused param is ok; unused function is ok; using only narrowed variable is ok. Using v only in `v[i] = e` is OK.

compiler/testSuite/01-function/assign12-v.bal
KavinduZoysa: This is a `-v` test to check narrowed used variable. I added it under assignment tests. Now it is moved to `10-unused` test folder.

compiler/testSuite/04-map/01-e.bal
KavinduZoysa: Changed

KavinduZoysa:
> We need some `-v` tests. For example: unused param is ok; unused function is ok; using only narrowed variable is ok. Using v only in `v[i] = e` is OK.

Added

Issue #133, opened 2021-06-25
Implement nil type descriptor


jclark:
Specified here https://ballerina.io/spec/lang/2021R1/#nil-type-descriptor

jclark:
I think we may have already allowed this in via type definitions. We need to check and make sure it works.

manuranga:
I'll first send a PR with just tests for the already working cases.

PR #521, opened 2021-10-03
Remove inline typedesc and add nil type desc

manuranga:
Part of #412

Opened for discussion purposes.


jclark:
nil-type-desc is a separate issue: #133.

There's not much point in allowing parentheses only at the top-level in a type definition, but it is useful inline.  So when #412 is implemented, we can either do what you suggest here or do

inline-type-desc = type-desc | "(" type-desc ")"

jclark (changes requested):
Will look again when #412 or #133 is done.

PR #528, opened 2021-10-05
Add nil type tests

manuranga:
Testing type references to nil type.

Should have been working since c63d9c9
Doing this in preparation for #133


manuranga:
Should I move these to `07-type`?

jclark:
09-nil is good

PR #530, opened 2021-10-05
Allow _ for local-var-decl-stmt

### LINKED ISSUE #492: Allow a variable declaration to use `_`
Allow `_` in a variable declaration to support the idiom described here: https://github.com/ballerina-platform/ballerina-spec/issues/887#issuecomment-902334830

_Originally posted by @jclark in https://github.com/ballerina-platform/nballerina/issues/356#issuecomment-905958650_

Follow on from #356

KavinduZoysa:
## Purpose
Fixes #492


compiler/modules/front/codeGen.bal
```diff
@@ -766,10 +766,19 @@ function codeGenVarDeclStmt(CodeGenContext cx, bir:BasicBlock startBlock, Enviro
...
-    bir:Register result = cx.createRegister(semType, varName);
-    bir:BasicBlock nextBlock = check codeGenAssign(cx, env, startBlock, result, initExpr, semType);
-    return { block: nextBlock, bindings: { name: varName, reg: result, prev: env.bindings, isFinal } };  
+    bir:Register result;
+    bir:BasicBlock nextBlock;
+    if varName == "_" {
+        result = cx.createRegister(t:ANY, varName);
+        nextBlock = check codeGenAssign(cx, env, startBlock, result, initExpr, t:ANY);
+        return { block: nextBlock };
+    }
```
KavinduZoysa: I have followed this approach because rhs should belong to type any if lhs has wild-card-binding-pattern. Therefore `error _ = error("Error");` should be an error.
@jclark, WDYT about this approach?

jclark: In a VarDecl the type of the register needs to come from the type declaration (that's the whole point https://github.com/ballerina-platform/ballerina-spec/issues/887#issuecomment-902334830) You need to explicitly check for and give an error if the intersection of the declared type and error is non-empty. See last para of https://ballerina.io/spec/lang/2021R1/#section_5.9.2

jclark (changes requested):

compiler/modules/front/codeGen.bal
KavinduZoysa: Fixed

compiler/modules/front.syntax/parseStmt.bal
```diff
@@ -250,7 +250,14 @@ function parseVarDeclStmt(Tokenizer tok, boolean isFinal = false) returns VarDec
 
 function finishVarDeclStmt(Tokenizer tok, TypeDesc td, boolean isFinal = false) returns VarDeclStmt|err:Syntax {
     Token? cur = tok.current();
-    string varName = check tok.expectIdentifier();
+    string varName;
+    if cur == "_" {
+        varName = "_";
```
jclark: Modeling this as a variable named `_` is not the right approach. (We can have a variable named `_` by using `'_`.) I suggest you first in the issue discuss how to model this in the AST.

jclark (changes requested):

compiler/modules/front.syntax/parseStmt.bal
KavinduZoysa: Changed the nodes as discussed in the issue.

compiler/modules/front/codeGen.bal
```diff
@@ -771,13 +771,30 @@ function codeGenPanicStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environm
...
+        return { block: nextBlock, bindings: { name: varName, reg: result, prev: env.bindings, isFinal } };  
     }
+}
+
+function codeGenWildcardDeclStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environment env, s:Expr expr, s:TypeDesc td) returns CodeGenError|StmtEffect {
     t:SemType semType = check cx.resolveTypeDesc(td);
-    bir:Register result = cx.createRegister(semType, varName);
-    bir:BasicBlock nextBlock = check codeGenAssign(cx, env, startBlock, result, initExpr, semType);
-    return { block: nextBlock, bindings: { name: varName, reg: result, prev: env.bindings, isFinal } };  
+    if t:isEmpty(cx.mod.tc, t:intersect(semType, t:ERROR)) {
```
jclark: Just do
```
if !t:isSubtype(cx.mod.tc, t:ANY) {
    return cx.semanticErr("type descriptor of wildcard should be a subtype of any");
}
// no need for else here
```

jclark (changes requested):

compiler/modules/front/codeGen.bal
KavinduZoysa: Changed

jclark (changes requested):
Please add some tests to check this with narrowing as described here https://github.com/ballerina-platform/ballerina-spec/issues/887#issuecomment-902334830

jclark (changes requested):
The tests should be in a new directory starting with `09` e.g. 09-wildcard.

jclark (approved):
Looks good.

PR #531, opened 2021-10-06
Extend grammar to have nil typedesc

manuranga:
Part of #133

### DIFF
--- a/docs/subset09.md
+++ b/docs/subset09.md
@@ -89,7 +89,11 @@ union-type-desc =
   optional-type-desc
   | union-type-desc "|" optional-type-desc
 
-optional-type-desc = builtin-type-name [ "?" ]
+optional-type-desc = builtin-type-desc [ "?" ]
+
+builtin-type-desc = nil-type-desc | builtin-type-name
+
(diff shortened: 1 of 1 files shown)


docs/subset09.md
```diff
@@ -89,7 +89,11 @@ union-type-desc =
   optional-type-desc
   | union-type-desc "|" optional-type-desc
 
-optional-type-desc = builtin-type-name [ "?" ]
+optional-type-desc = builtin-type-desc [ "?" ]
```
manuranga, line 92: Earlier I wrote it as `optional-type-desc = (nil-type-desc | builtin-type-name) [ "?" ]` but when I am implementing this realized that I do need a name for `nil-type-desc | builtin-type-name` and might as well write the grammar the same way.

@jclark is this a good name?

manuranga: On a second thought, I can use `InlineBuiltinTypeDesc|"("` as the type in code, in that case I can inline this.

jclark:
I suggest you finish #412 first. Then after that we can figure out the best fix for the grammar.

PR #532, opened 2021-10-06
Implement nil type descriptor

### LINKED ISSUE #133: Implement nil type descriptor

manuranga:
Fixes #133
Part of #412

### DIFF
--- a/compiler/modules/front.syntax/parseStmt.bal
+++ b/compiler/modules/front.syntax/parseStmt.bal
@@ -77,7 +77,20 @@ function parseStmt(Tokenizer tok) returns Stmt|err:Syntax {
         var td if td is InlineBuiltinTypeDesc|"map" => {
             return parseVarDeclStmt(tok);
         }
-        "("|[DECIMAL_NUMBER, _]|[STRING_LITERAL, _]|"true"|"false"|"null" => {
+        "(" => {
+            TokenizerState state = tok.save();
+            check tok.advance();
+            boolean isTypeDesc = check preparseParenTypeDesc(tok);
+            tok.restore(state);
+
+            if isTypeDesc {
+                return parseVarDeclStmt(tok);
+            }
+            else {
+                return parseMethodCallStmt(tok);
+            }
+        }
+        [DECIMAL_NUMBER, _]|[STRING_LITERAL, _]|"true"|"false"|"null" => {
             return parseMethodCallStmt(tok);
         }
     }
--- a/compiler/modules/front.syntax/preparseStmt.bal
+++ b/compiler/modules/front.syntax/preparseStmt.bal
@@ -1,0 +1,41 @@
+import wso2/nballerina.err;
+
+const PREDICT_TYPE_DESC = true;
+const PREDICT_EXPR = false;
+
+
+// Returns `true` if a statement whose first token was `(` begins a type descriptor rather than an expression
+// (implying that the statement is a local variable declaration rather than a method call).
+// This is a preparse: the statement will be parsed again according to the value returned.
+// For now, this only handles a statement starting with `()`. It will be expanded later.
+function preparseParenTypeDesc(Tokenizer tok) returns boolean|err:Syntax {
+    Token? t = tok.current();
+    if t != ")" {
+        return PREDICT_EXPR;
+    }
+    check tok.advance();
+    t = tok.current();
+    match t {
+        "." => {
+            // semantically invalid
+            return PREDICT_EXPR;
+        }
+        "?" => {
+            // SUBSET with optional field access this can be EXPR
+            return PREDICT_TYPE_DESC;
+        }
+        "[" => {
+            // SUBSET should match list binding pattern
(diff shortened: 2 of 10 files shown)


compiler/modules/front.syntax/preparseStmt.bal
```diff
@@ -0,0 +1,35 @@
+import wso2/nballerina.err;
+
+const TD = true;
```
jclark: These are globally scoped, so need to be a bit longer e.g. PREPARSE_TYPE_DESC PREPARSE_EXPR

compiler/modules/front.syntax/parseStmt.bal
```diff
@@ -77,7 +77,20 @@ function parseStmt(Tokenizer tok) returns Stmt|err:Syntax {
         var td if td is InlineBuiltinTypeDesc|"map" => {
             return parseVarDeclStmt(tok);
         }
-        "("|[DECIMAL_NUMBER, _]|[STRING_LITERAL, _]|"true"|"false"|"null" => {
+        "(" => {
+            TokenizerState state = tok.save();
+            check tok.advance();
+            boolean isTd = check preparseParenTypeDesc(tok);
```
jclark: I prefer isTypeDesc to isTd here.

compiler/modules/front.syntax/preparseStmt.bal
```diff
@@ -0,0 +1,35 @@
...
+const EXPR = false;
+
+function preparseParenTypeDesc(Tokenizer tok) returns boolean|err:Syntax {
```
jclark, line 11: This needs a comment to explain what's going on.

compiler/modules/front.syntax/preparseStmt.bal
```diff
@@ -0,0 +1,35 @@
...
+    Token? t = tok.current();
+    if t != ")" {
+        return EXPR;
+    }
+    check tok.advance();
+    t = tok.current();
+    if t == "." {
```
jclark: This is better done as a `match` statement.

compiler/modules/front.syntax/preparseStmt.bal
```diff
@@ -0,0 +1,35 @@
...
+        // semantically invalid
+        return EXPR;
+    }
+    else if t == "?" {
+        // SUBSET with optional field access this can be EXPR
+        return TD;
+    }
+    else if t == "[" {
+        check tok.advance();
```
jclark: SUBSET this can also be the start of a binding pattern (semantically invalid)

compiler/modules/front.syntax/preparseStmt.bal
```diff
@@ -0,0 +1,35 @@
...
+        }
+        else {
+            // SUBSET will need to recurse when T[n] td is supported
+            return EXPR;
+        }
+    }
+    else if t is [IDENTIFIER, string] {
+        return TD;
+    }
+    return tok.err("invalid statement");
```
jclark: I would be inclined to return EXPR here and leave it to the regular statement parsing return an error. Then you can leave out `else { return EXPR; }` above.  So the function's contract is to return true if it maybe a valid type descriptor, false otherwise, and returns an error if it cannot get sufficient tokens to decide.

compiler/modules/front.syntax/preparseStmt.bal
jclark: Returns true if a statement whose first token was `(` begins a type descriptor rather than an expression (implying that the statement is a local variable declaration rather than a method call). This is a preparse: the statement will be parsed again according to the value returned. For now, this only handles a statement starting with `()`. It will be expanded later.

jclark (changes requested):

compiler/modules/front.syntax/preparseStmt.bal
```diff
@@ -0,0 +1,41 @@
+import wso2/nballerina.err;
+
+const PREDICT_TYPE_DESC = true;
```
jclark, line 3: The name PREDICT_x is just not right. It doesn't fit with preparse. We are saying it can only be a expression or type descriptor, although it may not be a valid one: and that's not really what prediction means.

jclark (commented):
Please change to something else in #412.

compiler/modules/front.syntax/preparseStmt.bal
manuranga: I'll replace with PREPARSE_x in next pr.

manuranga: (I don't get your objection though. Does predict gives the idea that it can be the other way some times?)
Other options : signal, cue

jclark: predict is saying something about the future and implies uncertainty. Here we are not guessing how we are going to parse it, we are saying with certainty how it must be parsed.

PR #534, opened 2021-10-07
Add projection semtype testing

### LINKED ISSUE #499: Improve semtypes testing
Need to improve both tests and test framework.

ushirask:
## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/front.syntax/parseTypeTest.bal
```diff
@@ -0,0 +1,52 @@
...
+        string right = check tok.expectIdentifier();
+        return { op, left, right };
+    } 
+    else if tok.current() is "[" {
+        check tok.advance();
+        string index = check tok.expectIdentifier();
+        if !(tok.current() is "]") {
+            return parseError(tok);
+        }
+        check tok.advance();
```
manuranga: `tok.expect("]")`

compiler/modules/front.syntax/parseTypeTest.bal
```diff
@@ -0,0 +1,52 @@
...
+        }
+        else {
+            op = t;
+        }
+        string right = check tok.expectIdentifier();
+        return { op, left, right };
+    } 
+    else if tok.current() is "[" {
+        check tok.advance();
+        string index = check tok.expectIdentifier();
```
manuranga: I am not sure if you were planing to do this as separate PR, but we want this to be either Identifier or an int literal.(we'll have to add string literal as well, but we can do that later)

ushirask: I'll be adding that in the same PR

compiler/testSuite/01-boolean/boolean-t.bal
```diff
@@ -0,0 +1,23 @@
+// @type T < TF
+// @type TF = B
+// @type F < B
+// @type INTEGER <> B
+// @type BL[ONE] = B
```
jclark: It's better to put the `@type` lines just before the types they use (like we do with `@output` lines)

compiler/modules/front/front.bal
```diff
@@ -256,5 +256,14 @@ public function typesFromString(SourcePart[] sourceParts) returns [t:Env, map<t:
         check addModulePart(syms.defns, check s:parseModulePart(part));
     }
     check resolveTypes(syms);
-    return [env, createTypeMap(syms)];
+    map<string> values = {};
+    foreach var v in syms.defns {
+        if v is s:ConstDefn && v.expr is s:ConstValueExpr && (<s:ConstValueExpr> v.expr).value is string {
```
jclark: Too many casts. Split off some helper functions and introduce some temporary variables.

compiler/tests/testSuite.bal
```diff
@@ -164,3 +177,57 @@ function testCompileFile(string filename) returns CompileError? {
...
+                    test:assertFalse(t:isSubtype(tc, t2, t1), test.right + " is a subtype of " + test.left);
+                }
+                "=" => {
+                    test:assertTrue(t:isSubtype(tc, t1, t2), test.left + " is not a subtype of " + test.right);
+                    test:assertTrue(t:isSubtype(tc, t2, t1), test.right + " is not a subtype of " + test.left);
+                }
+            }
+        }
+        else {
+            t:SemType t2 = m.entries().get(test.member)[1];
```
jclark: Introduce temp variables to avoid all these casts.

compiler/tests/testSuite.bal
```diff
@@ -164,3 +177,57 @@ function testCompileFile(string filename) returns CompileError? {
     var [basename, _] = basenameExtension(filename);
     return compileBalFile(filename, basename, (), {}, {});
 }
+
+function testSubtypes(front:SourcePart[] sources, string[] expected) returns error? {
+    var [env, m, v] = check front:typesFromString(sources);
+    var tc = t:typeContext(env);
+    foreach var item in expected {
+        s:TypeTest test = check s:parseTypeTest(item);
+        t:SemType t1 = m.entries().get(test.left)[1];
```
jclark: I'm not sure what you're doing here, but you shouldn't be repeatedly calling m.entries().

jclark (changes requested):

compiler/tests/testSuite.bal
ushirask: each item in `expected` is a single `@test` and I'm getting the semtypes of the types in that test from the semtype map, can this be avoided?

ushirask: or do you mean that `m.entries()` should be kept outside the loop with just calling the `get` repeatedly

manuranga: No, the qustion is why there is `entries` at all. Isn't `m.entries().get(test.left)[1];` same as `m.get(test.left)` ?

jclark: The normal way to use a map is `m[k]`: that gives you a nil if the map doesn't contain an entry for k. If you somehow know for sure that it does, then you can use `m.get(k)`.  The normal use case for entries is something like

```
foreach var [k, v] in m.entries() {
   // do something with the key and value
}
```

compiler/testSuite/01-boolean/boolean-t.bal
```diff
@@ -0,0 +1,26 @@
+// @type T < TF
```
manuranga: When ever possible, try to have the types referred in an assertion before the assertion.

compiler/tests/testSuite.bal
```diff
@@ -164,3 +177,76 @@ function testCompileFile(string filename) returns CompileError? {
...
+                test:assertFail(test.member + " is not declared");
+            }
+            t:SemType t2 = m.get(test.member);
+            string? index = test.index;
+            if t:isSubtype(tc, t1, t:LIST) {
+                if index is string {
+                    if !(m.hasKey(index)) {
+                        test:assertFail(index + " is not declared");
+                    }
+                    t:SemType key = m.get(index);
```
manuranga: Don't lookup twice, `t:SemType? key = m[index];` then check if it's nil.

compiler/tests/testSuite.bal
```diff
@@ -164,3 +177,81 @@ function testCompileFile(string filename) returns CompileError? {
...
+            }
+            t:SemType t2 = <t:SemType> tn2;
+            string? index = test.index;
+            if t:isSubtype(tc, t1, t:LIST) {
+                if index is string {
+                    t:SemType? keyn = m[index];
+                    if keyn is null {
+                        test:assertFail(index + " is not declared");
+                    }
+                    t:SemType key = <t:SemType> keyn;
```
manuranga: We'll not need the cast from jBallerina beta 4, but until then lest write this in a `else` case to avoid the cast.

ushirask: should this be done only for this `key` semtype since for other types where a large portion of the code will have to be moved into an else block

manuranga: It'll have to be the whole thing, I agree it's getting too indented, can we move to a function, and maybe even reused it for below section?

(btw please don't use `null` use `()`)

compiler/modules/front.syntax/parseTypeTest.bal
```diff
@@ -0,0 +1,51 @@
...
+            check tok.advance();
+            op = "<>";
+        }
+        else {
+            op = t;
+        }
+        string right = check tok.expectIdentifier();
+        return { op, left, right };
+    } 
+    check tok.expect("[");
```
manuranga: Don't use expect here, this should be an `else` case of the above `if`.
Otherwise then someone writes a wrong assertion (eg `T > S`) it will say `[ expected found >`, which is wrong.

compiler/modules/front.syntax/parseTypeTest.bal
```diff
@@ -0,0 +1,51 @@
...
+    string? index = ();
```
manuranga: Make this `string? index;` and move `index = ();` to `if`.
General idea is to minimize the amount of mutable local variables.

compiler/modules/front.syntax/parseTypeTest.bal
```diff
@@ -0,0 +1,51 @@
+public function parseTypeTest(string str) returns TypeTest|error {
```
manuranga, line 16: This has to be broken up to two functions.
We need to support `T[1] < S` as well, so do the parsing of `T[1]` and `S` in one, and call it form here.
```
type-expr := identifier ("[" number "]")?
type-assertion := type-expr ("<"|"="|"<>") type-expr 
```

compiler/modules/front.syntax/parseTypeTest.bal
```diff
@@ -0,0 +1,65 @@
...
+        } 
+        else if op is "=" {
+            return {left, member : right.identifier, index : left.index};
+        }
+    }
+    return parseError(tok);
+    
+}
+
+function parseTypeExpr(Tokenizer tok) returns Operand|error {
```
manuranga: let's call this `parseTypeProjection`

ushirask: would that be appropriate since this handles the parsing of both projection and non types?

manuranga: I think so, since we defining a min grammar where this is parsing a rule like `type-projection := identifier ("[" ( number | identifier ) "]")?`.

compiler/modules/front/front.bal
```diff
@@ -256,5 +256,24 @@ public function typesFromString(SourcePart[] sourceParts) returns [t:Env, map<t:
...
+        } 
+        else if v is s:TypeDefn {
+            s:TypeDesc td = v.td;
+            if td is s:SingletonTypeDesc && td.value is string {
+                values[v.name] = <string> td.value;
+            }
+        }
+    }
+    return values;
+}
```
manuranga: is this used ?

ushirask: it's used for getting key value for `mappingMemberType` called in testSuite

compiler/modules/front.syntax/parseTypeTest.bal
```diff
@@ -0,0 +1,65 @@
...
+        check tok.expect("]");
+    }
+    return {identifier, index};
+}
+
+public type SubtypeTest record {
+    SubtypeTestOp op;
+    string left;
+    string right;
+};
```
manuranga:
```suggestion
public type SubtypeTest record {
    SubtypeTestOp op;
    string|TypeProjection left;
    string|TypeProjection right;
};
```
`parseTypeProjection` should return `string|TypeProjection`.
You will also have to change other types and testing code accordingly.

compiler/modules/front.syntax/parseTypeTest.bal
```diff
@@ -0,0 +1,59 @@
...
+        else {
+            return parseError(tok);
+        }
+        check tok.expect("]");
+        return {identifier, index};
+    }
+    return identifier;
+}
+
+public type TypeTest record {
```
manuranga: Try to arrange files such that,  things are defined before they are referred. So have these at the top of the file.

ushirask: fixed

compiler/modules/front.syntax/parseTypeTest.bal
```diff
@@ -0,0 +1,58 @@
...
+        string|TypeProjection right = check parseTypeProjection(tok);
+        return { op, left, right };
+    }
+    return parseError(tok);
+    
+}
+
+function parseTypeProjection(Tokenizer tok) returns string|TypeProjection|error {
+    string identifier = check tok.expectIdentifier();
+    int|string index;
```
manuranga: Should be moved inside `if`, just before the second `if`.

compiler/tests/testSuite.bal
```diff
@@ -164,3 +177,64 @@ function testCompileFile(string filename) returns CompileError? {
...
     return compileBalFile(filename, basename, (), {}, {});
 }
+
+function testSubtypes(front:SourcePart[] sources, string[] expected) returns error? {
+    var [env, m] = check front:typesFromString(sources);
+    var tc = t:typeContext(env);
+    foreach var item in expected {
+        s:TypeTest test = check s:parseTypeTest(item);
+        t:SemType t1 = resolveTestSemtype(tc, m, test.left);
+        t:SemType t2 = resolveTestSemtype(tc, m, test.right);
```
manuranga: I slightly prefer to call these two, `left` and `right`.

compiler/tests/testSuite.bal
```diff
@@ -164,3 +177,64 @@ function testCompileFile(string filename) returns CompileError? {
...
+                    return t:mappingMemberType(tc, t, index);
+                }
+                test:assertFail("index for mapping must be a string");
+            }
+            else {
+                test:assertFail(tn.identifier + " is not a list or a mapping type");
+            } 
+        }
+    }
+    return error("cannot resolve semtype");
```
manuranga: Isn't this line unreachable ? lets change the message to say, `"unreachable"`.
I think this can be fixed from jBallerina side. Please add JBUG comment referring to https://github.com/ballerina-platform/ballerina-lang/issues/31642

compiler/modules/front.syntax/tests/testParserOnTestSuite.bal
```diff
@@ -16,6 +16,9 @@ function testParserOnTestSuite() returns err:Syntax|io:Error|file:Error? {
                 continue;
             }
             string filename = test.absPath;
+            if filename.endsWith("-t.bal") {
+                continue;
```
manuranga, line 21: Does it fail for `t` cases ? I wonder why? perhaps we should fix.

ushirask: it fails with 
`something impossible happened (typedesc not implemented in typeDescToWords: '{"value":true}')`

ushirask: I'll try adding needed TypeDesc handing to typeDescToWords

compiler/tests/testSuite.bal
```diff
@@ -164,3 +177,65 @@ function testCompileFile(string filename) returns CompileError? {
...
+                }
+                test:assertFail("index for mapping must be a string");
+            }
+            else {
+                test:assertFail(tn.identifier + " is not a list or a mapping type");
+            } 
+        }
+    }
+    // JBUG: #31642 function must return a call
+    return error("unreachable");
```
manuranga: Sorry I should have said, `panic error("unreachable");`.
(BTW how does this even compile ? return type  is mismatched)

compiler/modules/front.syntax/tests/testParserOnTestSuite.bal
manuranga: hmm, unless it's a very small change let's do it in a separate PR, create an issue, and add comment referring to that issue here.

compiler/tests/testSuite.bal
ushirask: I was confused as well originally I returned  SemType|error and used a check however it worked even without

manuranga: It's very concerning if this compiles, lets create an issue in jBallerina if it does.

ushirask: created in https://github.com/ballerina-platform/ballerina-lang/issues/33175

compiler/modules/front.syntax/tests/testParserOnTestSuite.bal
ushirask: created issue in #555

manuranga: Please add a `xxx` comment here.

compiler/modules/front.syntax/tests/testParserOnTestSuite.bal
```diff
@@ -16,6 +16,10 @@ function testParserOnTestSuite() returns err:Syntax|io:Error|file:Error? {
                 continue;
             }
             string filename = test.absPath;
+            // XXX need to handle type test cases in typeDescToWords
```
manuranga: sorry I meant `xxx`

manuranga: `// xxx remove after fixing #555`

compiler/testSuite/01-boolean/boolean-t.bal
```diff
@@ -0,0 +1,24 @@
...
+type TF true|false;
+
+// @type INTEGER <> B
+type B boolean;
+type INTEGER int;
+type S string;
+
+// @type BL[1] = B
+// @type BL[2] = B
+type BL boolean[];
```
manuranga:
```suggestion
type B boolean;

// @type TF = B
type TF true|false;

// @type T < TF
type T true;

// @type F < B
type F false;

// @type INTEGER <> B
type INTEGER int;

// @type BL[1] = B
// @type BL[2] = B
type BL boolean[]; 
```

compiler/tests/testSuite.bal
```diff
@@ -164,3 +177,65 @@ function testCompileFile(string filename) returns CompileError? {
...
+        }
+    }
+    else {
+        t:SemType? t = m[tn.identifier];
+        if t is () {
+            test:assertFail(tn.identifier + " is not declared");
+        }
+        else {
+            int|string index = tn.index;
+            if t:isSubtype(tc, t, t:LIST) {
```
manuranga: I think `t:isSubtypeSimple(t, t:LIST)` is better.

manuranga: same for mapping

manuranga:
@jclark I think this is ready to be reviewed.

Please note that we only support literals in projection.

```
const R = 1;

type S [int, boolean, string];
type T 1|2;
type U boolean|string;
type V boolean;

// Below is useful to test, but I don't think SemTypes support it yet, does it?
// @type S[T] = U 

// This is not that useful, one can use literals instead.
// @type S[R] = V 
```

compiler/modules/front.syntax/parseTypeTest.bal
```diff
@@ -0,0 +1,58 @@
...
+        }
+        else if t is [STRING_LITERAL, string] {
+            index = t[1];
+            check tok.advance();
+        }
+        else {
+            return parseError(tok);
+        }
+        check tok.expect("]");
+        return {identifier, index};
```
manuranga: missing spaces.

compiler/modules/front.syntax/tests/testParserOnTestSuite.bal
```diff
@@ -16,6 +16,10 @@ function testParserOnTestSuite() returns err:Syntax|io:Error|file:Error? {
                 continue;
             }
             string filename = test.absPath;
+            // xxx remove after fixing #555
```
manuranga: Sorry, `XXX` was the correct thing.

manuranga: Sorry, `XXX` was the correct thing.

manuranga:
Answering my own question :
```
// Below is useful to test, but I don't think SemTypes support it yet, does it?
// @type S[T] = U 
```
It is not supported for any SemType `T` but it is supported when T is `int` for lists and `T` is `string` for mappings.

Opt1 : Support `// @type S[T] = U`, we'll look up `T` if check if it's `int`/`string` and pass `()` to corresponding projection function.

Opt2: Support `// @type S[string] = U`


@ushirask Shall we try above Opt1.

jclark:
@manuranga I haven't implemented that yet: we'll need this when we add support for singletons. Opt1 sounds good to me.

compiler/showTypes.bal
```diff
@@ -3,6 +3,7 @@ import ballerina/io;
 import wso2/nballerina.front;
 import wso2/nballerina.err;
 import wso2/nballerina.types as t;
+
```
jclark: Why is this changed?

compiler/testSuite/01-boolean/boolean-t.bal
```diff
@@ -0,0 +1,33 @@
+type B boolean;
```
jclark: This doesn't belong in 01-boolean.  A directory NN-foo is for subset NN.

jclark:
> This doesn't belong in 01-boolean.  A directory NN-foo is for subset NN. 

I suggest semtype/proj1-t.bal

compiler/tests/testSuite.bal
```diff
@@ -164,3 +177,86 @@ function testCompileFile(string filename) returns CompileError? {
...
 }
+
+function testSubtypes(front:SourcePart[] sources, string[] expected) returns error? {
+    var [env, m] = check front:typesFromString(sources);
+    var tc = t:typeContext(env);
+    foreach var item in expected {
+        s:TypeTest test = check s:parseTypeTest(item);
+        t:SemType left = resolveTestSemtype(tc, m, test.left);
+        t:SemType right = resolveTestSemtype(tc, m, test.right);
+        match test.op {
```
jclark, line 192: You can do this much better.  Do t:isSubtype(tc, left, right) and vice-versa once. Make it into a pair of booleans. Then check that the pair of booleans has the right result.

jclark (changes requested):

compiler/modules/front.syntax/parseTypeTest.bal
```diff
@@ -0,0 +1,61 @@
+public type TypeProjection record {
+    string identifier;
+    int|string index;
+    boolean literal;
```
manuranga: Please remove the boolean, Lets make the identifiers separate. `int|string|Identifier index;`. this means we'll need something like `type Identifier [string];`.

compiler/tests/testSuite.bal
```diff
@@ -164,3 +177,86 @@ function testCompileFile(string filename) returns CompileError? {
...
+                if index is int {
+                    return t:listMemberType(tc, t, index);
+                    //return t:listProj(tc, t, index);
+                }
+                else if !tn.literal {
+                    t:SemType? k = m[index];
+                    if k is () {
+                        test:assertFail(tn.identifier + " is not declared");
+                    }
+                    else if t:isSubtype(tc, k, t:INT){
```
manuranga: I think this should be `k == t:INT`

compiler/tests/testSuite.bal
```diff
@@ -164,3 +177,86 @@ function testCompileFile(string filename) returns CompileError? {
...
+                        return t:listMemberType(tc, t, ());
+                    }
```
manuranga: else we need to say we currently only support projections on int literals or the whole int type.

compiler/showTypes.bal
ushirask: It shouldn't be; fixed.

compiler/modules/front.syntax/parseTypeTest.bal
```diff
@@ -0,0 +1,60 @@
...
+
+public type TypeTest record {
+    SubtypeTestOp op;
+    Identifier|TypeProjection left;
+    Identifier|TypeProjection right;
+};
+
+public type SubtypeTestOp "<" | "=" | "<>";
+
+public type Identifier [string];
```
manuranga: I don't get why James is not too keen on using `[string]` since that is essentially what `token.bal` is doing with `VariableLengthToken`, anyway let's stick to `public type Identifier string;` for now.

compiler/tests/testSuite.bal
```diff
@@ -164,3 +177,86 @@ function testCompileFile(string filename) returns CompileError? {
...
+        if t:isSubtypeSimple(t, t:LIST) {
+            if index is int {
+                return t:listMemberType(tc, t, index);
+                //return t:listProj(tc, t, index);
+            }
+            else if index is s:Identifier {
+                t:SemType k = lookupSemtype(m, index);
+                if k == t:INT {
+                    return t:listMemberType(tc, t, ());
+                }
```
manuranga, line 222: we need a `else` case that calls `singleShape(k)` and check if it `is int`.

compiler/modules/front.syntax/parseTypeTest.bal
```diff
@@ -0,0 +1,56 @@
+public type TypeProjection record {
+    Identifier identifier;
+    int|string|Identifier index;
```
manuranga:
```suggestion
    int|Identifier index;
```

jclark (approved):

PR #540, opened 2021-10-08
Disallow underscore for org-name and module-name in import-decl

### LINKED ISSUE #444: Restrict underscores in import organization/module names
As described in https://github.com/ballerina-platform/ballerina-spec/issues/791

KavinduZoysa:
## Purpose
Fixes #444

### DIFF
--- a/compiler/modules/front.syntax/parse.bal
+++ b/compiler/modules/front.syntax/parse.bal
@@ -75,14 +75,14 @@ function parseImportDecl(Tokenizer tok, int partIndex) returns ImportDecl|err:Sy
     Position startPos = tok.currentStartPos();
     check tok.advance();
     Position namePos = tok.currentStartPos();
-    string firstModuleName = check tok.expectIdentifier();
+    string firstModuleName = check tok.expectImportPart();
     string? org = ();
     if tok.current() == "/" {
         // we have an org
         org = firstModuleName;
         check tok.advance();
         namePos = tok.currentStartPos();
-        firstModuleName = check tok.expectIdentifier();
+        firstModuleName = check tok.expectImportPart();
     }
     [string, string...] names = [firstModuleName];
     names.push(...check parseImportNamesRest(tok));
@@ -96,7 +96,7 @@ function parseImportNamesRest(Tokenizer tok) returns string[]|err:Syntax {
     string[] names = [];
     while tok.current() == "." {
         check tok.advance();
-        names.push(check tok.expectIdentifier());
+        names.push(check tok.expectImportPart());
     }
     return names;
 }
--- a/compiler/modules/front.syntax/parse.bal
+++ b/compiler/modules/front.syntax/parse.bal
@@ -107,15 +107,15 @@ function validImportPart(Tokenizer tok) returns string|err:Syntax {
     foreach var ch in identifier {
         if ch == "_" {
             if prevChar == () {
-                return tok.err("expect an identifier without leading underscores");
+                return tok.err("identifier in an import must not have leading underscores");
             } else if prevChar == "_" {
-                return tok.err("expect an identifier without consecutive underscores");
+                return tok.err("identifier in an import must not have consecutive underscores");
             }
         }
         prevChar = ch;
     }
     if prevChar == "_" {
-        return tok.err("expect an identifier without tailing underscores");
+        return tok.err("identifier in an import must not have tailing underscores");
     }
     return identifier;
 }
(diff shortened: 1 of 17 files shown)


compiler/modules/front.syntax/token.bal
```diff
@@ -286,6 +286,16 @@ class Tokenizer {
         return true;
     }
 
+    function expectImportPart() returns string|err:Syntax {
```
jclark: This should not be a method. Furthermore, the rule is not the identifier cannot contain identifiers.

runtime/error.c
```diff
@@ -362,10 +362,10 @@ static bool demangleCountedName(const char **pp, const char **name) {
     }
     const char *p = *name;
     while (nChars > 0) {
-        p++;
-        if (*p == '\0') {
+        if (*p == '_' || *p == '\0') {
```
jclark: This is not correct.

runtime/tests/demangle.c
```diff
@@ -20,24 +20,22 @@ static const DemangleTestCase tests[] = {
     { "_Bb1x6choreohello", "hello (ballerinax/choreo)" },
     { "_Bb0m4lang5valuetoString", "toString (ballerina/lang.value)" },
     { "_B4wso2m10nballerinam5front6syntaxparse", "parse (wso2/nballerina.front.syntax)" },
-    { "_B0m7foo_bar3bazhello", "hello (foo_bar.baz)" },
```
jclark: Not correct

jclark (changes requested):

runtime/tests/demangle.c
KavinduZoysa: @jclark, since we disallow underscores in module name, `_B0m7foo_bar3bazhello` should not be demangled. So the output should be `_B0m7foo_bar3bazhello`.
Is this output incorrect or test case invalid?

KavinduZoysa: Please ignore the above comment.

compiler/modules/front.syntax/token.bal
KavinduZoysa: Changed

runtime/error.c

runtime/tests/demangle.c
KavinduZoysa: Changed

runtime/error.c
```diff
@@ -361,11 +361,20 @@ static bool demangleCountedName(const char **pp, const char **name) {
         *name = *name + 1;
     }
     const char *p = *name;
+    long tem = nChars;
     while (nChars > 0) {
-        p++;
         if (*p == '\0') {
             return false;
         }
+        if (*p == '_') {
```
jclark: Don't make this change. We are not validating other characters. Apply Postel's law.

compiler/modules/front.syntax/parse.bal
```diff
@@ -96,11 +96,24 @@ function parseImportNamesRest(Tokenizer tok) returns string[]|err:Syntax {
     string[] names = [];
     while tok.current() == "." {
         check tok.advance();
-        names.push(check tok.expectIdentifier());
+        names.push(check validImportPart(tok));
     }
     return names;
 }
 
+function validImportPart(Tokenizer tok) returns string|err:Syntax {
```
jclark, line 104: You are not checking for a leading `_`. Right way to write this function is to have a local variable `string? prevChar` and then iterate over each character (`foreach var ch in identifier`); if it's an underscore, then check whether prevChar is () or `_`. At the end check for prevChar being `_`.

jclark (changes requested):

runtime/error.c
KavinduZoysa: Reverted runtime change

compiler/modules/front.syntax/parse.bal
KavinduZoysa: Fixed

compiler/modules/front.syntax/parse.bal
```diff
@@ -96,11 +96,30 @@ function parseImportNamesRest(Tokenizer tok) returns string[]|err:Syntax {
...
+    string identifier = check tok.expectIdentifier();
+    string? prevChar = ();
+    foreach var ch in identifier {
+        if ch == "_" {
+            if prevChar == () {
+                return tok.err("expect an identifier without leading underscores");
```
jclark: In all of the three messages instead "expect an identifier without" say "identifier in an import must not have"

jclark (changes requested):

compiler/modules/front.syntax/parse.bal
KavinduZoysa: Changed

PR #543, opened 2021-10-10
Allow statements to use full type descriptors

### LINKED ISSUE #412: Allow statements to use full type descriptors

manuranga:
Fixes #412

### DIFF
--- a/compiler/modules/front.syntax/preparseStmt.bal
+++ b/compiler/modules/front.syntax/preparseStmt.bal
@@ -1,41 +1,38 @@
 import wso2/nballerina.err;
 
-const PREDICT_TYPE_DESC = true;
-const PREDICT_EXPR = false;
-
+const PREPARSE_TYPE_DESC = true;
+const PREPARSE_EXPR = false;
 
 // Returns `true` if a statement whose first token was `(` begins a type descriptor rather than an expression
 // (implying that the statement is a local variable declaration rather than a method call).
 // This is a preparse: the statement will be parsed again according to the value returned.
-// For now, this only handles a statement starting with `()`. It will be expanded later.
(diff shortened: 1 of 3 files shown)


compiler/modules/front.syntax/preparseStmt.bal
```diff
@@ -1,41 +1,38 @@
...
-    check tok.advance();
-    t = tok.current();
-    match t {
-        "." => {
-            // semantically invalid
-            return PREDICT_EXPR;
+    return t != () && t != ".";
+}
+
+function skipToClose(Tokenizer tok, string? close) returns err:Syntax? {
```
jclark, line 20: I don't think the names are right. This function is trying to parsing a parenthesized (or otherwise bracketed) type descriptor.  The above function is trying to discriminate between a local variable declaration and a method call.

compiler/modules/front.syntax/preparseStmt.bal
```diff
@@ -1,41 +1,38 @@
...
+    while true {
+        Token? t = tok.current();
+        if t == () || t == ";" && close != "|}" {
```
jclark, line 23: Use parentheses to make precedence explicit here.

In the case of `;`, I think you should return an error. This function is declared as returning an error, but doesn't do so.

compiler/modules/front.syntax/preparseStmt.bal
jclark: When you add complexity, this function will need to return `err:Syntax|boolean?`, where the boolean case means you have seen something that tells you immediately that what kind of thing it is. I would add one of these cases here just to show the structure we will need.

jclark (approved):
I will merge and take a crack at improving it.

PR #547, opened 2021-10-11
Remove parseInlineTypeDesc with the help of peek

### LINKED ISSUE #412: Allow statements to use full type descriptors

manuranga:
Fixes #412


compiler/modules/front.syntax/parseStmt.bal
```diff
@@ -164,9 +164,10 @@ function finishOptQualIdentifierStmt(Tokenizer tok, string? prefix, string ident
 // Parse a statement that starts with the keyword `error`
 function parseErrorStmt(Tokenizer tok) returns Stmt|err:Syntax {
     Position pos = tok.currentStartPos();
-    check tok.advance();
-    if tok.current() == "(" {
-        check tok.advance();
+    VariableTokenCode|FixedToken? t = tok.peek();
```
jclark: Don't see any point to a temp variable here.

compiler/modules/front.syntax/parseStmt.bal
```diff
@@ -164,9 +164,10 @@ function finishOptQualIdentifierStmt(Tokenizer tok, string? prefix, string ident
...
+    if t == "(" {
+        check tok.advance(); // advance `error`
+        check tok.advance(); // advance `(`
         ErrorConstructorExpr errExpr = check finishErrorConstructorExpr(tok, pos);
```
jclark: I think we can simplify further further.  Don't call advance and instead call parseMethodCallStmt. Don't need to call currentStartPos.  Merge parseErrorStmt into parseStmt. Then merge finishErrorConstructExpr into parseExpr.

jclark (changes requested):

Issue #548, opened 2021-10-11
Migrate to jBallerina swan lake beta3


manuranga:
Migrate to jBallerina swan lake beta3, available at https://github.com/ballerina-platform/ballerina-distribution/releases/tag/vswan-lake-beta3

manuranga:
Word `linux` is dropped form the zip file name for some reason, but not the `windows`, need update OS picking for this.

jclark:
When it's available on ballerina.io, the documentation should be updated to require it.

We should also remove JBUGs that are not longer necessary.

manuranga:
jre has been removed form the linux zip, I'll have to use deb file.

manuranga:
I think with beta 3, we are expected to commit the `Dependencies.toml` file.

jclark:
Why do we need to commit `Dependencies.toml`? What will happen if we don't?

manuranga:
It will just create a new one on `bal build`, so everyone will have their own copy of it. But as far as I understand, the purpose of this file is to be committed so there will be a single copy of it specifying the exact dependencies for the project, at any given time (commit).

jclark:
So effectively it would be locking our dependencies? I see no need for us to do this at this point.

jclark:
This issue can wait until beta3 is up on ballerina.io and available via `bal dist pull`.

manuranga:
I started this since #547 was failing due to method too large on beta 2 which is fixed in beta 3. I'll split the method for now.

manuranga:
Updated the pr #552. 
What's left to do is to update the nightly build and to revert the large method `testParser`.
I don't think documentation needs any updating since it says to uses the latest form the ballerina.io.

sameerajayasoma:
@manuranga @jclark It is not mandatory to commit the generated `Dependencies.toml` file. Also, the existence of this file does not mean that the dependency versions are locked. 

If you run the `bal build` with `--sticky` option, then we use the `Dependencies.toml` as a lock file, otherwise, we use this file as an aid to perform compatible version updates. 

We generally recommend committing the `Dependencies.toml` file, because it stops you from getting incompatible version updates.

jclark:
@sameerajayasoma Thanks for the explanation. At this stage, I prefer to get incompatible version updates as soon as possible, so we can fix our code. (Plus we have very few external dependencies, so this is unlikely to be much of a problem for us.) When we get further along, we can commit `Dependencies.toml`.

[...]

PR #552, opened 2021-10-11
Migrate GitHub actions to jBallerina swan lake beta3

manuranga:
Part of #548

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -9,12 +9,12 @@ on:
   
 jobs:
   test:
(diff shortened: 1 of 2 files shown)


compiler/Dependencies.toml
```diff
@@ -0,0 +1,125 @@
+# AUTO-GENERATED FILE. DO NOT MODIFY.
```
jclark: I am not yet convinced we need to check this in.

jclark (commented):
Looks good to me

Issue #510, opened 2021-09-28
Implement field access lvalues


jclark:
This is x.y.z on the LHS of an assignment or compound assignment statement.

https://ballerina.io/spec/lang/2021R1/#field-access-lvexpr

manuranga:
@jclark
​​In the current subset, record fields can only be (unions of) basic types. So valid filed access lvalues will only be in the form `x.y` not `x.y.z`. Therefore I should keep `FieldAccessLExpr` as is, instead generalizing `mapping`. Is this understanding correct?

jclark:
Correct.  We will generalize lvalues when we support arbitrary member types for structures.

PR #556, opened 2021-10-12
Implement field access expressions

### LINKED ISSUE #509: Implement field access expressions
This is an expression `E.x` where the static type of E is a record type.

https://ballerina.io/spec/lang/2021R1/#field-access-expr

manuranga:
Fixes #509

### DIFF
--- a/compiler/modules/front.syntax/parseStmt.bal
+++ b/compiler/modules/front.syntax/parseStmt.bal
@@ -154,8 +154,12 @@ function finishOptQualIdentifierStmt(Tokenizer tok, string? prefix, string ident
     }
     else if cur == "." {
         VarRefExpr varRef = { varName: identifier, prefix };
-        MethodCallExpr expr = check finishMethodCallExpr(tok, varRef);
-        return finishCallStmt(tok, expr);
+        check tok.advance();
+        string name = check tok.expectIdentifier();
+        if tok.current() == "(" {
+            return finishCallStmt(tok, check finishMethodCallExpr(tok, varRef, name, pos));
+        }
+        // falls through to end
     }
     else if cur is [IDENTIFIER, string] {
         TypeDescRef ref = { prefix, typeName: identifier, pos };
--- a/compiler/modules/front/codeGen.bal
+++ b/compiler/modules/front/codeGen.bal
@@ -907,18 +907,27 @@ function codeGenCompoundAssignToListMember(CodeGenContext cx, bir:BasicBlock bb,
 
 function codeGenCompoundAssignToMappingMember(CodeGenContext cx, bir:BasicBlock bb, Environment env,
                                               s:MemberAccessLExpr lValue, bir:Register mapping, s:Expr rexpr, s:BinaryArithmeticOp|s:BinaryBitwiseOp op, err:Position pos) returns CodeGenError|StmtEffect {
-    var { result: k, block: nextBlock } = check codeGenExprForString(cx, bb, env, check cx.foldExpr(env, lValue.index, t:STRING));
-    t:SemType memberType = t:mappingMemberType(cx.mod.tc, mapping.semType, k is string ? k : ());
-    if t:isEmpty(cx.mod.tc, memberType) {
-        return cx.semanticErr("type of member access is never");
-    }
-    bir:Register member = cx.createRegister(memberType);
-    bir:MappingGetInsn getInsn = { result: member, operands: [mapping, k] };
-    nextBlock.insns.push(getInsn);
-    var { result, block } = check codeGenCompoundableBinaryExpr(cx, nextBlock, env, op, member, rexpr, pos);
+    var { result: k, block: block1 } = check codeGenExprForString(cx, bb, env, check cx.foldExpr(env, lValue.index, t:STRING));
+    var { result: member, block: block2 } = check codeGenMappingGet(cx, block1, mapping, false, k, pos);
+    var { result, block } = check codeGenCompoundableBinaryExpr(cx, block2, env, op, member, rexpr, pos);
     bir:MappingSetInsn setInsn = { operands:[ mapping, k, result], position: lValue.pos };
     block.insns.push(setInsn);
     return { block };
+}
+
+// if keyMustRequired is true, k must be a string
+function codeGenMappingGet(CodeGenContext cx, bir:BasicBlock block, bir:Register mapping, boolean keyMustRequired , bir:StringOperand k, err:Position pos) returns CodeGenError|RegExprEffect {
+    t:SemType memberType = t:mappingMemberType(cx.mod.tc, mapping.semType, k is string ? k : ());
+    if !(k is string) || !t:mappingMemberRequired(cx.mod.tc, mapping.semType, k) {
+        if keyMustRequired == true {
+            return cx.semanticErr(`${<string>k} must be a required key`, pos=pos);
+        }
+        memberType = t:union(memberType, t:NIL);
+    }
+    bir:Register result = cx.createRegister(memberType);
+    bir:MappingGetInsn insn = { result, operands: [mapping, k] };
+    block.insns.push(insn);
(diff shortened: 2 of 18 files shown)


compiler/modules/front/codeGen.bal
```diff
@@ -1121,6 +1121,19 @@ function codeGenExpr(CodeGenContext cx, bir:BasicBlock bb, Environment env, s:Ex
             }
             return cx.semanticErr("can only apply member access to list or mapping", pos=pos);
         }
+        // Field access
+        var { mapping, fieldName, pos } => {
+            var { result: l, block: nextBlock } = check codeGenExpr(cx, bb, env, check cx.foldExpr(env, mapping, ()));
+            if l is bir:Register {
+                t:SemType memberType = t:mappingMemberType(cx.mod.tc, l.semType, fieldName);
```
jclark: You need to implement the requirement in the spec https://ballerina.io/spec/lang/2021R1/#field-access-expr that "K must be a required key type for T".

We should factor out the common code between here and the member access case and maybe also codeGenCompoundAssignToMappingMember

manuranga:
I [tired](https://github.com/ballerina-platform/nballerina/commit/31472ea9e35c340eaf7bf7fdf9f25ceb254c4f5b) to refactor `codeGenCompoundAssignToMappingMember` but I wasn't sure about the `never` case, so I reverted.
 I'll have to think about following case a bit more.

```ballerina
import ballerina/io;

public function main() {
    map<int> m1 = { x: 1 };
    // intersection is empty map
    if m1 is map<string> {
        m1["x"] += "s";
        io:println(m1["x"]);
    }
    else {
        io:println("no");
    }
}
```

manuranga:
I think, for now we can say above case is not allowed but we with https://github.com/ballerina-platform/ballerina-spec/issues/191 this will be allowed and type of the map has to widened.

jclark:
It wouldn't be allowed with https://github.com/ballerina-platform/ballerina-spec/issues/191 unless the member type of m1 allowed nil, because this isn't allowed:

```
map<int> m = {};
m["foo"] = ();
```

manuranga:
Ya, lets say its 
```ballerina
import ballerina/io;

public function main() {
    map<int?> m1 = { x: 1 };
    // intersection is empty map
    if m1 is map<string> {
        m1["x"] += "s";
        io:println(m1["x"]);
    }
    else {
        io:println("no");
    }
}
```
Then it'll be allowed, won't it?, so even in current subset we have to give an error saying `+` is not allowed for operand types, but we currently say `type of member access is never` which I think is wrong.

compiler/modules/front/codeGen.bal
```diff
@@ -907,20 +907,32 @@ function codeGenCompoundAssignToListMember(CodeGenContext cx, bir:BasicBlock bb,
...
+    var { result: k, block: block1 } = check codeGenExprForString(cx, bb, env, check cx.foldExpr(env, lValue.index, t:STRING));
+    var { result: member, block: block2 } = check codeGenMappingGet(cx, block1, mapping, [false, k], pos);
+    var { result, block } = check codeGenCompoundableBinaryExpr(cx, block2, env, op, <bir:Register>member, rexpr, pos);
     bir:MappingSetInsn setInsn = { operands:[ mapping, k, result], position: lValue.pos };
     block.insns.push(setInsn);
     return { block };
 }
 
+// [0] = true if key must be required
+type MappingKey readonly & [false, bir:Register|string]|[true, string];
```
jclark: I think the code will be simpler and easier to understand if you get rid of this type and instead use two arguments: bir:StringOperand and boolean.

I would move codeGenMappingGet to somewhere after codeGenExpr.

compiler/modules/front.syntax/parseStmt.bal
```diff
@@ -153,9 +153,14 @@ function finishOptQualIdentifierStmt(Tokenizer tok, string? prefix, string ident
         return finishCallStmt(tok, expr);
     }
     else if cur == "." {
-        VarRefExpr varRef = { varName: identifier, prefix };
-        MethodCallExpr expr = check finishMethodCallExpr(tok, varRef);
-        return finishCallStmt(tok, expr);
+        VarRefExpr target = { varName: identifier, prefix };
```
jclark: Just inlining finishMethodCallExpr in both places where it was called leaves us with quite a lot of common code here and in finishPrimaryExpr. Can we do better? Maybe have finishMethodCallExpr at least do 

```
   Expr[] args = check parseExprList(tok, ")");
   MethodCallExpr methodCallExpr = { target, methodName, pos, args };
```

Not much but still worthwhile I think.

compiler/modules/front/codeGen.bal
manuranga: That is kind of how I started, but that means I'll have to consider the (illegal) state where boolean is `true` but the operand is a `Register`. Let me give it another try.

compiler/modules/front.syntax/parseStmt.bal
manuranga: With #510, it make scene to also move `check tok.advance();` to `finishMethodCallExpr`. Might as well move it here.

Btw, in future subsets patterns like `x.y[2].z` can occur on both sides. Same kind of logic duplication will happen.

compiler/modules/front/codeGen.bal
```diff
@@ -907,20 +907,29 @@ function codeGenCompoundAssignToListMember(CodeGenContext cx, bir:BasicBlock bb,
...
+    var { result: k, block: block1 } = check codeGenExprForString(cx, bb, env, check cx.foldExpr(env, lValue.index, t:STRING));
+    var { result: member, block: block2 } = check codeGenMappingGet(cx, block1, mapping, false, k, pos);
+    var { result, block } = check codeGenCompoundableBinaryExpr(cx, block2, env, op, member, rexpr, pos);
     bir:MappingSetInsn setInsn = { operands:[ mapping, k, result], position: lValue.pos };
     block.insns.push(setInsn);
     return { block };
 }
 
+// if keyMustRequired is true, k must be a string
+function codeGenMappingGet(CodeGenContext cx, bir:BasicBlock block, bir:Register mapping, boolean keyMustRequired , bir:StringOperand k, err:Position pos) returns CodeGenError|RegExprEffect {
```
jclark, line 919: There's an extra space after keyMustRequired.

The boolean argument (which I know I suggested) makes the callers not very clear. I think we can improve by doing

type MappingAccessType "."|"[";

Then replace `boolean keyMustRequired` by `MappingAccessType accessType`.

compiler/modules/front/codeGen.bal
```diff
@@ -907,20 +907,29 @@ function codeGenCompoundAssignToListMember(CodeGenContext cx, bir:BasicBlock bb,
...
+    t:SemType memberType = t:mappingMemberType(cx.mod.tc, mapping.semType, k is string ? k : ());
+    if !(k is string) || !t:mappingMemberRequired(cx.mod.tc, mapping.semType, k) {
+        if keyMustRequired == true {
+            return cx.semanticErr(`${<string>k} must be a required key`, pos=pos);
```
jclark, line 923: The cast to string bothers me.

jclark (commented):
This is fine. I see a couple of possible tweaks but I will do them.

compiler/modules/front/codeGen.bal
manuranga: Yes, `MappingKey` was used to get rid of it, another option is to add another `if`, but that is ugly because we are doing a runtime check for somthing we know at compile time.

jclark: You can see what I did in https://github.com/ballerina-platform/nballerina/commit/7cd357a348475831547e8e741b3de2273e848aab  I slightly prefer it because it avoids the cast and avoids doing `k is string` twice, but it's a bit longer and there are more local variables, so it's not much of an improvement.

PR #558, opened 2021-10-13
Make llvm:Function behave as a value

### LINKED ISSUE #557: Make llvm:FunctionDecl be an llvm:ConstPointerValue
At the moment we have a toValue method to work around that, but we should find a better way.

heshanpadmasiri:
## Purpose
Make llvm:Function behave as a value
Resolves #557

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/print.llvm/llvm.bal
+++ b/compiler/modules/print.llvm/llvm.bal
@@ -11,10 +11,10 @@ import ballerina/io;
 // Operand of an unnamed variable/basic block
 type Unnamed int;
 
-public type Value RegisterValue|Function;
+public type Value DataValue|Function;
 
 # Corresponds to LLVMValueRef
-public readonly distinct class RegisterValue {
+public readonly distinct class DataValue {
     string|Unnamed operand;
     Type ty;
     function init(Type ty, string|Unnamed operand) {
@@ -26,7 +26,7 @@ public readonly distinct class RegisterValue {
 # Subtype of Value that refers to a pointer
 # Ensures compile-time checking that stores and loads use the right kinds of Value
 public readonly class PointerValue {
-    *RegisterValue;
+    *DataValue;
     string|Unnamed operand;
     PointerType ty;
     function init(PointerType ty, string|Unnamed operand) {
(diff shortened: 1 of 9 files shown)


heshanpadmasiri:
@jclark wouldn't this (making value a union of "old" value (now RegisterValue) and Function types) be a good solution for our problem given that

- For everything outside of llvm this will have no effect. They can use Values as they used to, and also treat functions as Values when needed
- Inside the llvm side there is little to no change required. This is because once we add separate `operand` and `ty` fields (common with all other values) in addition to `function name` and `function type` (common among functions) everything that expect a "old" value are using operand and ty where as Function uses function fields (function name and function type, note that these don't take the same values)

I am asking this because we have bunch of type casts that try to cast RegisterValues ("old" Value) to its sub types and and changes to type of RegisterValues seems to create issues to them.

compiler/modules/print.llvm/llvm.bal
```diff
@@ -11,8 +11,10 @@ import ballerina/io;
 // Operand of an unnamed variable/basic block
 type Unnamed int;
 
-# Corresponds to LLVMValueRef 
-public readonly distinct class Value {
+public type Value RegisterValue|Function;
+
+# Corresponds to LLVMValueRef
+public readonly distinct class RegisterValue {
```
jclark: I don't think RegisterValue is a good name. How about DataValue?

compiler/modules/print.llvm/tests/testBuilder.bal
```diff
@@ -8,7 +8,7 @@ function builderStoreTypeCheck() {
     BasicBlock bb = f.appendBasicBlock();
     Builder builder = context.createBuilder();
     builder.positionAtEnd(bb);
-    Value s = new ("i64", "s");
+    Value s = new RegisterValue("i64", "s");
```
jclark: Can you declare s as type RegisterValue?

heshanpadmasiri: fixed

compiler/modules/print.llvm/llvm.bal
heshanpadmasiri: fixed

jclark (approved):
Looks good. Do we have a test of calling a function ptr that was loaded from memory (e.g. from a struct member)? If not, please add one.

Issue #550, opened 2021-10-11
Update JBUGs for slbeta3


jclark:
Go through all JBUG comments and either:

- if the bug has been fixed, remove the comment and the workaround
- if the bug has not been fixed and does not have a number, file a bug report

KavinduZoysa:
These are the list og JBUGs we need to update. 

- [x] compiler/tests/testBalt.bal: `JBUG: #32575`
- [x] compiler/tests/baltest.bal: `JBUG #31679 should use do`
- [x] compiler/compile.bal: `JBUG gets a bad, sad when done with a query expression`
- [x] compiler/balt.bal: `JBUG: can't use Char gives "'string' value 'x' cannot be converted to 'lang.string:Char'`
- [x] compiler/modules/nback/build.bal: `JBUG #31394 would be better to use shifts for these`
- [x] compiler/modules/nback/build.bal: `JBUG #31753 cast`
- [x] compiler/modules/nback/build.bal: `JBUG <int> cast needed (otherwise result is or'd with 0xFF)`
- [x] compiler/modules/nback/scaffold.bal: `JBUG cast needed #31867`
- [x] compiler/modules/nback/scaffold.bal: `JBUG this goes wrong when you use REPR_VOID as a type as in buildRet`
- [x] compiler/modules/nback/equal.bal: `JBUG cast`
- [x] compiler/modules/types/enumerable.bal: `JBUG: #17977`
- [x] compiler/modules/types/core.bal: `JBUG #31751 bad, sad if UT_OBJECT_RW + 1`
- [x] compiler/modules/front/const.bal: `JBUG toString should not be required`
- [x] compiler/modules/front/codeGen.bal: `JBUG #31311 does not allow `outer` here`
- [x] compiler/modules/front/codeGen.bal: `JBUG #31327 cast`
- [x] compiler/modules/front/codeGen.bal: `JBUG #31327 cast`
- [x] compiler/modules/front/codeGen.bal: `JBUG following line results in bad code`
- [x] compiler/modules/front/codeGen.bal: ``JBUG changing `returns` to `return` makes the parse get a NPE during error recovery``
- [x] compiler/modules/front/codeGen.bal: `JBUG build fails when the order of binding pattern is reversed`
- [x] compiler/modules/front/lib.bal: `JBUG temp variable`
- [x] compiler/modules/front/resolveTypes.bal: `JBUG if this is done with a select, then it gets a bad, sad at runtime if the check gets an error`
[...]

PR #559, opened 2021-10-13
Check JBUGs in compiler/tests

KavinduZoysa:
## Purpose
Check JBUGs in compiler/tests folders
Related to #550

### DIFF
--- a/compiler/tests/baltest.bal
+++ b/compiler/tests/baltest.bal
@@ -23,7 +23,7 @@ function subtypeCaseProvider() returns map<SubtypeTestCase>|error {
         where res is SubtypeTestCase && res[1].length() > 0
         do {
             tests[base] = res;
-        };
+        }; // JBUG #33227 semicolon should not be needed
 
     return tests;
 }
--- a/compiler/tests/baltest.bal
+++ b/compiler/tests/baltest.bal
@@ -13,23 +13,18 @@ function testSubtype(string filename, string[] lines, string[] expected) returns
 
 function subtypeCaseProvider() returns map<SubtypeTestCase>|error {
     map<SubtypeTestCase> tests = {};
-    var testList =
-        from var entry in check file:readDir("tests/data")
+
+    check from var entry in check file:readDir("tests/data")
         let string path = entry.absPath
         let string base = check file:basename(path)
         where path.endsWith(".bal")
         let SubtypeTestCase|error res = readTestCase(path)
         // skip files with no subtype lines on them
         where res is SubtypeTestCase && res[1].length() > 0
-        select [base, res];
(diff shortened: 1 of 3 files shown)


compiler/tests/testBalt.bal
```diff
@@ -7,7 +7,6 @@ import ballerina/file;
 type TestCaseMap map<[string, int, BaltTestHeader, string[]]>;
 
 // disable because there are no .balt tests
-// JBUG: #32575
 // @test:Config {
 //     dataProvider: parseBalts
```
KavinduZoysa, line 11: @jclark, Do I need to change the dataProvider run properly after .balt files are added? Currently, dataProvider does not work with the new folder structure of testSuite.

manuranga: We'll not have `.balt`s for a while, OK to remove the whole thing.

jclark:
I would start with stuff that is not in a tests/ directory.

compiler/tests/testBalt.bal
jclark: I don't want to do this. There are lots of conformance tests being done which we should test against soon.

jclark:
Please don't mix completely unrelated changes in one pull request.

compiler/tests/baltest.bal
```diff
@@ -13,23 +13,18 @@ function testSubtype(string filename, string[] lines, string[] expected) returns
...
         // skip files with no subtype lines on them
         where res is SubtypeTestCase && res[1].length() > 0
-        select [base, res];
-        // JBUG #31679 should use do
-        //  do {
-        //   tests[base] = res;
-        //  };
+        do {
+            tests[base] = res;
+        };
```
jclark: Don't need a semi-colon after the `}`

jclark (changes requested):

compiler/tests/baltest.bal
KavinduZoysa: It seems that removing semicolon causes for a compilation error. Created a new JBUG issue and add a comment accordingly.

KavinduZoysa:
> Please don't mix completely unrelated changes in one pull request.

Removed them.

compiler/tests/baltest.bal
KavinduZoysa: We discussed this in a language issue. Since this do-clause is a part of `checking-action` we need semi-colon. 
@jclark WDYT?

manuranga: Please link the lang issue.

rdulmina: https://github.com/ballerina-platform/ballerina-lang/issues/33227

PR #560, opened 2021-10-13
Resolve JBUG in compile.bal

KavinduZoysa:
## Purpose
Resolve JBUG in compiler/compile.bal
Related to #550

### DIFF
--- a/compiler/compile.bal
+++ b/compiler/compile.bal
@@ -135,7 +135,8 @@ function isSubModule(bir:ModuleId id) returns boolean {
 
 function subModuleSourceParts(string basename, bir:ModuleId id) returns front:SourcePart[]|file:Error|io:Error {
     string directory = check file:joinPath(basename + ".modules", subModuleSuffix(id));
-    return from var md in check file:readDir(directory)
+    return
+        from var md in check file:readDir(directory)
         where !md.dir
         let var [_, ext] = basenameExtension(md.absPath)
         where ext == SOURCE_EXTENSION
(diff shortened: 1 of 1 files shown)


compiler/compile.bal
```diff
@@ -136,22 +136,15 @@ function isSubModule(bir:ModuleId id) returns boolean {
...
-                    filename: check file:normalizePath(check file:joinPath(directory, check file:basename(md.absPath)), file:CLEAN),
-                    directory 
-                });
-            }
-
-        }
-    }
-    return sourceParts;
+    return from var md in check file:readDir(directory)
+        where !md.dir
```
jclark, line 140: Ah, this is so much nicer.

But please align let/where/select with from.

jclark (changes requested):

compiler/compile.bal
KavinduZoysa: Changed

compiler/compile.bal
```diff
@@ -136,22 +136,15 @@ function isSubModule(bir:ModuleId id) returns boolean {
...
-
-        }
-    }
-    return sourceParts;
+    return from var md in check file:readDir(directory)
+           where !md.dir
+           let var [_, ext] = basenameExtension(md.absPath)
+           where ext == SOURCE_EXTENSION
+           select {
+                lines: check io:fileReadLines(md.absPath),
```
jclark: This is indented 5 spaces relative to `select`, which is not correct. Maybe it will work better if you put the `from` on a new line.

KavinduZoysa: Changed

jclark (approved):
Looks good.

Issue #551, opened 2021-10-11
Need semtype testing strategy for set difference/complement


jclark:
Conditional type narrowing gives rise to semtypes that have use set difference (or equivalently complement/negation). We need a way to be able to test subtype relationships and projection with types that use set difference.

One possibility is to implement a `!` type operator. Is there another way?

manuranga:
Another option is to extended assertion grammar.

Currently planned grammar is:
```
type-projection := identifier ("[" number "]")?
type-assertion := type-projection("<"|"="|"<>") type-projection
```

We could extend it to support something like
```
// @test T = S - R
```
Or 
```
// @test T < S & !R
```

jclark:
We need to be able to test the projection of a difference, which would require something like

```
// @test (S - T)[1] = R
```

That feels too complicated to me.

jclark:
With #623 we can just support a not `!` when we are testing types.

PR #562, opened 2021-10-14
Introduce xml subtypes

### LINKED ISSUE #526: Implement xml subtypes in types module
The types module supports the `xml` type. This adds support for xml subtypes.

rdhananjaya:
## Purpose
Add xml:Element, xml:Comment, xml:ProcessingInstruction, xml:Text, xml&lt;xml:Text&gt;, and so on.

Fixes https://github.com/ballerina-platform/nballerina/issues/526

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,200 @@
...
+        rw: {
+            data: rwBitset,
+            sequence: false
+        }
+    };
+    UniformTypeCode typeCode = rwBitset == 0 ? UT_XML_RO : UT_XML_RW;
+    return uniformSubtype(typeCode, st);
+}
+
+public function xmlSequence(SemType constituentType) returns SemType {
```
jclark, line 37: Need to be clear about how this function deals with case when constituent type is not a subtype of XML.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,200 @@
...
+    if constituentType == XML {
+        return constituentType;
+    }
+    var roData = getComplexSubtypeData(<ComplexSemType>constituentType, UT_XML_RO);
```
jclark: These casts are not safe.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,200 @@
...
+    return { ro, rw };
+}
+
+function xmlSubtypeUnion(SubtypeData d1, SubtypeData d2) returns SubtypeData {
+    XmlSubtype v1 = <XmlSubtype>d1;
+    XmlSubtype v2 = <XmlSubtype>d2;
+    int dataRo = v1.ro.data | v2.ro.data;
+    int dataRw = v1.rw.data | v2.rw.data;
+    return createXmlSubtype(
+        { data :dataRo, sequence: bddUnion(v1.ro.sequence, v2.ro.sequence) },
```
jclark: Space like `data: dataRo`.

jclark (changes requested):

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,155 @@
...
+    else {
+        return d;
+    }
+}
+
+function createXmlSubtype(int data, Bdd sequence) returns SubtypeData {
+    if data == 0 && sequence == false {
+        return false;
+    }
+    else if (data & XML_NEVER) != 0 && sequence == true {
```
jclark: This isn't right: you need to check that it allows everything that is possible, which is different for the ro/rw case, so you will need an argument to the function.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,188 @@
+// Implementation specific to basic type xml.
+import wso2/nballerina.err;
+
+public type XmlSubtypeData readonly & record {|
+    int data;
```
jclark: `data` isn't a good name.  This needs a comment explaining what the values are.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,188 @@
...
+    Bdd sequence;
+|};
+
+const int XML_NEVER = 1;
+const int TEXT = 1 << 1;
+const int ELEM_RO = 1 << 2;
```
jclark: ELEMENT is more consistent

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,188 @@
...
+const int PI_RW = 1 << 6;
+const int COMMENT_RW = 1 << 7;
+
+const int XML_RO_MASK = XML_NEVER | TEXT | ELEM_RO | PI_RO | COMMENT_RO;
+const int XML_RW_MASK = ELEM_RW | PI_RW | COMMENT_RW;
+const int XML_RO_SEQ_MASK = TEXT | ELEM_RO | PI_RO | COMMENT_RO;
+
+final XmlSubtypeData emptyXmlSubtype = { data: 0, sequence: false };
+
+function xmlSingleton(int bits) returns SemType {
```
jclark: You are using `bits` here to mean what `data` means in XmlSubtypeData.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,188 @@
...
+        { data: bits & XML_RW_MASK, sequence: false }
+    );
+}
+
+public function xmlSequence(SemType constituentType, err:Location loc) returns SemType|err:Semantic {
+    if constituentType == XML {
+        return constituentType;
+    }
+    if constituentType is UniformTypeBitSet {
+        return err:semantic("Invalid xml constituent type", loc=loc);
```
jclark: Don't capitalize error messages,

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,188 @@
...
+    }
+    ComplexSemType t = <ComplexSemType>constituentType;
```
jclark: Use `else` rather than casting.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,188 @@
...
+    if hasNonXmlTypes(t) {
```
jclark: Just start off at the beginning by doing if !t:isSubtypeSimple(ty, t:XML)`

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,188 @@
...
+    return (t.all & ~XML) != 0 || (t.some & ~XML) != 0;
+}
+
+
+function makeSequence(XmlSubtypeData d) returns XmlSubtypeData {
+    if d.sequence == false {
+        return  { data: XML_NEVER | d.data, sequence: bddCreate(d.data, true, false, false) };
+    }
+    else {
+        return d;
```
jclark: That's not right. The singletons can occur in the sequence

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,188 @@
...
+    }
+}
+
+function createXmlSubtype(boolean isRo, int data, Bdd sequence) returns SubtypeData {
+    if data == 0 && sequence == false {
+        return false;
+    }
+    boolean containsAllData = isRo ? (data & XML_RO_MASK) == XML_RO_MASK : (data & XML_RW_MASK) == XML_RW_MASK;
```
jclark: I would write it like this:

```
if sequence == false {
    if data == 0 {
        return false;
    }
}
else if sequence == true {
    int mask = isRo ? XML_RO_MASK : XML_RW_MASK;
    if (data & mask) == mask {
        return true;
    }
}
return { data, sequence };
```

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,188 @@
...
+    if containsAllData && sequence == true {
+        return true;
+    }
+    return { data, sequence };
+}
+
+function createXmlSemtype(XmlSubtypeData ro, XmlSubtypeData rw) returns ComplexSemType {
```
jclark: This function doesn't seem right to me. Why are we getting an emptyXmlSubtype?

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,188 @@
...
+}
+
+function xmlSubtypeDiff(boolean isRo, SubtypeData d1, SubtypeData d2) returns SubtypeData {
+    XmlSubtypeData v1 = <XmlSubtypeData>d1;
+    XmlSubtypeData v2 = <XmlSubtypeData>d2;
+    int data = v1.data & ~v2.data;
+    return createXmlSubtype(isRo, data, bddDiff(v1.sequence, v2.sequence));
+}
+
+function xmlSubtypeComplement(boolean isRo, SubtypeData d) returns SubtypeData {
```
jclark, line 112: This doesn't look right.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,188 @@
...
+}
+
+function xmlSubtypeUnion(boolean isRo, SubtypeData d1, SubtypeData d2) returns SubtypeData {
+    XmlSubtypeData v1 = <XmlSubtypeData>d1;
+    XmlSubtypeData v2 = <XmlSubtypeData>d2;
+    int data = v1.data | v2.data;
+    return createXmlSubtype(isRo, data, bddUnion(v1.sequence, v2.sequence));
+}
+
+function xmlSubtypeIntersect(boolean isRo, SubtypeData d1, SubtypeData d2) returns SubtypeData {
```
jclark: intersect of two sets can only produce something smaller than the args. So there is no possibility of producing true from this. So you don't need isRo as an argument (because isRo is only used to detect whether you've gort true.)  So the ops tables can use this directly. Similarly for diff.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,188 @@
...
+    Bdd seq = bddCreate(rootAtom, bddComplement(sd.sequence), false, false);
+    return createXmlSubtype(isRo, complement, seq);
+}
+
+function xmlSubtypeIsEmpty(Context cx, SubtypeData d) returns boolean {
+    XmlSubtypeData sd = <XmlSubtypeData>d;
+    if sd.data != 0 {
+        return false;
+    }
+    return !xmlSequenceInhabited(cx, sd.sequence);
```
jclark: I would call this xmlBddEmpty and have it return the negation of what or currently returns

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,188 @@
...
+function xmlSequenceInhabited(Context cx, Bdd bdd) returns boolean {
+    boolean isEmpty = bddEvery(cx, bdd, (), (), xmlFormulaIsEmpty);
+    return !isEmpty;
+}
+
+function xmlFormulaIsEmpty(Context cx, Conjunction? pos, Conjunction? neg) returns boolean {
+    return hasIntersectionAll(allBits(pos), neg);
+}
+
+function hasIntersectionAll(int bits, Conjunction? c) returns boolean {
```
jclark: I don't think it's helpful to write this function or `allBits` recursively.

xmlFormulaIsEmpty should just loop over pos to build an `int` with the bits.
Then loop over the neg, returing true if any neg has all the pos bits.

jclark (commented):
This looks like it is on the right lines.  I haven't looked at the other files yet.

compiler/modules/types/xml.bal
rdhananjaya: So inline the call to createXmlSubtype?

jclark: Factor out the part of createXmlSubtype that deals with false.

compiler/modules/front/resolveTypes.bal
```diff
@@ -229,6 +229,10 @@ function resolveTypeDesc(ModuleSymbols mod, s:ModuleLevelDefn modDefn, int depth
     if td is s:XmlSequenceTypeDesc {
         t:SemType t = check resolveTypeDesc(mod, modDefn, depth, td.constituent);
         err:Location loc =  err:location(modDefn.part.file, td.pos);
+        
+        if !t:isSubtypeSimple(t, t:XML) {
+            return err:semantic("non xml type as xml constituent type", loc=loc);
```
manuranga: Should this be `constraint type`?

manuranga: Same for the `XmlSequenceTypeDesc` field.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,164 @@
...
+|};
+
+const int XML_NEVER = 1;
+const int TEXT = 1 << 1;
+const int ELEMENT_RO = 1 << 2;
+const int PI_RO = 1 << 3;
+const int COMMENT_RO = 1 << 4;
+const int ELEMENT_RW = 1 << 5;
+const int PI_RW = 1 << 6;
+const int COMMENT_RW = 1 << 7;
```
manuranga: minor: consider aligning.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,164 @@
...
+    }
+    return createXmlSubtypeOrEmpty(primitives, sequence);
+}
+
+function createXmlSubtypeOrEmpty(int primitives, Bdd sequence) returns SubtypeData {
+    if sequence == false {
+        if primitives == 0 {
+            return false;
+        }
+    }
```
manuranga: Can combine the `if`s.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,164 @@
...
+import wso2/nballerina.err;
+
+public type XmlSubtypeData readonly & record {|
+    // Each xml singleton and the empty sequence is considered a primitive.
+    int primitives;
+    Bdd sequence;
+|};
+
+const int XML_NEVER = 1;
+const int TEXT = 1 << 1;
```
jclark: We should put XML_ before these since their scope is the entire module not just this file.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,164 @@
+// Implementation specific to basic type xml.
+import wso2/nballerina.err;
+
+public type XmlSubtypeData readonly & record {|
+    // Each xml singleton and the empty sequence is considered a primitive.
+    int primitives;
+    Bdd sequence;
+|};
+
+const int XML_NEVER = 1;
```
jclark: There needs to be a comment that connects these consts to the primitives field and explains that the primitives field is a bitwise-or of these constants.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,164 @@
...
+const int COMMENT_RO = 1 << 4;
+const int ELEMENT_RW = 1 << 5;
+const int PI_RW = 1 << 6;
+const int COMMENT_RW = 1 << 7;
+
+const int XML_RO_MASK = XML_NEVER | TEXT | ELEMENT_RO | PI_RO | COMMENT_RO;
+const int XML_RW_MASK = ELEMENT_RW | PI_RW | COMMENT_RW;
+
+final XmlSubtypeData xmlRoAll = { primitives: XML_RO_MASK, sequence: bddAtom(XML_RO_MASK & ~XML_NEVER) };
+final XmlSubtypeData xmlRwAll = { primitives: XML_RW_MASK, sequence: bddAtom(XML_RW_MASK) };
```
jclark: You can have a XML value that is a sequence of length >= 2 that includes both mutable and immutable singletons, for example one mutable XML element singleton and one immutable XML element singleton. Such a sequence would be part of the XML_RW uniform type. This means that bddAtom in the Rw case needs to use a mask that allows immutable things as well as mutable things.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,164 @@
...
+const int XML_NEVER = 1;
+const int TEXT = 1 << 1;
+const int ELEMENT_RO = 1 << 2;
+const int PI_RO = 1 << 3;
+const int COMMENT_RO = 1 << 4;
+const int ELEMENT_RW = 1 << 5;
+const int PI_RW = 1 << 6;
+const int COMMENT_RW = 1 << 7;
+
+const int XML_RO_MASK = XML_NEVER | TEXT | ELEMENT_RO | PI_RO | COMMENT_RO;
```
jclark: I would define XML_RO_SINGLETONS (excluding XML_NEVER) and then use that later on.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,164 @@
...
+const int XML_RW_MASK = ELEMENT_RW | PI_RW | COMMENT_RW;
+
+final XmlSubtypeData xmlRoAll = { primitives: XML_RO_MASK, sequence: bddAtom(XML_RO_MASK & ~XML_NEVER) };
+final XmlSubtypeData xmlRwAll = { primitives: XML_RW_MASK, sequence: bddAtom(XML_RW_MASK) };
+
+function xmlSingleton(int primitives) returns SemType {
+    return createXmlSemtype(
+        { primitives: primitives & XML_RO_MASK, sequence: false },
```
jclark: I believe you need to use createXmlSubtype before calling createXmlSemtype.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,164 @@
...
+final XmlSubtypeData xmlRwAll = { primitives: XML_RW_MASK, sequence: bddAtom(XML_RW_MASK) };
+
+function xmlSingleton(int primitives) returns SemType {
+    return createXmlSemtype(
+        { primitives: primitives & XML_RO_MASK, sequence: false }, 
+        { primitives: primitives & XML_RW_MASK, sequence: false }
+    );
+}
+
+public function xmlSequence(SemType constituentType, err:Location loc) returns SemType {
```
jclark: Consider `xml<xml:Element>`. The XML_RW half of this has to allow a combination of mutable and immutable elements. In other words, you cannot do makeSequence independently for the ro/rw cases.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,164 @@
...
+        return false;
+    }
+    return xmlBddEmpty(cx, sd.sequence);
+}
+
+function xmlBddEmpty(Context cx, Bdd bdd) returns boolean {
+    return bddEvery(cx, bdd, (), (), xmlFormulaIsEmpty);
+}
+
+function xmlFormulaIsEmpty(Context cx, Conjunction? pos, Conjunction? neg) returns boolean {
```
jclark: For the RW case, we also have to detect when it allows only RO constituents.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,164 @@
...
+public function xmlSequence(SemType constituentType, err:Location loc) returns SemType {
+    if constituentType is UniformTypeBitSet {
+        return constituentType;
+    }   
+    else {
+        SubtypeData ro = getComplexSubtypeData(constituentType, UT_XML_RO);
+        ro = ro is boolean ? ro : makeSequence(<XmlSubtypeData>ro);
+        
+        SubtypeData rw = getComplexSubtypeData(constituentType, UT_XML_RW);
+        rw = rw is boolean ? rw : makeSequence(<XmlSubtypeData>rw);
```
manuranga: Shouldn't some reusing of BDDs happen here? or is it happening elsewhere?

jclark (changes requested):
I think we're nearly there, but we have to deal with sequences that contain a mixture of mutable and immutable singletons.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,164 @@
...
+        
+        return createXmlSemtype(ro, rw);
+    }
+}
+
+function makeSequence(XmlSubtypeData d) returns XmlSubtypeData {
+    return { 
+        primitives: XML_NEVER | d.primitives, 
+        sequence: bddUnion(bddAtom(d.primitives & ~XML_NEVER), d.sequence) 
+        };
```
manuranga: minor: formatting looks bit off to me.

compiler/modules/front/resolveTypes.bal
rdhananjaya: `constituent` is spec term used in XML sequences.

compiler/modules/types/xml.bal
rdhananjaya: How about `XML_TEXT_BIT`, `XML_TEXT` is already taken for the semtype.

compiler/modules/front/resolveTypes.bal
manuranga: `constituent` in spec refer to the values not the type `T` in `xml<T>`.
But it makes sense here, because it's followed by the word 'type', not sure about it's use in `XmlSequenceTypeDesc`.

compiler/modules/types/xml.bal
jclark: I suggest you use a prefix of `XML_PRIMITIVE_` for all of them, which ties them to the `primitives` field.

jclark:
I just merged in Ushira's new semtype testing stuff. Shall we try to use that? The readonly aspect of this is particularly tricky, and we are inventing all this rather than relying on the Castagna paper, so we need tests that probe all the difficult edge cases.

compiler/modules/front/resolveTypes.bal
rdhananjaya: I really don't have a strong opinion on this. I was tempted to use `constraint` here as it's what we have in jBallerina type. But since xml sequence is immutable, I didn't feel constraint means anything here. In a map<T>, T is a constraint as we can't manipulate the value to go over T, but in XML case it just who what is this sequence made out of.

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -94,7 +93,6 @@ function parsePrimaryTypeDesc(Tokenizer tok) returns TypeDesc|err:Syntax {
         |  "readonly" => {
             Position endPos = tok.currentEndPos();
             check tok.advance();
-            // JBUG should not need cast #30191
```
manuranga: is this removed by accident ?

rdhananjaya: oh, yes my bad, let me reset this

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,196 @@
...
+const int XML_PRIMITIVE_NEVER_BIT      = 1;
+const int XML_PRIMITIVE_TEXT_BIT       = 1 << 1;
+const int XML_PRIMITIVE_ELEMENT_RO_BIT = 1 << 2;
+const int XML_PRIMITIVE_PI_RO_BIT      = 1 << 3;
+const int XML_PRIMITIVE_COMMENT_RO_BIT = 1 << 4;
+const int XML_PRIMITIVE_ELEMENT_RW_BIT = 1 << 5;
+const int XML_PRIMITIVE_PI_RW_BIT      = 1 << 6;
+const int XML_PRIMITIVE_COMMENT_RW_BIT = 1 << 7;
+
+const int XML_PRIMITIVE_RO_SINGLTONS = XML_PRIMITIVE_TEXT_BIT | XML_PRIMITIVE_ELEMENT_RO_BIT | XML_PRIMITIVE_PI_RO_BIT | XML_PRIMITIVE_COMMENT_RO_BIT;
```
jclark: SINGLETONS has an E

compiler/modules/types/core.bal
```diff
@@ -188,7 +188,7 @@ public class Context {
     }
 }
 
-type ProperSubtypeData StringSubtype|DecimalSubtype|FloatSubtype|IntSubtype|BooleanSubtype|BddNode;
+type ProperSubtypeData StringSubtype|DecimalSubtype|FloatSubtype|IntSubtype|BooleanSubtype|XmlSubtypeData|BddNode;
```
jclark: Naming here is not consistent. Should use `XmlSubtype` here.

compiler/modules/types/core.bal
```diff
@@ -364,6 +364,10 @@ public final UniformTypeBitSet SIMPLE_OR_STRING = uniformTypeUnion((1 << UT_NIL)
 public final UniformTypeBitSet NUMBER = uniformTypeUnion((1 << UT_INT) | (1 << UT_FLOAT) | (1 << UT_DECIMAL));
 public final SemType BYTE = intWidthUnsigned(8);
 public final SemType STRING_CHAR = stringChar();
+public final SemType XML_PRIMITIVE_ELEMENT = xmlSingleton(XML_PRIMITIVE_ELEMENT_RO_BIT | XML_PRIMITIVE_ELEMENT_RW_BIT);
```
jclark: No need for `_PRIMITIVE` in the name of these.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,196 @@
...
+    // Each xml singleton and the empty sequence is considered a primitive.
+    int primitives;
+    Bdd sequence;
+|};
+
+// Each bit represents a XML singleton subtype or a empty sequence.
+// `primitives` fields are composed of allowed values from this singleton set.
+// Atom of the `Bdd` represented by `sequence` field is also composed of these to indicate 
+// non-empty sequence of allowed singletons.
+const int XML_PRIMITIVE_NEVER_BIT      = 1;
```
jclark: I don't think we need BIT in the name of these as well as PRIMITIVE.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,196 @@
+// Implementation specific to basic type xml.
+
+public type XmlSubtypeData readonly & record {|
```
jclark: This should be just XmlSubtype.

compiler/modules/front/resolveTypes.bal
jclark: "type parameter for xml is not a subtype of xml"

compiler/modules/front/resolveTypes.bal
```diff
@@ -252,6 +252,15 @@ function resolveTypeDesc(ModuleSymbols mod, s:ModuleLevelDefn modDefn, int depth
             return t:intersect(l, r);
         }
     }
+    if td is s:XmlSequenceTypeDesc {
+        t:SemType t = check resolveTypeDesc(mod, modDefn, depth, td.constituent);
+        err:Location loc =  err:location(modDefn.part.file, td.pos);
+        
+        if !t:isSubtypeSimple(t, t:NEVER) && !t:isSubtypeSimple(t, t:XML) {
```
jclark: `!t:isSubtypeSimple(t, t:NEVER)` is the same as `t != t:NEVER`. But why do we need this? IIRC, `isSubtypeSimple(t:NEVER, t:XML)` is true.

jclark (commented):
Are we recognizing that `xml<xml:Text>` and `xml:Text` are the same thing? I don't think we are...

I also think we need to be able to test with `!` to make sure we get stuff right. We can now do that: https://github.com/ballerina-platform/nballerina/issues/551#issuecomment-953462130

jclark:
I have now implemented  a `!` type operator that is allowed in type descriptors when you are doing showTypes or `-t` tests. So we can now put in tests for all the various difficult cases we discussed.

compiler/testSuite/semtype/xml-complex-ro-t.bal
```diff
@@ -0,0 +1,41 @@
...
+// @type E < NonEmptyS
+// @type P < NonEmptyS
+// @type C < NonEmptyS
+type NonEmptyS S & !N;
+
+// @type NonEmptyS < UX
+type UX XE|XP|XC|T;
+
+// @type XNonEmptyS = RX
+// @type XNonEmptyS < X
```
jclark: Don't understand this one: `xml<T>` means sequence of length 0 or more where every member belongs to T. So I would say: XNonEmptyS = X. What am I missing?

jclark (commented):
I still need to look at this some more.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,196 @@
...
+        
+        SubtypeData rw = getComplexSubtypeData(constituentType, UT_XML_RW);
+        rw = rw is boolean ? rw : makeSequence(false, <XmlSubtype>rw);
+        
+        return createXmlSemtype(ro, rw);
+    }
+}
+
+function makeSequence(boolean roPart, XmlSubtype d) returns XmlSubtype {
+    return {
```
jclark: Shouldn't this use createXmlSubtype? Otherwise, you are not going to recognize all the cases where this is equivalent to true/false.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,196 @@
...
+        ro = ro is boolean ? ro : makeSequence(true, <XmlSubtype>ro);
+        
+        SubtypeData rw = getComplexSubtypeData(constituentType, UT_XML_RW);
+        rw = rw is boolean ? rw : makeSequence(false, <XmlSubtype>rw);
+        
+        return createXmlSemtype(ro, rw);
+    }
+}
+
+function makeSequence(boolean roPart, XmlSubtype d) returns XmlSubtype {
```
jclark: How does this deal with the case where we are creating the rw part, and d.sequence is false and all the primitives are ro?

e.g. `xml<readonly & xml:Element> & !readonly` should equivalent to `never`.

rdhananjaya: I think I forgot to update this when addressing https://github.com/ballerina-platform/nballerina/pull/562#pullrequestreview-780698845; will update.

compiler/testSuite/semtype/xml-complex-ro-t.bal
rdhananjaya: I improved the names reflect the type.
Here since T is `readonly`, xml<T> can only contain readonly parts hence it's a subtype of `X`

jclark: OK. Got it.

compiler/modules/types/xml.bal
```diff
@@ -0,0 +1,197 @@
...
+const int XML_PRIMITIVE_PI_RO      = 1 << 3;
+const int XML_PRIMITIVE_COMMENT_RO = 1 << 4;
+const int XML_PRIMITIVE_ELEMENT_RW = 1 << 5;
+const int XML_PRIMITIVE_PI_RW      = 1 << 6;
+const int XML_PRIMITIVE_COMMENT_RW = 1 << 7;
+
+const int XML_PRIMITIVE_RO_SINGLETONS = XML_PRIMITIVE_TEXT | XML_PRIMITIVE_ELEMENT_RO | XML_PRIMITIVE_PI_RO | XML_PRIMITIVE_COMMENT_RO;
+const int XML_PRIMITIVE_RO_MASK = XML_PRIMITIVE_NEVER | XML_PRIMITIVE_RO_SINGLETONS;
+const int XML_PRIMITIVE_RW_MASK = XML_PRIMITIVE_ELEMENT_RW | XML_PRIMITIVE_PI_RW | XML_PRIMITIVE_COMMENT_RW;
+const int XML_RW_ATOM = XML_PRIMITIVE_RO_SINGLETONS | XML_PRIMITIVE_RW_MASK;
```
jclark: This name seems a bit inconsistent with the others. I would call this `XML_PRIMITIVE_SINGLETON` and rename `XML_PRIMITIVE_RO_SINGLETONS` to `XML_PRIMITIVE_RO_SINGLETON`.

jclark (changes requested):
Looks just. Just a small naming tweak and we are done!

jclark (changes requested):
Can you also update docs?

* Add to docs/type-subset.md
* Cross off from compiler/modules/types/TODO.md
* In compiler/modules/types/NOTES.md, update and move to the Done section

PR #564, opened 2021-10-14
Resolve JBUG in build.bal related to const 

KavinduZoysa:
## Purpose
Use shift operator for defined const
Related to #550

### DIFF
--- a/compiler/modules/nback/build.bal
+++ b/compiler/modules/nback/build.bal
@@ -6,9 +6,10 @@ import wso2/nballerina.print.llvm;
 type Alignment 1|8;
 
 // Pointer tagging
-                     //1234567812345678
-const POINTER_MASK = 0x00fffffffffffff8;
 const TAG_SHIFT = 56;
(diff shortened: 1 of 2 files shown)


compiler/modules/nback/build.bal
```diff
@@ -6,28 +6,25 @@ import wso2/nballerina.print.llvm;
 type Alignment 1|8;
 
 // Pointer tagging
-// JBUG #31394 would be better to use shifts for these
                      //1234567812345678
-const TAG_FACTOR   = 0x0100000000000000;
 const POINTER_MASK = 0x00fffffffffffff8;
```
jclark: Compute this in terms of TAG_SHIFT.  Something like `((1 << TAG_SHIFT) - 1) & ~(ALIGN_HEAP - 1)` (please check I got that right)

KavinduZoysa: Changed

jclark (approved):
Looks good.

PR #567, opened 2021-10-15
Add positions to expression AST nodes

heshanpadmasiri:
## Purpose
Add positions to expression AST nodes
Related #513

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/front.syntax/parseExpr.bal
+++ b/compiler/modules/front.syntax/parseExpr.bal
@@ -32,7 +32,7 @@ function parseBitwiseOrExpr(Tokenizer tok) returns Expr|err:Syntax {
         if t == "|" {
             check tok.advance();
             Expr right = check parseBitwiseXorExpr(tok);
-            Position endPos = tok.currentEndPos();
+            Position endPos = tok.previousEndPos();
             BinaryBitwiseExpr bin = { startPos, endPos,  bitwiseOp: t, left: expr, right };
             expr = bin;
         }
@@ -51,7 +51,7 @@ function parseBitwiseXorExpr(Tokenizer tok) returns Expr|err:Syntax {
         if t == "^" {
             check tok.advance();
             Expr right = check parseBitwiseAndExpr(tok);
-            Position endPos = tok.currentEndPos();
+            Position endPos = tok.previousEndPos();
             BinaryBitwiseExpr bin = { startPos, endPos, bitwiseOp: t, left: expr, right };
             expr = bin;
         }
@@ -70,7 +70,7 @@ function parseBitwiseAndExpr(Tokenizer tok) returns Expr|err:Syntax {
         if t == "&" {
             check tok.advance();
             Expr right = check parseEqualityExpr(tok);
-            Position endPos = tok.currentEndPos();
+            Position endPos = tok.previousEndPos();
             BinaryBitwiseExpr bin = { startPos, endPos, bitwiseOp: t, left: expr, right };
             expr = bin;
         }
@@ -89,7 +89,7 @@ function parseEqualityExpr(Tokenizer tok)  returns Expr|err:Syntax {
         if t is BinaryEqualityOp {
             check tok.advance();
             Expr right = check parseRelationalExpr(tok);
-            Position endPos = tok.currentEndPos();
+            Position endPos = tok.previousEndPos();
             BinaryEqualityExpr bin = { startPos, endPos, equalityOp: t, left: expr, right };
             expr = bin;
         }
@@ -107,7 +107,7 @@ function parseRelationalExpr(Tokenizer tok) returns Expr|err:Syntax {
     if t is BinaryRelationalOp {
         check tok.advance();
         Expr right = check parseShiftExpr(tok);
-        Position endPos = tok.currentEndPos();
+        Position endPos = tok.previousEndPos();
         BinaryRelationalExpr bin = { startPos, endPos, relationalOp: t, left: expr, right };
         return bin;
     }
@@ -141,7 +141,7 @@ function parseRangeExpr(Tokenizer tok) returns RangeExpr|err:Syntax {
     Expr lower = check parseAdditiveExpr(tok);
     check tok.expect("..<");
     Expr upper = check parseAdditiveExpr(tok);
-    Position endPos = tok.currentEndPos();
+    Position endPos = tok.previousEndPos();
     return { startPos, endPos, lower, upper };
 }
 
@@ -153,7 +153,7 @@ function parseShiftExpr(Tokenizer tok) returns Expr|err:Syntax {
         if t is ("<<"|">>>"|">>") {
             check tok.advance();
             Expr right = check parseAdditiveExpr(tok);
-            Position endPos = tok.currentEndPos();
+            Position endPos = tok.previousEndPos();
             BinaryBitwiseExpr shift = { startPos, endPos, bitwiseOp: t, left: expr, right };
             expr = shift;
         }
@@ -173,7 +173,7 @@ function parseAdditiveExpr(Tokenizer tok) returns Expr|err:Syntax {
             Position pos = tok.currentStartPos();
             check tok.advance();
             Expr right = check parseMultiplicativeExpr(tok);
-            Position endPos = tok.currentEndPos();
+            Position endPos = tok.previousEndPos();
             BinaryArithmeticExpr bin = { startPos, endPos, arithmeticOp: t, left: expr, right, pos };
             expr = bin;
         }
@@ -193,7 +193,7 @@ function parseMultiplicativeExpr(Tokenizer tok) returns Expr|err:Syntax {
             Position pos = tok.currentStartPos();
             check tok.advance();
             Expr right = check parseUnaryExpr(tok);
-            Position endPos = tok.currentEndPos();
+            Position endPos = tok.previousEndPos();
             BinaryArithmeticExpr bin = { startPos, endPos, arithmeticOp: t, left: expr, right, pos };
             expr = bin;
         }
@@ -211,14 +211,14 @@ function parseUnaryExpr(Tokenizer tok) returns Expr|err:Syntax {
         Position pos = tok.currentStartPos();
         check tok.advance();
         Expr operand = check parseUnaryExpr(tok);
-        Position endPos = tok.currentEndPos();
+        Position endPos = tok.previousEndPos();
         UnaryExpr expr = { startPos, endPos, op: t, operand, pos };
         return expr;
     }
     else if t is CheckingKeyword {
         check tok.advance();
         Expr operand = check parseUnaryExpr(tok);
-        Position endPos = tok.currentEndPos();
+        Position endPos = tok.previousEndPos();
         CheckingExpr expr = { startPos, endPos, checkingKeyword: t, operand };
         return expr;
     }
@@ -236,7 +236,7 @@ function parseTypeCastExpr(Tokenizer tok, Position startPos) returns Expr|err:Sy
     check tok.expect(">");
     tok.setMode(MODE_NORMAL);
     Expr operand = check parseUnaryExpr(tok);
-    Position endPos = tok.currentEndPos();
+    Position endPos = tok.previousEndPos();
     TypeCastExpr expr = { startPos, endPos, pos, td, operand };
     return expr;
 }
--- a/compiler/modules/front.syntax/parseExpr.bal
+++ b/compiler/modules/front.syntax/parseExpr.bal
@@ -2,16 +2,19 @@ import wso2/nballerina.err;
 
 function parseExpr(Tokenizer tok) returns Expr|err:Syntax {
     Token? t = tok.current();
+    Position startPos = tok.currentStartPos();
     if t == "[" {
         check tok.advance();
         Expr[] members = check parseExprList(tok, "]");
-        ListConstructorExpr expr = { members };
+        Position endPos = tok.previousEndPos();
+        ListConstructorExpr expr = { startPos, endPos, members };
         return expr;
     }
     else if t == "{" {
         check tok.advance();
         Field[] fields = check parseFields(tok);
-        MappingConstructorExpr expr = { fields };
+        Position endPos = tok.previousEndPos();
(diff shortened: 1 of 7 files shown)


compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -184,78 +206,92 @@ function parseMultiplicativeExpr(Tokenizer tok) returns Expr|err:Syntax {
...
+            return finishFunctionCallExpr(tok, prefix, varName, pos, startPos);
         }
-        return { prefix, varName };
+        if prefix != () {
+            endPos = tok.previousEndPos();
+        }
+        else {
+            endPos = tok.currentEndPos();
+        }
+        return { startPos, endPos,prefix, varName };
```
jclark: Space missing before `prefix`.

compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -184,78 +206,92 @@ function parseMultiplicativeExpr(Tokenizer tok) returns Expr|err:Syntax {
...
+    Position startPos = tok.currentStartPos();
+    return finishPrimaryExpr(tok, check startPrimaryExpr(tok), startPos);
 }
 
 function startPrimaryExpr(Tokenizer tok) returns Expr|err:Syntax {
     Token? t = tok.current();
+    Position startPos = tok.currentStartPos();
+    Position endPos = tok.currentEndPos();
     if t is [IDENTIFIER, string] {
         Position pos = tok.currentStartPos();
```
jclark, line 254: This section needs reworking. Why do we need pos and startPos here?

compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -184,78 +206,92 @@ function parseMultiplicativeExpr(Tokenizer tok) returns Expr|err:Syntax {
...
             check tok.advance();
-            return finishFunctionCallExpr(tok, prefix, varName, pos);
+            return finishFunctionCallExpr(tok, prefix, varName, pos, startPos);
         }
-        return { prefix, varName };
+        if prefix != () {
+            endPos = tok.previousEndPos();
+        }
+        else {
+            endPos = tok.currentEndPos();
```
jclark: Don't think this is right. We advanced over the identifier. Isn't it just endPos saved above?

compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -264,63 +300,66 @@ function startPrimaryExpr(Tokenizer tok) returns Expr|err:Syntax {
...
     if t == "[" {
         Position pos = tok.currentStartPos();
+        Position accessStartPos = pos;
         check tok.advance();
         Expr index = check parseInnerExpr(tok);
-        check tok.expect("]");
-        MemberAccessExpr accessExpr = { container: expr, index, pos };
-        return finishPrimaryExpr(tok, accessExpr);
+        Position accessEndPos = check tok.expectEnd("]");
+        MemberAccessExpr accessExpr = { startPos:accessStartPos, endPos: accessEndPos, container: expr, index, pos };
```
jclark: Space after colon

compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -264,63 +300,66 @@ function startPrimaryExpr(Tokenizer tok) returns Expr|err:Syntax {
...
         return parseError(tok);
     }
 }
 
-function finishPrimaryExpr(Tokenizer tok, Expr expr) returns Expr|err:Syntax {
+function finishPrimaryExpr(Tokenizer tok, Expr expr, Position startPos) returns Expr|err:Syntax {
     Token? t = tok.current();
     if t == "[" {
         Position pos = tok.currentStartPos();
+        Position accessStartPos = pos;
```
jclark: This doesn't look right. You're not using startPos. pos is getting duplicated.

compiler/modules/front.syntax/parseStmt.bal
```diff
@@ -149,13 +150,13 @@ function finishOptQualIdentifierStmt(Tokenizer tok, string? prefix, string ident
...
-        FunctionCallExpr expr = check finishFunctionCallExpr(tok, prefix, identifier, pos);
-        return finishCallStmt(tok, expr);
+        FunctionCallExpr expr = check finishFunctionCallExpr(tok, prefix, identifier, pos, startPos);
+        return finishCallStmt(tok, expr, startPos);
     }
     else if cur == "." {
-        VarRefExpr varRef = { varName: identifier, prefix };
-        MethodCallExpr expr = check finishMethodCallExpr(tok, varRef);
-        return finishCallStmt(tok, expr);
+        VarRefExpr varRef = { startPos, endPos:tok.previousEndPos(), varName: identifier, prefix };
```
jclark, line 159: Space after colon

compiler/modules/front.syntax/parseStmt.bal
```diff
@@ -165,20 +166,22 @@ function finishOptQualIdentifierStmt(Tokenizer tok, string? prefix, string ident
 }
 
 function parseMethodCallStmt(Tokenizer tok) returns MethodCallExpr|err:Syntax {
+    Position startPos = tok.currentStartPos();
     Expr expr = check startPrimaryExpr(tok);
     Token? cur = tok.current();
     if cur == "." || cur == "[" {
-        expr = check finishPrimaryExpr(tok, expr);
+        expr = check finishPrimaryExpr(tok, expr, startPos);
```
jclark, line 192: This isn't right. We now need to distinguish MethodCallStmt and MethodCallExpr since the former includes the semi-colon and the latter doesn't.

compiler/modules/front.syntax/parseStmt.bal
```diff
@@ -249,7 +255,7 @@ function parseReturnStmt(Tokenizer tok, Position startPos) returns ReturnStmt|er
     Position endPos;
     if tok.current() == ";" {
         endPos = tok.currentEndPos();
-        returnExpr = { value: () }; // ConstValueExpr
+        returnExpr = { startPos: tok.currentStartPos(), endPos, value: () }; // ConstValueExpr
```
jclark: startPos can't be right.

jclark (commented):
I don't think we have an adequate testing strategy because there are several cases where there would have been test failures if this was being tested properly.

compiler/modules/front.syntax/parseExpr.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/parseExpr.bal
heshanpadmasiri: removed pos (It was used only for error messages, now replaced with startPos)

compiler/modules/front.syntax/parseExpr.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/parseExpr.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/parseExpr.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/parseStmt.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/parseStmt.bal
heshanpadmasiri: Currently we don't have an AST node type called MethodCallStmt (we only have MethodCallExpr) should I add a MethodCallStmt that wraps the MethodCallExpr. (Current code changed such that MethodCallExpr don't include `;`)

compiler/modules/front.syntax/parseStmt.bal
heshanpadmasiri: This is the case where the optional returnExpr in returnStmt token is empty (ie. return;). Though the returnExpr is optional in grammer it is required in the current AST implementation (it seems everywhere it is used first we check if returnExpr is ConstValueExpr and value is (), and if so ignored). May I turn it to an optional?

heshanpadmasiri:
> I don't think we have an adequate testing strategy because there are several cases where there would have been test failures if this was being tested properly.

Current testing strategy has two limitations

1. Whether we are considering the tokenizers previous end position or the current end position as the actual end position is hardcoded based on the node type for expressions. Thus if I made a error of judgement I am likely to set the position wrong both at the parser and test. Currently we are checking if this end point is white space but to fully catch such issues I think we will have to hardcode what should be the value of end token.
2. If I set the start position wrong and that wrong start position is the start position of a child node that end in the same position as the parent, if both child and parent nodes can be parsed using the same parsing function there will be no errors. (This is because pareser didn't failed and end point is correct). To fix this we have to check if the both the new node we got and original nodes are the same. For this now I have added a test that convert both nodes to string and see if they are equal. Is this there a better way to see if nodes are same?

jclark:
> Currently we don't have an AST node type called MethodCallStmt (we only have MethodCallExpr) should I add a MethodCallStmt that wraps the MethodCallExpr. (Current code changed such that MethodCallExpr don't include `;`) 

Have a look at the spec. I think it's CallStmt that can wrap function or method call.

jclark:
> This is the case where the optional returnExpr in returnStmt token is empty (ie. return;). Though the returnExpr is optional in grammer it is required in the current AST implementation (it seems everywhere it is used first we check if returnExpr is ConstValueExpr and value is (), and if so ignored). May I turn it to an optional? 

Yes. Good idea.

compiler/modules/front/codeGen.bal
```diff
@@ -749,7 +749,17 @@ function unnarrowBinding(Binding binding) returns Binding {
 
 function codeGenReturnStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environment env, s:ReturnStmt stmt) returns CodeGenError|StmtEffect {
     var { returnExpr } = stmt;
-    var { result: operand, block: nextBlock } = check codeGenExpr(cx, startBlock, env, check cx.foldExpr(env, returnExpr, cx.returnType));
+    bir:BasicBlock nextBlock;
+    bir:Operand operand;
+    if returnExpr is s:Expr {
+        var codeGenExpr = check codeGenExpr(cx, startBlock, env, check cx.foldExpr(env, returnExpr, cx.returnType));
+        operand = codeGenExpr.result;
+        nextBlock = codeGenExpr.block;
```
manuranga:
```suggestion
        { result: operand, block: nextBlock } = check codeGenExpr(cx, startBlock, env, check cx.foldExpr(env, returnExpr, cx.returnType));
```

heshanpadmasiri:
> > This is the case where the optional returnExpr in returnStmt token is empty (ie. return;). Though the returnExpr is optional in grammer it is required in the current AST implementation (it seems everywhere it is used first we check if returnExpr is ConstValueExpr and value is (), and if so ignored). May I turn it to an optional?
> 
> Yes. Good idea.

fixed

compiler/modules/front/codeGen.bal
```diff
@@ -881,11 +889,13 @@ function codeGenCompoundAssignStmt(CodeGenContext cx, bir:BasicBlock startBlock,
...
     s:Expr expr;
+    s:Position startPos = rexpr.startPos;
+    s:Position endPos = rexpr.endPos;
     if op is s:BinaryArithmeticOp {
-        expr = { arithmeticOp: op, left: lValue, right: rexpr, pos: pos };
+        expr = { startPos, endPos, arithmeticOp: op, left: lValue, right: rexpr, pos: pos };
     }
     else {
-        expr = { bitwiseOp: <s:BinaryBitwiseOp> op, left: lValue, right: rexpr };
+        expr = { startPos, endPos, bitwiseOp: <s:BinaryBitwiseOp> op, left: lValue, right: rexpr };
```
jclark: Do we really need the cast here? If so, I think it's a JBUG.

jclark (commented):
This generally looks good, but there are a lot of changes, so I need to have another look tomorrow.

compiler/modules/front/codeGen.bal

compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -22,15 +25,17 @@ function parseInnerExpr(Tokenizer tok) returns Expr|err:Syntax {
...
 function parseBitwiseOrExpr(Tokenizer tok) returns Expr|err:Syntax {
+    Position startPos = tok.currentStartPos();
     Expr expr = check parseBitwiseXorExpr(tok);
     while true {
         Token? t = tok.current();
         if t == "|" {
             check tok.advance();
             Expr right = check parseBitwiseXorExpr(tok);
-            BinaryBitwiseExpr bin = { bitwiseOp: t, left: expr, right };
+            Position endPos = tok.currentEndPos();
```
jclark: This has advanced over the last token in the expression already, hasn't it? So shouldn't it be previousEndPos(). I think this applies to many other cases in this file.

jclark (changes requested):
How can our testing guarantee that we don't have this sort of problem (using currentEndPos when we should have used previousEndPos). I think we have to test when we have an binary expression E with subexpressions L and R, that the only things within E (i.e. between the startPos and endPos) but not within L and not within R are the operator and whitespace/comments.

compiler/modules/front.syntax/parseExpr.bal
heshanpadmasiri: fixed

heshanpadmasiri:
> How can our testing guarantee that we don't have this sort of problem (using currentEndPos when we should have used previousEndPos). I think we have to test when we have an binary expression E with subexpressions L and R, that the only things within E (i.e. between the startPos and endPos) but not within L and not within R are the operator and whitespace/comments.

Added two seperate functions one (`testValidInterExpressionRange`) generally validating the range between child nodes and one (`testValidExprEnd`) that check the endpos to check if it is something that can't be the endpos for a given expression

jclark (approved):
In the next pull request, can you please separate out the postion related code in testParserOnTestSuite into a testPos file?

PR #569, opened 2021-10-17
Implement field access lvalues

### LINKED ISSUE #510: Implement field access lvalues

manuranga:
Fixes #510

### DIFF
--- a/compiler/modules/front/codeGen.bal
+++ b/compiler/modules/front/codeGen.bal
@@ -825,7 +825,7 @@ function codeGenAssign(CodeGenContext cx, Environment env, bir:BasicBlock block,
     return nextBlock;
 }
 
-function codeGenAssignToMember(CodeGenContext cx, bir:BasicBlock startBlock, Environment env, s:MemberAccessLExpr lValue, s:Expr expr) returns CodeGenError|StmtEffect {
+function codeGenAssignToMember(CodeGenContext cx, bir:BasicBlock startBlock, Environment env, s:MemberAccessLExpr|s:FieldAccessLExpr lValue, s:Expr expr) returns CodeGenError|StmtEffect {
     bir:Register reg = (check lookupVarRefBinding(cx, lValue.container.varName, env)).reg;
     t:UniformTypeBitSet indexType;
     t:SemType memberType;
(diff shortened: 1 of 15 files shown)


manuranga:
This PR is done on top of #556, so please merge that one first.

jclark (commented):
Can you update? The diff doesn't make sense to me (e.g. 1152 of codeGen).

compiler/modules/front/codeGen.bal
```diff
@@ -825,8 +825,10 @@ function codeGenAssign(CodeGenContext cx, Environment env, bir:BasicBlock block,
     return nextBlock;
 }
 
-function codeGenAssignToMember(CodeGenContext cx, bir:BasicBlock startBlock, Environment env, s:MemberAccessLExpr lValue, s:Expr expr) returns CodeGenError|StmtEffect {
-    bir:Register reg = (check lookupVarRefBinding(cx, lValue.container.varName, env)).reg;
+function codeGenAssignToMember(CodeGenContext cx, bir:BasicBlock startBlock, Environment env, s:MemberAccessLExpr|s:FieldAccessLExpr lValue, s:Expr expr) returns CodeGenError|StmtEffect {
+    // JBUG #33270 needs parentheses
+    s:VarRefExpr mappingRef = lValue is s:FieldAccessLExpr ? (lValue.mapping) : lValue.container;
```
jclark: We should fix FieldAccessLExpr to use `container` rather than `mapping`, so we don't need the condition.

compiler/modules/front/codeGen.bal
```diff
@@ -1160,6 +1173,19 @@ function codeGenMappingGet(CodeGenContext cx, bir:BasicBlock block, bir:Register
     return { result, block };
 }
 
+function codeGenMappingKey(CodeGenContext cx, bir:BasicBlock block, Environment env, s:MemberAccessLExpr|s:FieldAccessLExpr mappingLValue, t:SemType mappingType) returns CodeGenError|StringExprEffect {
```
jclark: I would like to name this so it's clear it is dealing with lvalues. Maybe codeGenLExprMappingKey?

docs/subset09.md
```diff
@@ -353,6 +356,7 @@ Two kinds of `import` are supported.
 * Nil type descriptor
 * Proper parsing of type descriptors in statements: all type descriptors that are allowed in a type definition are now allowed within statements
 * Filed access expression
+* Field access lvalue
```
jclark: I would put these two field access items as subitems under closed record types.

These two also need to be mentioned in the subset summary at the top.

jclark (changes requested):

compiler/modules/front/codeGen.bal
manuranga: Should `FieldAccessExpr` also be changed for consistency ?

jclark: Yes, good idea.

PR #570, opened 2021-10-17
Fix wildcard destructuring assignment bug

### LINKED ISSUE #568: Parser bug in wildcard destructuring assignment statement
Following complies.
```
public function main() {
    _ equals 1;
}
```

manuranga:
Fixes #568


jclark (approved):
Good catch!

PR #585, opened 2021-10-18
Re-inline test function that used to be too large

manuranga:
Related to #548

Please close without merging if this is unnecessary.


compiler/modules/front.syntax/tests/testParser.bal
```diff
@@ -189,88 +189,7 @@ function invalidTokenSourceFragments() returns map<TokenizerTestCase>|error {
 // JBUG moved to a separate method due to method too large, move back to validTokenSourceFragments after beta3
```
jclark: Should remove JBUG line.

manuranga: Ahg.

jclark (approved):

PR #588, opened 2021-10-19
Resolve JBUGs in enumerable.bal

KavinduZoysa:
## Purpose
Related to #550


compiler/modules/types/enumerable.bal
```diff
@@ -172,13 +172,13 @@ function compareEnumerable(EnumerableType v1, EnumerableType v2) returns Order {
     else {
         float f2 = <float>v2;
         // JBUG: #17977 can't use `==`
```
manuranga: Needs to be removed

KavinduZoysa: Yes, removed

jclark (approved):

PR #590, opened 2021-10-19
Limit where widening can occur in BIR

### LINKED ISSUE #508: BIR verification should limit where widening can occur
BIR verification should limit where widening can occur to be consistent with https://github.com/ballerina-platform/nballerina/issues/204#issuecomment-917780707

This means that in several cases it needs to switch from checking for a subtype relationship to checking for an equivalence relationship.  

Equivalence between two subtypes should be tested by first using `==` on the SemType and then, if that fails, by checking whether each is a subtype of the other.

manuranga:
Fixes #508

### DIFF
--- a/compiler/modules/bir/verify.bal
+++ b/compiler/modules/bir/verify.bal
@@ -21,6 +21,10 @@ class VerifyContext {
         return t:isSubtype(self.tc, s, t);
     }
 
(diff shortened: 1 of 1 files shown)


jclark:
This is fine. Could you also add XXX comments in places where we are not verifying the insn result type but should be? Looks like MappingConstruct, ListConstruct and Call

PR #593, opened 2021-10-20
Add positions to Type descriptor AST nodes

### LINKED ISSUE #513: Add positions to AST nodes

heshanpadmasiri:
## Purpose
Add positions to Type descriptor AST nodes
Resolves #513

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/front.syntax/ast.bal
```diff
@@ -426,42 +429,57 @@ public type FieldDesc record {|
...
+    *PositionFields;
     (string|float|int|boolean|decimal) value;
 |};
 
-// This is the subtype of BuiltinTypeDesc that we currently allow inline.
-public type InlineBuiltinTypeDesc "boolean"|"int"|"float"|"string"|"error"|"any";
-public type BuiltinTypeDesc InlineBuiltinTypeDesc|"byte"|"decimal"|"handle"|"json"|"never"|"readonly"|"typedesc"|"xml"|"()";
+public type BuiltinType "boolean"|"int"|"float"|"string"|"error"|"any"|"byte"|"decimal"|"handle"|"json"|"never"|"readonly"|"typedesc"|"xml"|"()";
+
+public type BuiltinTypeDesc record {|
```
manuranga: Why can't `BuiltinTypeDesc` be `"boolean"|"int"|"float"|"string"|"error"|"any"|"byte"|"decimal"|"handle"|"json"|"never"|"readonly"|"typedesc"|"xml"|"()"` ?

jclark: Because it needs to have a position.

compiler/modules/front.syntax/ast.bal
```diff
@@ -426,42 +429,57 @@ public type FieldDesc record {|
...
 
 public type SingletonTypeDesc record {|
+    *PositionFields;
     (string|float|int|boolean|decimal) value;
 |};
 
-// This is the subtype of BuiltinTypeDesc that we currently allow inline.
-public type InlineBuiltinTypeDesc "boolean"|"int"|"float"|"string"|"error"|"any";
-public type BuiltinTypeDesc InlineBuiltinTypeDesc|"byte"|"decimal"|"handle"|"json"|"never"|"readonly"|"typedesc"|"xml"|"()";
+public type BuiltinType "boolean"|"int"|"float"|"string"|"error"|"any"|"byte"|"decimal"|"handle"|"json"|"never"|"readonly"|"typedesc"|"xml"|"()";
```
jclark: I don't like `BuiltinType` here. Let's call it `BuiltinTypeName`, but change `"()"` to `"null"` (since both `null` is equivalent to `()`).

compiler/modules/front.syntax/ast.bal
manuranga: Ahh Ok.

compiler/modules/front.syntax/ast.bal
```diff
@@ -426,42 +429,57 @@ public type FieldDesc record {|
...
+
+public type BuiltinTypeDesc record {|
+    *PositionFields;
+    BuiltinType builtinType;
```
jclark: Then use builtinTypeName for the field name

compiler/modules/front.syntax/parse.bal
```diff
@@ -166,10 +166,12 @@ function parseTypeDefinition(Tokenizer tok, ModulePart part, Visibility vis, Pos
...
     Token? t = tok.current();
-    InlineBuiltinTypeDesc? td = ();
-    if t is InlineBuiltinTypeDesc {
+    BuiltinTypeDesc? td = ();
+    if t is BuiltinType {
+        Position tdStartPos = tok.currentStartPos();
+        Position tdEndPos = tok.currentEndPos();
         check tok.advance();
-        td = t;
+        td = { startPos: tdStartPos, endPos: tdEndPos, builtinType:t };
```
jclark: Space after colon

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -162,9 +193,9 @@ function parseNumericLiteralTypeDesc(Tokenizer tok, Position? signPos = ()) retu
                 if signPos != () {
                     value = -value;
                 }
-                return { value };
+                return { startPos, endPos:expr.endPos, value };
```
jclark: Space after colon

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -175,9 +206,9 @@ function parseNumericLiteralTypeDesc(Tokenizer tok, Position? signPos = ()) retu
...
```
jclark: Space after colon

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -195,7 +226,7 @@ function parseNumericLiteralTypeDesc(Tokenizer tok, Position? signPos = ()) retu
             else {
                 value = -n;
             }
-            return { value };
+            return { startPos, endPos:expr.endPos, value };
```
jclark: Space after colon

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -244,16 +277,19 @@ function parseFunctionTypeDesc(Tokenizer tok, string[]? paramNames = ()) returns
         ret = check parseTypeDesc(tok);
     }
     else {
-        ret = "()";
+        ret = { startPos:tok.currentStartPos(), endPos, builtinType:"()" };
```
jclark: Space after colon

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -303,6 +340,7 @@ function parseRecordTypeDesc(Tokenizer tok) returns MappingTypeDesc|err:Syntax {
         }
         check tok.expect(";");
     }
+    Position endPos = tok.currentEndPos();
     check tok.advance();
-    return { fields, rest: rest ?: "never" };
+    return { startPos, endPos, fields, rest: rest ?: { startPos, endPos:startPos, builtinType:"never" } };
```
jclark: Space after colon

rest bit looks bogus. Can we make it optional in the AST?

jclark (commented):
I haven't reviewed this fully yet.

compiler/modules/front.syntax/ast.bal

compiler/modules/front.syntax/ast.bal

compiler/modules/front.syntax/parse.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/parseTypeDesc.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/parseTypeDesc.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/parseTypeDesc.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/parseTypeDesc.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/parseTypeDesc.bal
heshanpadmasiri: fixed. Made rest optional in the ast

compiler/modules/front.syntax/ast.bal
```diff
@@ -426,42 +429,57 @@ public type FieldDesc record {|
...
-public type InlineBuiltinTypeDesc "boolean"|"int"|"float"|"string"|"error"|"any";
-public type BuiltinTypeDesc InlineBuiltinTypeDesc|"byte"|"decimal"|"handle"|"json"|"never"|"readonly"|"typedesc"|"xml"|"()";
+public type BuiltinTypeName "boolean"|"int"|"float"|"string"|"error"|"any"|"byte"|"decimal"|"handle"|"json"|"never"|"readonly"|"typedesc"|"xml"|"null";
+
+public type BuiltinTypeDesc record {|
+    *PositionFields;
+    BuiltinTypeName builtinTypeName;
+|};
 
 public type LeafTypeDesc BuiltinTypeDesc;
```
jclark: Do we still need LeafTypeDesc?  I think I was using LeafTypeDesc more likle BuiltinTypeName rather than BuiltinTypeDesc.

compiler/modules/front.syntax/ast.bal
```diff
@@ -426,42 +429,57 @@ public type FieldDesc record {|
...
     Position pos;
 |};
 
 public type SingletonTypeDesc record {|
+    *PositionFields;
     (string|float|int|boolean|decimal) value;
 |};
 
-// This is the subtype of BuiltinTypeDesc that we currently allow inline.
-public type InlineBuiltinTypeDesc "boolean"|"int"|"float"|"string"|"error"|"any";
```
jclark, line 464: Simply removing this rid of this doesn't quite work.  We are relying on this in parseConstDefinition.  The subset grammar has

```
const-defn = ["public"] "const" [builtin-type-name] identifier "=" const-expr ";"
builtin-type-name = "any" | "boolean" | "int" | "float" | "string" | "error"
```

With this pull request as is, the code will crash in resolveBuiltinTypeDesc if we use something like `json` in a const definition.


So let's fix this as follows. Keep InlineBuiltinTypeDesc but rename SubsetBuiltinTypeDesc, and add a comment that it is used for ConstDefinition.  Use it in ConstDefinition and resolveBuiltinTypeDesc.

compiler/modules/front.syntax/parseStmt.bal
```diff
@@ -79,7 +79,9 @@ function parseStmt(Tokenizer tok) returns Stmt|err:Syntax {
             // JBUG cast
             return finishCheckingCallStmt(tok, <CheckingKeyword>cur, startPos);
         }
-        var td if td is InlineBuiltinTypeDesc|"map"|"record" => {
+        var td if td is "boolean"|"int"|"float"|"string"|"error"|"any"|"byte"|
```
jclark: No. We need to parse the subset grammar: this would be accepting things not in the subset. Use SubsetBuiltinTypeName here.

jclark (changes requested):
Turns out we still need a type corresponding to InlineBuiltinTypeDesc to distinguish types accepted in our subset.

compiler/modules/front.syntax/ast.bal
heshanpadmasiri: LeafTypeDesc removed. (replaced by BuilinTypeDesc)

compiler/modules/front.syntax/ast.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/parseStmt.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -10,39 +10,46 @@ function parseTypeDesc(Tokenizer tok) returns TypeDesc|err:Syntax {
...
 
 function parsePostfixTypeDesc(Tokenizer tok) returns TypeDesc|err:Syntax {
+    Position startPos = tok.currentStartPos();
     TypeDesc td = check parsePrimaryTypeDesc(tok);
     while true {
         if tok.current() == "?" {
+            Position endPos = tok.currentEndPos();
             check tok.advance();
-            BinaryTypeDesc bin =  { op: "|", left: td, right: "()" };
+            BinaryTypeDesc bin =  { startPos, endPos, op: "|", left: td,
```
jclark: I prefer to format multi-line constructors like this:

```
BinaryTypeDesc bin =  {
    startPos,
    endPos, 
    op: "|",
    left: td,
    right: { startPos: endPos, endPos, builtinTypeName: "null"}
}; 
```

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -56,16 +63,27 @@ function parsePostfixTypeDesc(Tokenizer tok) returns TypeDesc|err:Syntax {
...
+                endPos = tok.currentEndPos();
                 check tok.advance();
-                return "()";
+                return { startPos, endPos, builtinTypeName: "null" };
             }
             TypeDesc td = check parseTypeDesc(tok);
-            check tok.expect(")");
-            return td;
+            endPos = check tok.expectEnd(")");
+            if td is BuiltinTypeDesc {
```
jclark: What's going on here? This doesn't look right.

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -56,16 +63,27 @@ function parsePostfixTypeDesc(Tokenizer tok) returns TypeDesc|err:Syntax {
...
+                // td is readonly
+                return {startPos, endPos, builtinTypeName: td.builtinTypeName};
```
jclark: Spacing

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -77,50 +95,63 @@ function parsePrimaryTypeDesc(Tokenizer tok) returns TypeDesc|err:Syntax {
...
             if tok.current() != ":" {
-                return <LeafTypeDesc> cur;
+                // JBUG should not need cast #30191
+                return { startPos, endPos, builtinTypeName: <BuiltinTypeName>cur };
             }
             check tok.advance();
-            return { prefix: <LeafTypeDesc> cur, typeName: check tok.expectIdentifier(), pos };
+            string typeName = check tok.expectIdentifier();
+            endPos = tok.previousEndPos();
+            return { startPos, endPos, prefix: <BuiltinTypeName>cur, typeName, pos };
```
jclark: Type of `prefix` is string, so better to cast to `<string>`

jclark (changes requested):

compiler/modules/front.syntax/parseTypeDesc.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/parseTypeDesc.bal
heshanpadmasiri: This is parsing something of the form `basic-type-desc ="(" union-type-desc ")"`. So I need to set the endPos to ")". `td` get's endPoint before the ")" (at the end of `union-type-desc`) so I am updating the endPos of td here. However I have made `BuiltinTypeDesc` readonly (to make `SubsetBuiltinTypeDesc` subtype of it) so I can't update the endPos field here if I get a `BuiltinTypeDesc` as `td`. Therefore I am making a new copy of the `td` with endPos having the correct value.

compiler/modules/front.syntax/parseTypeDesc.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/parseTypeDesc.bal
heshanpadmasiri: fixed

jclark (changes requested):
> This is parsing something of the form `basic-type-desc ="(" union-type-desc ")"`. So I need to set the endPos to ")". `td` get's endPoint before the ")" (at the end of `union-type-desc`) so I am updating the endPos of td here. However I have made `BuiltinTypeDesc` readonly (to make `SubsetBuiltinTypeDesc` subtype of it) so I can't update the endPos field here if I get a `BuiltinTypeDesc` as `td`. Therefore I am making a new copy of the `td` with endPos having the correct value. 

The underlying problem here is that we are not representing parenthesized type descriptors explicitly in the AST. If we have something like `(int)` then there are two type-descriptors with distinct start/end positions: `int` and `(int)`. To deal with this cleanly, we need to introduce another kind of TypeDesc, say ParenTypeDesc. (I suspect we have a similar problem with expressions and should introduce ParenExpr also. You can deal with this in a separate follow on pull request.)

heshanpadmasiri:
> > This is parsing something of the form `basic-type-desc ="(" union-type-desc ")"`. So I need to set the endPos to ")". `td` get's endPoint before the ")" (at the end of `union-type-desc`) so I am updating the endPos of td here. However I have made `BuiltinTypeDesc` readonly (to make `SubsetBuiltinTypeDesc` subtype of it) so I can't update the endPos field here if I get a `BuiltinTypeDesc` as `td`. Therefore I am making a new copy of the `td` with endPos having the correct value.
> 
> The underlying problem here is that we are not representing parenthesized type descriptors explicitly in the AST. If we have something like `(int)` then there are two type-descriptors with distinct start/end positions: `int` and `(int)`. To deal with this cleanly, we need to introduce another kind of TypeDesc, say ParenTypeDesc. (I suspect we have a similar problem with expressions and should introduce ParenExpr also. You can deal with this in a separate follow on pull request.)

Removed extending start and end pos to cover parenthesis. Yes we have the same issue with expressions as well

jclark:
So you're going to do ParenTypeDesc as well as ParenExpr in a separate pull request?

heshanpadmasiri:
> So you're going to do ParenTypeDesc as well as ParenExpr in a separate pull request?

Yes

Issue #580, opened 2021-10-18
Make buildHasListType work like buildHasMappingType


jclark:
buildHasMappingType works by calling a function pointer in a structure created by init.bal; buildHasListType should work the same way.

jclark:
When we've changed, we can merge buildHasMappingType and buildHasListType into a single function buildHasComplexSemType.

PR #594, opened 2021-10-20
Make list values start with a descriptor ponter

### LINKED ISSUE #582: Make list values start with a descriptor pointer
The runtime representation of  a list should start with a pointer to a ListDescriptor, in a similar way to a mapping.  Initially, the format of a ListDescriptor would be simply a uint32_t containing a bitSet. (Currently it stores the bitSet where the pointer should be.)

This is a prerequisite for #329.

manuranga:
Fixes #582

### DIFF
--- a/compiler/modules/nback/typeTest.bal
+++ b/compiler/modules/nback/typeTest.bal
@@ -6,7 +6,7 @@ final RuntimeFunction listHasTypeFunction = {
     name: "list_has_type",
     ty: {
         returnType: "i1",
-        paramTypes: [LLVM_TAGGED_PTR, "i64"]
+        paramTypes: [LLVM_TAGGED_PTR, LLVM_BITSET]
     },
     attrs: ["readonly"]
 };
@@ -15,7 +15,7 @@ final RuntimeFunction listExactifyFunction = {
     name: "list_exactify",
     ty: {
         returnType: LLVM_TAGGED_PTR,
-        paramTypes: [LLVM_TAGGED_PTR, "i64"]
+        paramTypes: [LLVM_TAGGED_PTR, llvm:pointerType(llInherentType)]
     },
     attrs: ["readonly"]
 };
@@ -152,7 +152,7 @@ function buildMappingExactify(llvm:Builder builder, Scaffold scaffold, llvm:Poin
 function buildHasListType(llvm:Builder builder, Scaffold scaffold, llvm:PointerValue tagged, t:SemType targetType) returns llvm:Value {
     t:UniformTypeBitSet bitSet = <t:UniformTypeBitSet>t:simpleArrayMemberType(scaffold.typeContext(), targetType);
     return <llvm:Value>builder.call(scaffold.getRuntimeFunctionDecl(listHasTypeFunction),
-                                    [tagged, llvm:constInt(LLVM_INT, bitSet)]);     
+                                    [tagged, llvm:constInt(LLVM_BITSET, bitSet)]);
 }
 
 function buildListExactify(llvm:Builder builder, Scaffold scaffold, llvm:PointerValue tagged, t:SemType targetType) returns llvm:PointerValue {
@@ -163,7 +163,7 @@ function buildListExactify(llvm:Builder builder, Scaffold scaffold, llvm:Pointer
     }
     else {
         return <llvm:PointerValue>builder.call(scaffold.getRuntimeFunctionDecl(listExactifyFunction),
-                                               [tagged, llvm:constInt(LLVM_INT, bitSet)]);   
+                                               [tagged, scaffold.getInherentType(t:intersect(targetType, t:LIST_RW))]);
     }
 }
 
--- a/runtime/balrt.h
+++ b/runtime/balrt.h
@@ -79,10 +79,12 @@ typedef struct {
     GC TaggedPtr *members;
 } TaggedPtrArray;
 
-typedef int64_t ListDesc;
+typedef struct {
+    uint32_t bitSet;
+} ListDesc, *ListDescPtr;
 
 typedef GC struct List {
-    ListDesc desc;
+    ListDescPtr desc;
     // This isn't strictly portable because void* and TaggedPtr* might have different alignments/sizes
     // But we ain't writing portable code here
     union {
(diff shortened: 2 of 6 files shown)


runtime/list.c
```diff
@@ -34,7 +34,7 @@ static bool getFiller(ListDesc desc, TaggedPtr *valuePtr) {
 
 PanicCode _bal_list_set(TaggedPtr p, int64_t index, TaggedPtr val) {
     ListPtr lp = taggedToPtr(p);
-    if ((lp->desc & (1 << (getTag(val) & UT_MASK))) == 0) {
+    if ((*(lp->desc) & (1 << (getTag(val) & UT_MASK))) == 0) {
```
jclark: Can we make this consistent with mapping.c and use a temp

```
ListDescPtr ldp = lp->desc;
```

runtime/balrt.h
```diff
@@ -79,10 +79,10 @@ typedef struct {
     GC TaggedPtr *members;
 } TaggedPtrArray;
 
-typedef int64_t ListDesc;
+typedef uint32_t ListDesc, *ListDescPtr;
```
jclark: Can we make this consistent with Mapping and use a struct containing just a uint32_t?

jclark (changes requested):
Great. Just a few tweaks to improve consistency with mapping.

manuranga:
I should also change `buildListExactify` and `buildHasListType` to use `getInherentType`.

jclark:
> I should also change `buildListExactify` and `buildHasListType` to use `getInherentType`.

No.

jclark:
Changing buildHasListType is #580.

runtime/list.c
```diff
@@ -53,7 +55,7 @@ PanicCode _bal_list_set(TaggedPtr p, int64_t index, TaggedPtr val) {
         // we have a gap to fill
         // from length..<index
         TaggedPtr filler;
-        if (!getFiller(lp->desc, &filler)) {
+        if (!getFiller(*(lp->desc), &filler)) {
```
jclark: Should use ldp.  Also better to make getFiller take a ListDescPtr.

jclark (changes requested):
Also I was incorrect about buildListExactify: better to change that also.

runtime/list.c
```diff
@@ -53,7 +55,8 @@ PanicCode _bal_list_set(TaggedPtr p, int64_t index, TaggedPtr val) {
         // we have a gap to fill
         // from length..<index
         TaggedPtr filler;
-        if (!getFiller(lp->desc, &filler)) {
+        ListDescPtr ldp = lp->desc;
```
jclark: Duplicate declaration

runtime/list_inline.c
```diff
@@ -24,12 +26,14 @@ bool _bal_list_has_type(TaggedPtr p, ListDesc desc) {
         return false;
     }
     ListPtr lp = taggedToPtr(p);
-    return (lp->desc & ~desc) == 0;
+    ListDescPtr ldp = lp->desc;
+    return (ldp->bitSet & ~desc.bitSet) == 0;
 }
 
 TaggedPtr _bal_list_exactify(TaggedPtr p, ListDesc desc) {
```
jclark: There's a mismatch between how these (has_type and exactify) are called and how they are defined. They are called from generated code with a uint32_t, but their arguments are declared here as a struct.

jclark (changes requested):

runtime/list_inline.c
manuranga: Ya, that is why I was thinking of changing  `buildListExactify` and `buildHasListType` to use `getInherentType`. I'll  bitcast and pass until then.

manuranga: I need to pass a struct on stack.

In LLVM IR terms, what I need to do is one of
1) `%x = call _bal_list_exactify(i8* %list, { i32 } { i32 8386559} )`
2) `%y = bitcast i32 %a to { i32 }` and `%x = call _bal_list_exactify( i8* %list, { i32 } %y )`
3) `%y = alloca { i32 }` followed by a GEP and load
I don't think any of these are supported yet. I am trying to figure out what would be simplest to implement.

@heshanpadmasiri any suggestions?

jclark: No need. Either use

- getInherentType and pass a pointer; declare desc as ListDescPtr
- leave bal code unchanged and change desc parameter to uint32_t

heshanpadmasiri: If the values of the struct is known at the compile time in theory you should be able to use the already implemented constStruct and pass the return value to call as an arg.  (to get 1) If not (ie. you only have the struct type) then extending the bitcast to deal with struct type (2) will be the easiest (just allowing the argument to be struct type should be sufficient because rest is type to string and it can already handle struct types)

compiler/modules/nback/typeTest.bal
```diff
@@ -163,7 +163,7 @@ function buildListExactify(llvm:Builder builder, Scaffold scaffold, llvm:Pointer
     }
     else {
         return <llvm:PointerValue>builder.call(scaffold.getRuntimeFunctionDecl(listExactifyFunction),
-                                               [tagged, llvm:constInt(LLVM_INT, bitSet)]);   
+                                               [tagged, scaffold.getInherentType(targetType)]);
```
manuranga: Do I need to intersect `targetType` with `t:LIST`?

manuranga: Mapping exactify does that, and I am not sure why.

manuranga: Mapping exactify does that, and I am not sure why.

manuranga: Mapping exactify does that, and I am not sure why.

jclark: Not yet. When we add support for unions of structures, we will need to do things a bit differently.

jclark: Actually, it is necessary: what it's doing is removing the readonly part of the type, because init.bal is expecting inherent types to be purely RW.

manuranga: Updated.

jclark (approved):

PR #611, opened 2021-10-21
Resolve JBUGs in err.bal

KavinduZoysa:
## Purpose
Related to #550

### DIFF
--- a/compiler/modules/err/err.bal
+++ b/compiler/modules/err/err.bal
@@ -38,7 +38,6 @@ public type Detail record {
 public type Any Syntax|Semantic|Unimplemented;
 
 // Grammatical inconsistency here is reluctantly intentional
-// JBUG defining these as `error<Detail> & distinct error` results in these errors not being distinct
 public type Syntax distinct error<Detail>;
 public type Semantic distinct error<Detail>;
 public type Unimplemented distinct error<Detail>;


compiler/modules/err/err.bal
```diff
@@ -37,11 +37,9 @@ public type Detail record {
 
 public type Any Syntax|Semantic|Unimplemented;
 
-// Grammatical inconsistency here is reluctantly intentional
```
jclark, line 40: Don't remove this line

compiler/modules/err/err.bal
```diff
@@ -37,11 +37,9 @@ public type Detail record {
...
-// JBUG defining these as `error<Detail> & distinct error` results in these errors not being distinct
-public type Syntax distinct error<Detail>;
```
jclark, line 42: No need to change these lines. Just remove the JBUG line

jclark (changes requested):

compiler/modules/err/err.bal
KavinduZoysa: Added.

compiler/modules/err/err.bal
KavinduZoysa: Changed

Issue #203, opened 2021-07-11
Make calculation of sizes of structs and arrays on the heap be portable


jclark:
At the moment `typeSize` in build.bal makes non-portable assumptions about data layout.  We should do this right.

PR #624, opened 2021-10-23
Move list construction logic to C

### LINKED ISSUE #609: Move the logic in`buildListConstruct` to a new `list_construct` function in C
Move the logic in`buildListConstruct` to new `list_construct` function in C
Get rid of `buildUntypedAlloc`

manuranga:
Fixes  #609, #203

### DIFF
--- a/compiler/modules/nback/structure.bal
+++ b/compiler/modules/nback/structure.bal
@@ -28,6 +28,16 @@ final RuntimeFunction mappingIndexedSetFunction = {
     },
     attrs: []
 };
+
+final RuntimeFunction listConstructFunction = {
+    name: "list_construct",
+    ty: {
+        returnType: heapPointerType(llListType),
+        paramTypes: [llvm:pointerType(llInherentType), LLVM_INT]
+    },
+    attrs: []
+};
+
 final RuntimeFunction mappingGetFunction = {
     name: "mapping_get",
     ty: {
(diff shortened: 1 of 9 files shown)


compiler/modules/nback/structure.bal
```diff
@@ -67,41 +86,22 @@ final RuntimeFunction mappingConstructFunction = {
...
-        builder.store(llvm:constInt(LLVM_INT, length),
-                      builder.getElementPtr(struct, [llvm:constInt(LLVM_INT, 0), llvm:constInt(LLVM_INDEX, i)], "inbounds"));
+    t:SemType listType = insn.result.semType;
+    llvm:ConstPointerValue inherentType = scaffold.getInherentType(listType);
+    llvm:PointerValue list = <llvm:PointerValue>builder.call(scaffold.getRuntimeFunctionDecl(listConstructFunction),
+                                                             [inherentType, llvm:constInt(LLVM_INT, length)]);
+    // Cases that are not UniformTypeBitSet should have been filtered out before
+    t:UniformTypeBitSet memberType = <t:UniformTypeBitSet>t:simpleArrayMemberType(scaffold.typeContext(), listType);
+    foreach int i in 0 ..< length {
+        _ = builder.call(scaffold.getRuntimeFunctionDecl(listInitMemberFunction),
```
jclark: I don't see why it's an improvement to do this with a function call.

runtime/container.h
```diff
@@ -0,0 +1,12 @@
+#include "balrt.h"
+
+static void initArray(GC GenericArray *ap, int64_t capacity, int shift) {
```
jclark: Why can't this be inline and go in balrt.h (with a suitable name)?

runtime/balrt.h
```diff
@@ -542,3 +543,14 @@ static READONLY inline int64_t stringArrayCompare(TaggedPtr lhs, TaggedPtr rhs)
 static READONLY inline int64_t booleanArrayCompare(TaggedPtr lhs, TaggedPtr rhs) {
     return arrayCompare(lhs, rhs, &taggedBooleanCompare);
 }
+
+static inline void initMemberArray(GC GenericArray *ap, int64_t capacity, int shift) {
```
jclark: Better to call this initGenericArray (since the type is GenericArray)

compiler/modules/nback/structure.bal
```diff
@@ -67,47 +77,41 @@ final RuntimeFunction mappingConstructFunction = {
...
-    final llvm:Type unsizedArrayType = llvm:arrayType(LLVM_TAGGED_PTR, 0);
-    final llvm:PointerType ptrUnsizedArrayType = heapPointerType(unsizedArrayType);
-    final llvm:Type structType = llvm:structType([llvm:pointerType(llInherentType), LLVM_INT, LLVM_INT, ptrUnsizedArrayType]);
-    // Cases that are not UniformTypeBitSet should have been filtered out before
-    t:UniformTypeBitSet memberType = <t:UniformTypeBitSet>t:simpleArrayMemberType(scaffold.typeContext(), insn.result.semType);
     final int length = insn.operands.length();
-    llvm:PointerValue array;
+    t:SemType listType = insn.result.semType;
+    llvm:ConstPointerValue inherentType = scaffold.getInherentType(listType);
+    llvm:PointerValue struct = <llvm:PointerValue>builder.call(scaffold.getRuntimeFunctionDecl(listConstructFunction),
```
jclark, line 83: listConstruct is or'ing in flags which you then immediately remove with the ptrmask.  Better for listConstruct to return a ListPtr and then use buildTaggedPtr at the end, as before.

jclark (changes requested):

compiler/modules/nback/structure.bal
```diff
@@ -67,53 +77,39 @@ final RuntimeFunction mappingConstructFunction = {
...
-    foreach int i in 1 ..< 3 {
         builder.store(llvm:constInt(LLVM_INT, length),
-                      builder.getElementPtr(struct, [llvm:constInt(LLVM_INT, 0), llvm:constInt(LLVM_INDEX, i)], "inbounds"));
+                      builder.getElementPtr(struct, [llvm:constInt(LLVM_INT, 0), llvm:constInt(LLVM_INDEX, 1)], "inbounds"));
     }
-    builder.store(array,
-                  builder.getElementPtr(struct, [llvm:constInt(LLVM_INT, 0), llvm:constInt(LLVM_INDEX, 3)], "inbounds"));
-    // Don't need to convert here
-    builder.store(buildTaggedPtr(builder, structMem, TAG_LIST_RW|FLAG_EXACT), scaffold.address(insn.result));
+    builder.store(buildTaggedPtr(builder, builder.bitCast(struct, LLVM_TAGGED_PTR), TAG_LIST_RW|FLAG_EXACT), scaffold.address(insn.result));
```
manuranga, line 104: It would have been slightly better to do

```suggestion
    builder.store(builder.bitCast(buildTaggedPtr(builder, struct, TAG_LIST_RW|FLAG_EXACT), LLVM_TAGGED_PTR), scaffold.address(insn.result));
```

but that gives the wrong result. I think this is because `buildTaggedPtr` uses `getElementPtr` to do pointer calculation and it only works when it is a `i8*`.

jclark (approved):

PR #635, opened 2021-10-26
Resolve JBUGs in const.bal

KavinduZoysa:
## Purpose
Related to #550

### DIFF
--- a/compiler/modules/err/err.bal
+++ b/compiler/modules/err/err.bal
@@ -5,7 +5,7 @@ public type Message string|Template;
 public type Template object {
     *object:RawTemplate;
     public (readonly & string[]) strings;
-    public (string|int)[] insertions;
+    public (string|int|float)[] insertions;
 };
 
 const QUOTE = "'";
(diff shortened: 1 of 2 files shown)


compiler/modules/err/err.bal
```diff
@@ -5,7 +5,7 @@ public type Message string|Template;
 public type Template object {
     *object:RawTemplate;
     public (readonly & string[]) strings;
-    public (string|int)[] insertions;
+    public (string|int|float)[] insertions;
```
jclark, line 8: Why is this changed?

jclark (changes requested):

compiler/modules/err/err.bal
KavinduZoysa: Because, at https://github.com/ballerina-platform/nballerina/blob/64b296e57ffe1078106fc4f38109507aeaa87f73/compiler/modules/front/const.bal#L423, type of`value` is float and float is not available in this union `(string|int)[] insertions`.

jclark (approved):

PR #641, opened 2021-10-27
Parse inline type refs more generally

### LINKED ISSUE #640: Parse all type descriptors starting with an identifier in local var decl statements
future/fieldexpr1-uv.bal is giving a parse error, which is not right.

manuranga:
Fixes #640


compiler/modules/front.syntax/parseStmt.bal
```diff
@@ -178,6 +185,12 @@ function finishOptQualIdentifierStmt(Tokenizer tok, string? prefix, string ident
         }
         // falls through to end
     }
+    // SUBSET no module level lists, so `x:y[` implies a td
+    else if cur is "|" | "?" | "[" {
```
jclark, line 189: We will need `&` also.

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -9,6 +9,10 @@ function parseTypeDesc(Tokenizer tok) returns TypeDesc|err:Syntax {
     return parseUnion(tok);
 }
 
+function finishTypeDesc(Tokenizer tok, TypeDesc first, Position startPos) returns TypeDesc|err:Syntax {
+    return check finishUnion(tok, first, startPos);
```
jclark, line 13: Are we going to need to do something different here based on what token we got? For example

```
A&B|C&D x = y;
```

I think we are going to have to do finishIntersection, and then feed the result of that into finishUnion.

Fortunately type desc grammar is not too complicated

manuranga: Current logic parses `A&B|C&D x = 1;` correctly (once I add `"&"` as per other comment).

manuranga:
Closing in favor of  #648

Issue #636, opened 2021-10-26
Make the build pass with debug information enabled for the runtime


manuranga:
Build should pass successfully with  `make CFLAGS="-g" test`.

Currently it breaks in two places
1. There is a test failure in runtime, at `compareBacktrace`
2. LLVM linking fails with `linking module flags 'Debug Info Version': IDs have conflicting behaviors`

KavinduZoysa:
This is the reason for runtime test failure.

`CFLAGS="-g"` overrides the -O2 and now backtrace has more trace lines than the previous one. 
Printing these trace lines is controlled by the parameter `start` as shown below.
https://github.com/ballerina-platform/nballerina/blob/64b296e57ffe1078106fc4f38109507aeaa87f73/runtime/error.c#L143

Since we have more trace lines, the assertion fails in the test.
But if we use `CFLAGS="-g -O2"` runtime tests pass.

jclark:
Can we make the test work even when -O2 is not specified?

KavinduZoysa:
Yes, we change the value of `start` parameter to skip unnecessary trace lines.

KavinduZoysa:
The reason for the 2nd error is,

The behaviour flag for `"Debug Info Version"` is different. With the -g flag, it creates behaviour flag 2 (warning) and our compiler produces behaviour flag 1(error). As mentioned [here](https://releases.llvm.org/12.0.0/docs/LangRef.html#module-flags-metadata), when linking, the same flag ID cannot have multiple behaviours except Require and Override behaviours.

jclark:
OK, I understand now.

KavinduZoysa:
@jclark, can we close this issue?

jclark:
Have we fixed point 1?

KavinduZoysa:
@jclark, as explained in https://github.com/ballerina-platform/nballerina/issues/636#issuecomment-952707688, the reason for this failure is non-optimized code has more trace lines than expected. 
Do we need to change the test to pass for both cases (-g -O2 and -O2)?

jclark:
Yes.

jclark:
On Ubuntu 21.10, `make test` in runtime fails with segfaults in error_backtrace1 and failure in error_backtrace2.

The code is broken: cmpTrace is not being passed the length of the expectedTrace array, so it can access memory past the end of the array if the traces don't match.

We need a more robust testing strategy.

jclark:
valgrind output
```
[...]

PR #642, opened 2021-10-27
Change behaviour flag of `Debug Info Version` module flag

KavinduZoysa:
## Purpose
The second error explained in #636 happens because the behaviour flags are different for `Debug Info Version` for the LLVM IRs generated from the source file and runtime.

This PR changes the behaviour flag of the ballerina source


manuranga:
Are the debug info versions different in runtime vs genarated ll files? If so can we make them the same?

jclark:
Yeah, I don't understand why the versions are different in the first place.  Maybe the version we are emitted in nback is mismatched with what LLVM 12 is using. If so, we should just fix to match.

KavinduZoysa:
Here the debug versions are not different(In both cases debug version is 3). Only the behaviour flags are different, and behaviour flags should be the same except the behaviours called Require and Override. 
https://releases.llvm.org/12.0.0/docs/LangRef.html#module-flags-metadata

PR #645, opened 2021-10-28
Handle record types in typeDescToWords

### LINKED ISSUE #643: Fix typeDescToWords to handle record types
They are currently getting serialized as `map<never>`.

ushirask:
Fixes #643

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/front.syntax/astString.bal
+++ b/compiler/modules/front.syntax/astString.bal
@@ -262,12 +262,12 @@ function typeDescToWords(Word[] w, TypeDesc td, boolean|BinaryTypeOp wrap = fals
                 typeDescToWords(w, rest);
                 w.push("...", CLING, ";");
             }
-            w.push(<Word>(firstInBlock ? LF : LF_OUTDENT), "|}");
+            w.push(<Word>LF_OUTDENT, "|}");
         }
         else {
             w.push("map", CLING, "<", CLING);
             if rest == () {
-                typeDescToWords(w, { startPos: td.startPos, endPos: td.endPos, builtinTypeName: "never" });
+                w.push("never");
             }
             else {
                 typeDescToWords(w, rest);
--- a/compiler/modules/front.syntax/astString.bal
+++ b/compiler/modules/front.syntax/astString.bal
@@ -247,6 +247,7 @@ function typeDescToWords(Word[] w, TypeDesc td, boolean|BinaryTypeOp wrap = fals
         return;
     }
     else if td is MappingTypeDesc {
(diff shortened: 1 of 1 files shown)


compiler/modules/front.syntax/astString.bal
```diff
@@ -247,6 +247,18 @@ function typeDescToWords(Word[] w, TypeDesc td, boolean|BinaryTypeOp wrap = fals
...
+            w.push("record", "{|");
+            boolean firstInBlock = true;
+            foreach var f in td.fields {
+                w.push(<Word>(firstInBlock ? LF_INDENT : LF));
+                firstInBlock = false;
+                typeDescToWords(w, f.typeDesc);
+                w.push(f.name, CLING, ";");
+            }
+            w.push(<Word>(firstInBlock ? LF : LF_OUTDENT), "|}");
+            return;
```
jclark: Please use `else` here.

Also handle the td.rest using `...`.

jclark (changes requested):

compiler/modules/front.syntax/astString.bal
```diff
@@ -247,16 +247,33 @@ function typeDescToWords(Word[] w, TypeDesc td, boolean|BinaryTypeOp wrap = fals
...
+        if td.fields.length() > 0 {
+            w.push("record", "{|");
+            boolean firstInBlock = true;
+            foreach var f in td.fields {
+                w.push(<Word>(firstInBlock ? LF_INDENT : LF));
+                firstInBlock = false;
+                typeDescToWords(w, f.typeDesc);
+                w.push(f.name, CLING, ";");
+            }
+            if rest != () {
```
jclark, line 260: I don't think the firstInBlock logic is working right with rest. Are we properly handling the case where there is a `rest` field but no `members`?

jclark (changes requested):

compiler/modules/front.syntax/astString.bal
ushirask: Can we differentiate between `type M map<int>` and
```
type R record {|
    int ...;
|}
```
in MappingTypeDesc ?

manuranga: No. Those are two ways of writing the same thing. We should pick `map<int>` as the canonical form.

ushirask: in the current implementation second form will also be serialized as `map<int>`

jclark (approved):

PR #647, opened 2021-10-28
Reintroduce misinterpreted tokenizer test cases

### LINKED ISSUE #646: Misinterpreted tokenizer test cases
#630 has misinterpreted [two tokenizer cases](https://github.com/ballerina-platform/nballerina/pull/630/files#r738614213).

1. Reintroduce those cases
2. Prevent this from happening in the future.

manuranga:
Fixes #646


jclark (approved):
What does the `O` flag mean?

manuranga:
I think this was added very early when strings still used to be an "Over-implementation".

PR #648, opened 2021-10-29
Parse type refs in local var decl statements

### LINKED ISSUE #640: Parse all type descriptors starting with an identifier in local var decl statements

manuranga:
Fixes #640


compiler/modules/front.syntax/token.bal
```diff
@@ -236,12 +236,14 @@ class Tokenizer {
         }
     }
 
+    // peeks the next token. considers qualified identifier as a single token, ie if current token
```
jclark: I suggest the special treatment of qualified identifier should be controllable by a parameter.

jclark: It also troubles me that the return value doesn't distinguish whether is is returning the next token of the token after a qualified identifier.

jclark (commented):
I agree this is a better approach.

compiler/modules/front.syntax/token.bal
manuranga: +1 for having a parameter such as `skipQualIdent` that control the behavior.

But couldn't find an nice way to return the additional tokens, I tried

```ballerina
type Peekable VariableTokenCode|FixedToken;
function peek() returns Peekable|[Peekable, Peekable, Peekable...] { ... }
```
Single Peekable is returned as is. if more than one is returned, it's a list. This is ugly because other side need to an additional check.


```ballerina
type Peekable VariableTokenCode|FixedToken;
function peek() returns [Peekable, Peekable...] { ... }
```
This look nicer, but there is unnecessary wrapping going on.

We are only using `peek` in two places, need for this is not concrete just yet. Shall we stick to the parameter for now?

jclark: OK, but add a comment to make it clear what the semantic of `peek` are.

compiler/modules/front.syntax/parseStmt.bal
```diff
@@ -21,6 +21,20 @@ function parseStmt(Tokenizer tok) returns Stmt|err:Syntax {
     Position startPos = tok.currentStartPos();
     match cur {
         [IDENTIFIER, var identifier] => {
+            var peeked = tok.peek(true);
```
jclark: With boolean arguments, it's clearer to use a named argument i.e `tok.peek(skipQualIdent=true);`

jclark: We don't always do this, but I think we should aim too.

compiler/modules/front.syntax/token.bal
```diff
@@ -236,12 +236,14 @@ class Tokenizer {
         }
     }
 
-    function peek() returns VariableTokenCode|FixedToken? {
+    // Peeks the next token. When skipQualIdent is passed, considers qualified identifier as a single token,
+    // ie if current token is the first identifier of a qualified identifier, returns the token after the second identifier.
+    function peek(boolean skipQualIdent) returns VariableTokenCode|FixedToken? {
```
jclark: I think this should default to `false`, since that is just a plain `peek` (and then use this default).

jclark (commented):
Looks great!

jclark (approved):

Issue #644, opened 2021-10-28
Fix breaking changes with jBallerina Swan Lake Beta4


manuranga:
Current breaking changes are here:
https://github.com/ballerina-platform/nballerina/actions/runs/1392592562

(Some of these warnings should go away with https://github.com/ballerina-platform/ballerina-lang/pull/33435 but this has not landed in dist yet.)

jclark:
I will fix the one in llvm.bal. Instead of panic we can use this idiom:

```
else {
            GeneralIntrinsicName _ = name;
```

where previously we would panic if it's not a GeneralIntrinsicName.

KavinduZoysa:
Getting `redeclared symbol` error is a JBUG bug, created an issue for that https://github.com/ballerina-platform/ballerina-lang/issues/33510

KavinduZoysa:
There is a fix for the warning `this function should explicitly return a value` in release branch, but it is not available in the beta4 rc1 release. It will be available from next release.

manuranga:
Some warninga will go away with ballerina-platform/ballerina-lang#33435, whatever remains we should fix under this issue.

manuranga:
@KavinduZoysa since the first RC is out, we need to close this issue as soon as possible. We need to report back issues before the voting mail closes.

manuranga:
Found https://github.com/ballerina-platform/ballerina-lang/issues/33694, I could inline `VariableLengthToken` as a workaround for now.

jclark:
Can we get this fixed for beta4?

hasithaa:
Yes, we are checking this.

jclark:
Does beta4 fix https://github.com/ballerina-platform/ballerina-lang/issues/33532? We need that in order to get rid of some of the warnings.

manuranga:
Yes it's fixed.

jclark:
We'll leave the rest of this for when beta4 is out.

PR #650, opened 2021-10-29
Resolve compile errors occured in beta4 rc1

KavinduZoysa:
## Purpose
Related to #644

### DIFF
--- a/compiler/balt.bal
+++ b/compiler/balt.bal
@@ -109,12 +109,10 @@ function parseBalt(string path) returns  BaltTestCase[]|io:Error|file:Error|err:
         else if s == CONTENT {
             content.push(l);
         }
-        else if s == BOF {
+        else {
+            BOF _ = s;
             // xxx add file path
             return error("file should start with 'Test-Case:' header field");
-        }
-        else {
-            panic err:impossible("balt parser illegal state");
         }
     }
     if s == HEADER {
--- a/compiler/modules/front.syntax/parseStmt.bal
+++ b/compiler/modules/front.syntax/parseStmt.bal
@@ -6,7 +6,7 @@ function parseStmtBlock(Tokenizer tok) returns StmtBlock|err:Syntax {
         Position startPos = tok.currentStartPos();
         Stmt[] stmts = [];
         check tok.advance();
-        cur = tok.current();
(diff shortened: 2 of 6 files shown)


compiler/modules/types/core.bal
```diff
@@ -405,6 +405,7 @@ class SubtypePairIteratorImpl {
                 }
             }
             else if self.i2 >= self.t2.length() {
+                // JBUG #33510 should be able to use variable `code`
```
KavinduZoysa, line 408: This fix will be available on rc2 release

compiler/balt.bal
```diff
@@ -109,13 +109,11 @@ function parseBalt(string path) returns  BaltTestCase[]|io:Error|file:Error|err:
         else if s == CONTENT {
             content.push(l);
         }
-        else if s == BOF {
+        else {
+            State _ = BOF;
```
jclark: That's not how it works. You need to do

BOF _ = s;

jclark (changes requested):

KavinduZoysa:
A fix for warning `this function should explicitly return a value` will be available in the next release. I will test it and update this PR after the new release

compiler/balt.bal
KavinduZoysa: Fixed

compiler/modules/front.syntax/parseStmt.bal
```diff
@@ -1,11 +1,11 @@
 import wso2/nballerina.err;
 
 function parseStmtBlock(Tokenizer tok) returns Stmt[]|err:Syntax {
-    Token? cur = tok.current();
-    if cur == "{" {
+    Token? firstToken = tok.current();
+    if firstToken == "{" {
```
manuranga: Can be inlined. But I don't understand what this is an error in beta4.

KavinduZoysa: Since we cannot assign value to the variable which is narrowed outside the loop. Please check https://github.com/ballerina-platform/ballerina-lang/pull/32293

KavinduZoysa: Changed

manuranga:
Commits moved to #680

PR #656, opened 2021-11-01
Move parser test data to an external file

### LINKED ISSUE #651: Move parser/tokenizer test data to an external file
Move test data in `testParser.bal` to an external file due to method too large issue. (jBallerina should fix it under ballerina-platform/ballerina-lang#33534).

manuranga:
Fixes #651


manuranga:
I decided not to move tokenizer tests for now because it has different structure, it'll make the code ugly, (or we should put that in a separate file).

compiler/modules/front.syntax/astString.bal
```diff
@@ -258,6 +258,12 @@ function typeDescToWords(Word[] w, TypeDesc td, boolean|BinaryTypeOp wrap = fals
         w.push(CLING, ">");
         return;
     }
+    else if td is ErrorTypeDesc {
+        w.push("map", CLING, "<", CLING);
```
jclark: It's usually a record. Why aren't we calling typeDescToWords on the detail?

jclark (changes requested):

compiler/modules/front.syntax/astString.bal
manuranga: I have made a mistake, this should be `error`.

compiler/modules/front.syntax/tests/data/testParser.json
```diff
@@ -0,0 +1,373 @@
+[
+    ["E", "expr", ""],
+    { "comment" : "not lexical errors" },
```
jclark, line 3: This is a bit ugly: some of these are akin to section names; some apply to a particular test. But I guess it's not worth spending time on at this point.

jclark (approved):

compiler/modules/front.syntax/tests/data/testParser.json
manuranga: I thought about adding another level, and decided against it thinking everything might not neatly fall into categories.
I'll create an issue to consider in the future.

PR #659, opened 2021-11-01
Change runtime backtrace tests

KavinduZoysa:
## Purpose
Related to #636

### DIFF
--- a/runtime/tests/error_backtrace2.c
+++ b/runtime/tests/error_backtrace2.c
@@ -7,43 +7,62 @@ HASH_DEFINE_KEY;
 
 // We have to check the function and filename seperately
 // because filepath is not a constant
-static const ExpectedTrace expectedTrace[] = {
-    { "    func2 ", "error_backtrace2.c:36" },
-    { "    func1 ", "error_backtrace2.c:28" },
-    { "    func2 ", "error_backtrace2.c:33" },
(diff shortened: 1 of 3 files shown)


runtime/tests/error_backtrace2.c
```diff
@@ -7,43 +7,62 @@ HASH_DEFINE_KEY;
...
-void func1(int, FILE *fp);
+static const ExpectedTrace expectedTrace2[] = {
+    { "    func2 ", "error_backtrace2.c:48" },
+    { "    func1 ", "error_backtrace2.c:40" },
+    { "    func2 ", "error_backtrace2.c:45" },
+    { "    func1 ", "error_backtrace2.c:40" },
+    { "    func2 ", "error_backtrace2.c:45" },
+    { "    func1 ", "error_backtrace2.c:40" },
+    { "    genAndCmpTrace ", "error_backtrace2.c:57" },
+    { "    main ", "error_backtrace2.c:66" }
```
manuranga, line 29: It looks like difference between having `-O2` and not is this line number. How about we just skip this line?

KavinduZoysa: I think it is good to check the whole backtrace and compare it with actual one, which is same as done in error_backtrace1.c

PR #666, opened 2021-11-04
Add position to BIR insns

### LINKED ISSUE #605: Add position to BIR insns
Every BIR insn should have a single position (not a range).  Long-term this will be used mainly for debugging info. Also used for error messages from verifier and backend.

Typically position should come from position of operator in AST (#604).

As we do this, we should fix verifier errors to use the position.

heshanpadmasiri:
## Purpose
Add position to BIR insns
Resolves #605 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/bir/bir.bal
+++ b/compiler/modules/bir/bir.bal
@@ -320,65 +320,65 @@ public type CompareInsn readonly & record {|
 
 # Constructs a new mutable list value.
 public type ListConstructInsn readonly & record {|
+    *InsnBase;
     INSN_LIST_CONSTRUCT_RW name = INSN_LIST_CONSTRUCT_RW;
     // The type of the result gives the inherent type of the constructed list
     Register result;
     Operand[] operands;
-    Position pos;
 |};
 
 # Gets a member of a list at a specified index.
 # This is a PPI (since the index may be out of bounds).
 public type ListGetInsn readonly & record {|
+    *InsnBase;
     INSN_LIST_GET name = INSN_LIST_GET;
     Register result;
     [Register, IntOperand] operands;
-    Position pos;
 |};
 
 # Sets a member of a list at a specified index.
 # This is a PPI (since the index may be out of bounds).
 public type ListSetInsn readonly & record {|
+    *InsnBase;
     INSN_LIST_SET name = INSN_LIST_SET;
     [Register, IntOperand, Operand] operands;
-    Position pos;
 |};
 
 # Constructs a new mutable list value.
 public type MappingConstructInsn readonly & record {|
+    *InsnBase;
     INSN_MAPPING_CONSTRUCT_RW name = INSN_MAPPING_CONSTRUCT_RW;
     // The type of the result gives the inherent type of the constructed list
     Register result;
     string[] fieldNames;
     Operand[] operands;
-    Position pos;
 |};
 
 # Gets a member of a mapping with a specified key.
 # This returns nil if there is no such member.
 # So this is not a PPI
 public type MappingGetInsn readonly & record {|
+    *InsnBase;
     INSN_MAPPING_GET name = INSN_MAPPING_GET;
     Register result;
     [Register, StringOperand] operands;
-    Position pos;
 |};
 
 # Sets a member of a mapping with a specified key.
 # This is a PPI.
 public type MappingSetInsn readonly & record {|
+    *InsnBase;
     INSN_MAPPING_SET name = INSN_MAPPING_SET;
     [Register, StringOperand, Operand] operands;
-    Position pos;
 |};
 
 # Constructs an error value.
 # Operand must be of type string.
 public type ErrorConstructInsn readonly & record {|
+    *InsnBase;
     INSN_ERROR_CONSTRUCT name = INSN_ERROR_CONSTRUCT;
     Register result;
     StringOperand operand;
-    Position pos;
 |};
 
 # This does equality expressions.
--- a/compiler/modules/bir/bir.bal
+++ b/compiler/modules/bir/bir.bal
@@ -188,6 +188,8 @@ public type InsnBase record {
     # Whether an instruction is a terminator or a PPI is determined
     # just by its name.
     InsnName name;
+    # Position based on the AST node that generated the instruction
+    Position pos;
 };
 
 public type Insn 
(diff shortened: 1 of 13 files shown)


compiler/modules/bir/bir.bal
```diff
@@ -188,6 +188,8 @@ public type InsnBase record {
     # Whether an instruction is a terminator or a PPI is determined
     # just by its name.
     InsnName name;
+    # Position based on the AST node that generated the instruction
+    Position pos;
```
jclark, line 192: This is good, but you should include it in all the insns, even the ones that provide their own defninition of `name`

compiler/modules/front/codeGen.bal
```diff
@@ -269,14 +269,14 @@ function codeGenFunction(ModuleSymbols mod, s:FunctionDefn defn, bir:FunctionSig
     }
     var { block: endBlock } = check codeGenStmts(cx, startBlock, { bindings }, defn.body);
     if endBlock != () {
-        bir:RetInsn ret = { operand: () };
+        bir:RetInsn ret = { operand: (), pos: defn.endPos };
         endBlock.insns.push(ret);
     }
-    codeGenOnPanic(cx);
+    codeGenOnPanic(cx, defn.startPos);
```
jclark: I think the endPos would be better here.

compiler/modules/front/codeGen.bal
```diff
@@ -389,13 +389,13 @@ function codeGenForeachStmt(CodeGenContext cx, bir:BasicBlock startBlock, Enviro
     initLoopVar.insns.push(init);
     bir:BasicBlock loopHead = cx.createBasicBlock();
     bir:BasicBlock exit = cx.createBasicBlock();
-    bir:BranchInsn branchToLoopHead = { dest: loopHead.label };
+    bir:BranchInsn branchToLoopHead = { dest: loopHead.label, pos: stmt.startPos };
```
jclark: Using the start position of the stmt feels a bit weird to me in these insns. Not sure what the right answer is.

jclark (changes requested):
I think we should have a StmtBlock AST node that wraps the `Stmt[]` in compound statements and can have its own start/endPos, distinct from the position of the statement that includes the block. Specifically, the startPos would be the position of the opening `{`. Then consider whether we should use that for any insns.

compiler/modules/bir/bir.bal
heshanpadmasiri: fixed

compiler/modules/front/codeGen.bal

compiler/modules/front/codeGen.bal
heshanpadmasiri: changed the `pos` of `BranchInsn` to the start of the loop to `stmt.endPos` (because we use it at the end of the loop) and kept the `pos` of `CondBranchInsn` at `stmt.startPos` (because we use it at the beginning of the loop)

heshanpadmasiri:
> I think we should have a StmtBlock AST node that wraps the `Stmt[]` in compound statements and can have its own start/endPos, distinct from the position of the statement that includes the block. Specifically, the startPos would be the position of the opening `{`. Then consider whether we should use that for any insns.

Added `StmtBlock`. Given we get all the current `Stmt[]` values from a single parser function (`function parseStmtBlock(Tokenizer tok) returns Stmt[]|err:Syntax`) and typically passed on to functions as the array it self (ex: `function blockToWords(Word[] w, Stmt[] body)` in `astString.bal` and `function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialEnv, s:Stmt[] stmts) returns CodeGenError|StmtEffect` in `codeGen.bal`) I think the easiest way is to make all AST nodes use `StmtBlock` and make all the functions that currently uses a `Stmt[]` deal with `StmtBlock` (this is what I have done now). Then if needed we can selectively make individual AST nodes not use it (in favor of `Stmt[]`)

jclark (changes requested):
The way you have added StmtBlock looks great!

Shouldn't we be using the startPos of the StmtBlock in some cases in foreach/while rather than the startPos if the foreach/while? Or perhaps the startPos of the expression following `while`? When you are iterating, you are repeating the StmtBlock multiple times (rather than the foreach/while), so it makes sense that insns with the position of the start of the StmtBlock should be executed multiple times.

heshanpadmasiri:
> The way you have added StmtBlock looks great!
> 
> Shouldn't we be using the startPos of the StmtBlock in some cases in foreach/while rather than the startPos if the foreach/while? Or perhaps the startPos of the expression following `while`? When you are iterating, you are repeating the StmtBlock multiple times (rather than the foreach/while), so it makes sense that insns with the position of the start of the StmtBlock should be executed multiple times.

Changed pos values to fallowings
```
"foreach" "int" identifier "in" additive-expr "..<" additive-expr "{" 
                            ^                  ^                   ^
                            |                  |                   |
                           (1)                (2)                 (3)

statement* "}"
            ^
            |
           (4)

1: position of assignment instruction (`bir:AssignInsn init = { pos:stmt.startPos, result: loopVar, operand: lower };`)
1: position of the increment (`bir:IntNoPanicArithmeticBinaryInsn increment = { op: "+", pos: stmt.startPos, operands: [loopVar, 1], result: loopVar };`)
1: position of the branch to loop step (`bir:BranchInsn branchToLoopStep = { dest: (<bir:BasicBlock>loopStep).label, pos: stmt.startPos };`)

2: position of compare instruction (`bir:CompareInsn compare = { op: "<", pos: stmt.range.opPos, orderType: t:UT_INT, operands: [loopVar, upper], result: condition };`)
2: position of the conditional branch (`bir:CondBranchInsn branch = { operand: condition, ifFalse: exit.label, ifTrue: loopBody.label, pos: stmt.range.opPos };`)

3: position of the branch to loop head (`bir:BranchInsn branchToLoopHead = { dest: loopHead.label, pos: stmt.endPos };`)

"while" expression "{" statement* "}"
        ^           ^              ^
        |           |              |
       (1)         (2)            (3)
              
1: position of conditional branch (`branch = <bir:CondBranchInsn>{ operand: condition, ifFalse: exit.label, ifTrue: loopBody.label, pos: stmt.condition.startPos };`)

2: position of branch to loop head (`bir:BranchInsn branchToLoopHead = { dest: loopHead.label, pos: stmt.body.endPos };`)
2: position of branch to to loop body (`branch = <bir:BranchInsn>{ dest: loopBody.label, pos: stmt.condition.startPos };`)

3: position of the branch to loop exit (`branch = <bir:BranchInsn> { dest: exit.label, pos: stmt.condition.startPos };)

```
Idea being conditional branch instructions will have the position of the condition and branch instructions will take a position corresponding to destination (break continue stmts add their own branch instructions that takes the start position of the statement)

jclark (approved):
That seems like a nice approach.

PR #671, opened 2021-11-08
Add decNumber

manuranga:
Add decNumber to third party branch


jclark:
On the main branch, I don't want to have a nested decNumber/decNumber directory.

manuranga:
This is just for the third-party branch, top level becomes the real name, inner names is the original name, in this case they happens to be the same.

PR #675, opened 2021-11-08
Support null as synonym for ()

### LINKED ISSUE #598: Support `null` as synonym for `()`
Ballerina allows `null` as a synonym for `()` both as a type descriptor and a literal.

https://ballerina.io/spec/lang/2021R1/#nil-type-descriptor
https://ballerina.io/spec/lang/2021R1/#nil-literal

heshanpadmasiri:
## Purpose
Support null as synonym for ()
Resolves #598

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


jclark (changes requested):
I suggest you call the test subdirectory `10-null`.

Also please update subset10.md to match.

jclark:
As you raised in the comment, `null` should only be allowed as a literal (including in `const`) not as a type-descriptor.

heshanpadmasiri:
> I suggest you call the test subdirectory `10-null`.
> 
> Also please update subset10.md to match.

fixed
`subset10.md` is already correct ([ref](https://github.com/ballerina-platform/nballerina/blob/f2f149e85a0508cc1750fcd903c5f0e7d36f563b/docs/subset10.md?plain=1#L297))

heshanpadmasiri:
> As you raised in the comment, `null` should only be allowed as a literal (including in `const`) not as a type-descriptor.

Currently when I get `null` as a type I am directly mapping that directly to the builtin type (since it can take only one value unlike say a string literal) so it will be same as if we receive (`()`). Is this correct or should I extend the `SingletonTypeDesc` to handle `null`?

jclark:
You raise an interesting point. To summarize:

* singleton-type-descriptor is a simple-const-expr, which allows nil-literal
* so `()` as a type-descriptor can be parsed in two ways: as a nil-type-descriptor or a singleton-type-descriptor
* but `null` can be parsed only as a singleton-type-descriptor

This is not something I had thought about while writing the spec.

jclark:
The idea of allowing `null` is to improve compatibility with JSON i.e. any JSON value should be valid Ballerina expression. Apart from that, we want people to use `()` rather than `null`. So I would expect that we would want to fix this issue by not allowing `null` in a simple-const-expr.

We don't currently support singletons in the subset grammar, so I think I would not at this point want to allow `null` in type descriptors.

heshanpadmasiri:
> The idea of allowing `null` is to improve compatibility with JSON i.e. any JSON value should be valid Ballerina expression. Apart from that, we want people to use `()` rather than `null`. So I would expect that we would want to fix this issue by not allowing `null` in a simple-const-expr.
> 
> We don't currently support singletons in the subset grammar, so I think I would not at this point want to allow `null` in type descriptors.

removed `null` from `simple-const-expr` and type descriptors

jclark:
We discussed this in the language design meeting this morning, and decided after discussion to clarify the spec to allow `null` as a type descriptor: https://github.com/ballerina-platform/ballerina-spec/issues/991

So can we can go back to the version that allows `null` as simple-const-expr.  Sorry to keep changing.  On the plus side, this has led to an improvement in the spec.

jclark (approved):

PR #676, opened 2021-11-08
Change runtime backtrace tests

KavinduZoysa:
## Purpose
Related to #636

### DIFF
--- a/runtime/tests/Makefile
+++ b/runtime/tests/Makefile
@@ -19,7 +19,12 @@ testCoverage: $(TESTPROG_PROF_RAW) $(TESTPROG_PROF)
 $(TESTPROG_PROF): $(TESTPROG_PROF_RAW)
 	llvm-profdata merge -sparse $(TESTPROG_PROF_RAW) -o $@
 
-%.stamp: %
+$(filter error_backtrace%, $(TESTPROG_STAMPS)): error_backtrace%.stamp: error_backtrace%
+	./$<
+	@./$<.sh || { exit 1; }
+	touch $@ 
+
+$(filter-out error_backtrace%, $(TESTPROG_STAMPS)): %.stamp: %
 	./$<
 	touch $@
 
--- a/runtime/tests/Makefile
+++ b/runtime/tests/Makefile
@@ -6,6 +6,7 @@ TESTPROG_STAMPS=$(addsuffix .stamp, $(TESTPROGS) $(TESTINLINERTPROGS))
 TESTPROG_PROF_RAW=$(addsuffix .profraw, $(TESTPROGS))
 TESTPROG_PROF = test.profdata
 TESTPROG_EXE=$(addprefix -object ./, $(TESTPROGS))
+SHELL=/bin/bash
 
 all: $(TESTPROGS)
 
@@ -19,12 +20,12 @@ testCoverage: $(TESTPROG_PROF_RAW) $(TESTPROG_PROF)
 $(TESTPROG_PROF): $(TESTPROG_PROF_RAW)
 	llvm-profdata merge -sparse $(TESTPROG_PROF_RAW) -o $@
 
-$(filter error_backtrace%, $(TESTPROG_STAMPS)): error_backtrace%.stamp: error_backtrace%
-	./$<
-	@./$<.sh || { exit 1; }
+error_backtrace%.stamp: error_backtrace%
+	set -o pipefail; \
+	./$< | ./$<.sh
 	touch $@ 
 
-$(filter-out error_backtrace%, $(TESTPROG_STAMPS)): %.stamp: %
+%.stamp: %
 	./$<
 	touch $@
 
--- a/runtime/tests/Makefile
+++ b/runtime/tests/Makefile
@@ -22,7 +22,7 @@ $(TESTPROG_PROF): $(TESTPROG_PROF_RAW)
 
 error_backtrace%.stamp: error_backtrace%
 	set -o pipefail; \
-	./$< | ./$<.sh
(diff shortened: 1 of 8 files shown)


runtime/tests/Makefile
```diff
@@ -19,7 +19,12 @@ testCoverage: $(TESTPROG_PROF_RAW) $(TESTPROG_PROF)
 $(TESTPROG_PROF): $(TESTPROG_PROF_RAW)
 	llvm-profdata merge -sparse $(TESTPROG_PROF_RAW) -o $@
 
-%.stamp: %
+$(filter error_backtrace%, $(TESTPROG_STAMPS)): error_backtrace%.stamp: error_backtrace%
```
jclark: Just do two rules

```
%.stamp: %
```

and

```
error_backtrace%.stamp
```

GNU make will choose the longest stem

https://www.gnu.org/software/make/manual/html_node/Pattern-Match.html

runtime/tests/error_backtrace1.c
```diff
@@ -3,65 +3,29 @@
...
-        assert(genAndCmpTrace(ep, expectedTrace1, 2));
-    }
-    func2();
+    _bal_error_backtrace_print(ep, 1, fp);
+    func2(fp);
 }
 
 int main() {
-    func1();
+    FILE *fp = fopen("backtrace.txt", "w+");
```
jclark: Don't write to `backtrace.txt`. This won't work in parallel. Write to stdout.

runtime/tests/Makefile
```diff
@@ -19,7 +19,12 @@ testCoverage: $(TESTPROG_PROF_RAW) $(TESTPROG_PROF)
...
+	./$<
```
jclark: Pipe the output of the error_backtrace program into the shell script.  Set pipefail (as is done in the test Makefile).

runtime/tests/error_backtrace1.sh
```diff
@@ -0,0 +1,21 @@
...
+"    func2 .*\/error_backtrace1.c:13" \
+"    func1 .*\/error_backtrace1.c:23" \
+"    main .*\/error_backtrace1.c:29" \
+"    func3 .*\/error_backtrace1.c:7" \
+"    func2 .*\/error_backtrace1.c:16" \
+"    func1 .*\/error_backtrace1.c:23" \
+"    main .*\/error_backtrace1.c:29" \
+)
+
+for t in "${trace[@]}"; do
```
jclark: Can you explain what this is doing?  Why can't you massage the output with sed and then use diff? That would give a much more informative error message.

jclark (changes requested):

runtime/tests/error_backtrace1.sh
KavinduZoysa: @jclark, What I am doing here is, check whether the expected line is available in the actual backtrace. We have to check line by line because without optimizations actual backtrace will have more trace lines.

jclark: I would use `join` for that.

jclark: https://man7.org/linux/man-pages/man1/join.1.html

jclark: I would avoid using bash extensions as much as possible, and stick to portable sh constructs, which everybody will be familiar with.

runtime/tests/Makefile
KavinduZoysa: Fixed

runtime/tests/error_backtrace1.c
KavinduZoysa: Fixed

runtime/tests/Makefile
KavinduZoysa: Fixed

runtime/tests/error_backtrace1.sh
KavinduZoysa: @jclark, I am not sure whether I understand this comment correctly. Let's say, in actual and expected traces, function names and filenames are identical but line numbers are different. In this case, join commands run successfully. But In our case, we should get an error.
Can you please explain more about this?

jclark: You can join on whatever you need, including line number. But perhaps the comm command work better for this case.

jclark: Let me make sure I have understood the problem properly. We have two files F1 and F2 and we need to test whether F2 is the same as F1 except possibly with some lines added. Is that right?

If so, join/comm aren't quite right because they won't ensure lines are in the right order.  So I think the easiest way is to diff F1 F2 (you can diff stdin by using an argument of `-`), and then grep the result for changes that are deletions. If you find any, then the test fails.

KavinduZoysa: > Let me make sure I have understood the problem properly. We have two files F1 and F2 and we need to test whether F2 is the same as F1 except possibly with some lines added. Is that right?
> 
@jclark, Yes, But in the backtrace, we have the absolute file path with file name, that is why I try to match with a pattern to ignore the path(ex: `main .*\/error_backtrace1.c:29`) 
In diff command we always get a difference because paths are different.

jclark: So then you pipe the output from the error_backtrace through sed to get rid of the absolute paths before piping that into diff.

KavinduZoysa: Changed.

jclark (changes requested):
You only need one .sh file. Pass in the .txt filename as an argument.

KavinduZoysa:
> You only need one .sh file. Pass in the .txt filename as an argument.

Changed

Issue #289, opened 2021-08-09
Implement `decimal` basic type


jclark:
We can use the decNumber package from 

https://github.com/hercules-390/decNumber-icu-368

or

https://github.com/SDL-Hercules-390/decNumber

This has a more permissive license than the copy that comes with GCC.

Sub issues:

- [x] #755
- [x] #766
- [x] #759
- [x] #779
- [x] #777
- [x] #778
- [x] #782
- [x] #780
- [x] #781
- [x] #783

jclark:
Note that Ballerina uses the simplified X3.274 subset described here

http://speleotrove.com/decimal/dax3274.html

jclark:
We should have an immediate format for decimals. decNumber uses DPD format for encoding decimals, described here:

http://speleotrove.com/decimal/DPDecimal.html

This packs 3 decimal digits into 10 bits.

We have 56 bits available. So one possibility is

- 1 sign bit
- 5 bits representing an integer in the range 0-31 saying where the decimal point goes
- 50 bits encoding 15 digits (10 bits for each of 5 groups of 3 digits)

Another possibility would be BCD:

- 1 sign bit
- 7 bits representing an integer in the range 0-128 saying where the decimal point goes
- 48 bits encoding 12 digits (4 bits per digit)

or

- 1 sign bit
- 3 bits representing an integer in the range 0-7 saying how many digits after the decimal point
- 52 bits encoding 13 digits (4 bits per digit)

jclark:
@KavinduZoysa I suggest following steps

- [ ] Familiarize yourself with http://speleotrove.com/decimal/
- [ ] Look at the two versions of decNumber I linked to, analyze the differences and figure out which one we should use
- [ ] Bring chosen one into third-party
- [ ] Figure out how to integrate into our build (as you did with libbacktrace)
- [ ] Determine whether library has built-in support for X3.274 subset, or do we need to do it ourselves

jclark:
We could also consider this library: https://www.bytereef.org/mpdecimal/

KavinduZoysa:
I was able to build the two versions of decNumber and run a simple example. 
If we consider about the diff between these two version there is not a difference in the core implementation (in `include` `folder` and source folder). In the [forked version](https://github.com/SDL-Hercules-390/decNumber) authors has done some build related changes such as improve and refactor build sripts, add visual studio support for windows etc and fixed some formating issues. 

My next step is,

[...]

PR #688, opened 2021-11-10
Generate decimal constants

KavinduZoysa:
## Purpose
Related to #289


runtime/decimal.c
```diff
@@ -0,0 +1,17 @@
...
+
+TaggedPtr _bal_decimal_const(const char *decString) {
+    GC decimal128 *dp = _bal_alloc(sizeof(decimal128));
+
+    decContext ctx;
+    decContextDefault(&ctx, DEC_INIT_DECQUAD);
+    decQuadFromString(balDecToDecQuad(dp), decString, &ctx);
+    return ptrAddFlags(dp, (uint64_t)TAG_DECIMAL << TAG_SHIFT);
+}
+
```
manuranga, line 22: Move this to line 4 so you don't need a declaration.

runtime/tests/decimal.c
```diff
@@ -0,0 +1,26 @@
...
+#include "../balrt.h"
+#include "../hash.h"
+#include "../third-party/decNumber/decQuad.h"
+
+HASH_DEFINE_KEY;
+
+void genAndValidateConstDec(const char *decStr, const char *dec) {
+    TaggedPtr d = _bal_decimal_const(decStr);
+    char str[DECQUAD_String];
+    decQuadToString((decQuad *)taggedToPtr(d), str);
```
manuranga: Lets asset the tag as well

runtime/decimal.c
```diff
@@ -0,0 +1,17 @@
+#include "balrt.h"
+#include "third-party/decNumber/decQuad.h"
+
+static inline decQuad *balDecToDecQuad(GC decimal128 *dp);
+
+TaggedPtr _bal_decimal_const(const char *decString) {
+    GC decimal128 *dp = _bal_alloc(sizeof(decimal128));
+
+    decContext ctx;
+    decContextDefault(&ctx, DEC_INIT_DECQUAD);
```
jclark: decQuad/decFloat is documented to only use status and round fields. Eventually we will want to tweak decQuad to use a smaller context object. In the meantime, we should create our own inline function to initialize just these two members.

runtime/balrt.h
```diff
@@ -69,6 +69,10 @@ typedef uint64_t PackedPanic;
 
 typedef uintptr_t PC;
 
+typedef struct {
```
jclark: Given we are just having a tagged representation, we don't need this any more.

runtime/decimal.c
KavinduZoysa: This method was removed

runtime/decimal.c
KavinduZoysa: FIxed

runtime/balrt.h
KavinduZoysa: Removed.

runtime/tests/decimal.c
```diff
@@ -0,0 +1,49 @@
+#include <string.h>
+#include <assert.h>
+#include "../balrt.h"
+#include "../hash.h"
+#include "../third-party/decNumber/decQuad.h"
+
+HASH_DEFINE_KEY;
+
+void validate(TaggedPtr tp, const char *decStr) {
```
jclark: Call this file something other than decimal.c to avoid confusion with file in parent directory. Also add filename to .gitignore.

runtime/decimal.c
```diff
@@ -23,14 +18,19 @@ static inline PanicCode statusPanicCode(uint32_t status) {
...
-    return ptrAddFlags(dp, (uint64_t)TAG_DECIMAL);
+    return ptrAddFlags(dp, (uint64_t)TAG_DECIMAL<< TAG_SHIFT);
 }
 
 static inline decQuad *taggedToDecQuad(TaggedPtr tp) {
     UntypedPtr p = taggedToPtr(tp);
     return (void *)p;
 }
 
+static inline void updateContext(decContext *cx) {
```
jclark: initContext

runtime/decimal.c
```diff
@@ -23,14 +18,19 @@ static inline PanicCode statusPanicCode(uint32_t status) {
 static inline TaggedPtr createDecimal(decQuad *dq) {
     DecimalPtr dp = _bal_alloc(sizeof(decQuad));
     *dp = *dq;
-    return ptrAddFlags(dp, (uint64_t)TAG_DECIMAL);
+    return ptrAddFlags(dp, (uint64_t)TAG_DECIMAL<< TAG_SHIFT);
```
jclark: Space before `<<`

runtime/decimal.c

runtime/tests/decimal.c
KavinduZoysa: Fixed

runtime/decimal.c
```diff
@@ -23,14 +18,19 @@ static inline PanicCode statusPanicCode(uint32_t status) {
 static inline TaggedPtr createDecimal(decQuad *dq) {
     DecimalPtr dp = _bal_alloc(sizeof(decQuad));
     *dp = *dq;
-    return ptrAddFlags(dp, (uint64_t)TAG_DECIMAL);
+    return ptrAddFlags(dp, (uint64_t)TAG_DECIMAL << TAG_SHIFT);
 }
 
 static inline decQuad *taggedToDecQuad(TaggedPtr tp) {
```
jclark: Let's make this return `const decQuad *` to enforce the immutability of decimal values.

runtime/decimal.c
```diff
@@ -23,14 +18,19 @@ static inline PanicCode statusPanicCode(uint32_t status) {
...
     UntypedPtr p = taggedToPtr(tp);
     return (void *)p;
 }
 
+static inline void initContext(decContext *cx) {
```
jclark, line 23: Use this in _bal_decimal_add also.

runtime/decimal.c
```diff
@@ -47,5 +47,10 @@ TaggedPtrPanicCode _bal_decimal_add(TaggedPtr tp1, TaggedPtr tp2) {
     return result;
```
jclark: This will occur in all the operations. So let's factor out a function:

```
static inline TaggedPtrPanicCode finish(decQuad *dq, decContext *cx) {
    TaggedPtrPanicCode result;
    result.panicCode = statusPanicCode(cx->status);
    if (result.panicCode == 0) {
        result.ptr = createDecimal(dq);
    }
    return result;  
}
```

We we will also need to deal with negative zero (which I think can only happen with underflow). So let's inline staticPanicCode into finish above.

KavinduZoysa: @jclark, we need to add this method to the `_bal_decimal_const` right? because in that method also we have to deal with errors like underflow/overflow.

runtime/decimal.c
KavinduZoysa: Done

runtime/decimal.c
KavinduZoysa: Done

runtime/decimal.c
jclark: No: these are constants generated by the compiler. The compiler needs to generate only valid constants.

runtime/decimal.c
```diff
@@ -23,29 +18,42 @@ static inline PanicCode statusPanicCode(uint32_t status) {
 static inline TaggedPtr createDecimal(decQuad *dq) {
     DecimalPtr dp = _bal_alloc(sizeof(decQuad));
     *dp = *dq;
-    return ptrAddFlags(dp, (uint64_t)TAG_DECIMAL);
+    return ptrAddFlags(dp, (uint64_t)TAG_DECIMAL << TAG_SHIFT);
+}
+
+static inline TaggedPtrPanicCode finish(decQuad *dq, decContext *cx) {
+    TaggedPtrPanicCode result;
+    result.panicCode = statusPanicCode(cx->status);
```
jclark: Inline statusPanicCode.

jclark (changes requested):

runtime/tests/decimal_operations.c
```diff
@@ -0,0 +1,49 @@
...
+    genAndValidateConstDec("12345678901234567890123456789012345", "1.234567890123456789012345678901234E+34");
+    genAndValidateConstDec("123456789123456789123456789123456789", "1.234567891234567891234567891234568E+35");
+    genAndValidateConstDec("99999999999999999999999999999999999", "1.000000000000000000000000000000000E+35");
+    genAndValidateConstDec("123456789123456789123456789123456750", "1.234567891234567891234567891234568E+35");
+}
+
+void genAndValidateDecAdd(const char *decStr1, const char *decStr2, const char *sum) {
+    TaggedPtr tp1 = _bal_decimal_const(decStr1);
+    TaggedPtr tp2 = _bal_decimal_const(decStr2);
+    TaggedPtrPanicCode tp = _bal_decimal_add(tp1, tp2);
```
jclark: Need to assert that the panicCode is zero.

runtime/decimal.c

runtime/decimal.c

runtime/tests/decimal_operations.c
KavinduZoysa: Changed

runtime/decimal.c
```diff
@@ -3,49 +3,59 @@
...
+    TaggedPtrPanicCode result;
+    uint32_t status = cx->status;
+    if (status & DECIMAL_STATUS_FAIL) {
+        // XXX decide what panic codes to use 
+        result.panicCode = (status & DEC_Division_by_zero) ? PANIC_DIVIDE_BY_ZERO : PANIC_ARITHMETIC_OVERFLOW;
+    }
+    else {
+        result.panicCode = 0;
+    }
+    if (result.panicCode == 0) {
```
manuranga: this if needs to merge with above if-else

runtime/decimal.c
```diff
@@ -3,49 +3,57 @@
...
-
 static inline TaggedPtr createDecimal(decQuad *dq) {
     DecimalPtr dp = _bal_alloc(sizeof(decQuad));
     *dp = *dq;
-    return ptrAddFlags(dp, (uint64_t)TAG_DECIMAL);
+    return ptrAddFlags(dp, (uint64_t)TAG_DECIMAL << TAG_SHIFT);
 }
 
-static inline decQuad *taggedToDecQuad(TaggedPtr tp) {
+static inline TaggedPtrPanicCode finish(decQuad *dq, decContext *cx) {
```
jclark: Since we're now calling decQuadClass, let's not inline this.

runtime/tests/decimal_operations.c
```diff
@@ -0,0 +1,59 @@
...
+    if (tp.panicCode > 0) {
+        assert(tp.panicCode == code);
+    }
+    else {
+        validate(tp.ptr, sum);
+    }
+}
+
+// TODO: Should validate errorCode
```
jclark: You are doing this now, aren't you?

runtime/tests/decimal_operations.c
```diff
@@ -0,0 +1,59 @@
...
+void testAdd() {
+    genAndValidateDecAdd("1", "1", "2", 0);
+    genAndValidateDecAdd("1000", "1", "1001", 0);
+    genAndValidateDecAdd("1234567890123456789012345678901234", "1234567890123456789012345678901231", "2469135780246913578024691357802465", 0);
+    genAndValidateDecAdd("12345678901234567890123456789012341", "12345678901234567890123456789012312", "2.469135780246913578024691357802465E+34", 0);
+    genAndValidateDecAdd("1234567890123456789012345678901234", "12345678901234567890123456789012312", "1.358024679135802467913580246791354E+34", 0);
+    genAndValidateDecAdd("9e6144", "1e6144", "", PANIC_ARITHMETIC_OVERFLOW);
```
jclark: Can we do cases that are close to the maximum, minimum representable value, and close to the minimum magnitude?

runtime/tests/decimal_operations.c
```diff
@@ -0,0 +1,59 @@
+#include <string.h>
+#include <assert.h>
+#include "../balrt.h"
+#include "../hash.h"
+#include "../third-party/decNumber/decQuad.h"
+
+HASH_DEFINE_KEY;
+
+void validate(TaggedPtr tp, const char *decStr) {
```
jclark: I suggest you call this decimal_arith.c and use it for all arithmetic operations.

runtime/decimal.c
```diff
@@ -3,49 +3,57 @@
...
     UntypedPtr p = taggedToPtr(tp);
     return (void *)p;
 }
 
+static inline void initContext(decContext *cx) {
+    cx->status = 0;
+    cx->round = DEC_ROUND_HALF_EVEN;
+}
+
 TaggedPtrPanicCode _bal_decimal_add(TaggedPtr tp1, TaggedPtr tp2) {
```
jclark, line 28: Let's do `_bal_decimal_sub` as part of this pull request, since it does not raise any different issues.

jclark (changes requested):

runtime/decimal.c
KavinduZoysa: Added

runtime/tests/decimal_operations.c
KavinduZoysa: Changed

runtime/tests/decimal_operations.c
KavinduZoysa: yes, added

runtime/tests/decimal_operations.c
KavinduZoysa: Yes, removed

runtime/decimal.c

runtime/decimal.c
```diff
@@ -3,49 +3,69 @@
...
+    return finish(&d, &cx);    
 }
 
+TaggedPtrPanicCode _bal_decimal_sub(TaggedPtr tp1, TaggedPtr tp2) {
+    const decQuad *dq1 = taggedToDecQuad(tp1);
+    const decQuad *dq2 = taggedToDecQuad(tp2);
+    decQuad d;
+    decContext cx;
+    initContext(&cx);
+    decQuadSubtract(&d, dq1, dq2, &cx);
```
jclark, line 44: The dq1 and dq2 variables aren't buying us anything. So let's inline them.

```
decQuadSubtract(&d, taggedToDecQuad(tp1), taggedToDecQuad(tp2), &cx);
```

You can do this in the next pull request.

jclark (approved):

PR #692, opened 2021-11-11
Store the position with "named" registers 

### LINKED ISSUE #691: Store the position with "named" registers
We need to store the position of registers with a `varName` so errors caused by those registers can have meaningful positions.

heshanpadmasiri:
## Purpose
Store the position with "named" registers 
Resolves #691

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/bir/bir.bal
+++ b/compiler/modules/bir/bir.bal
@@ -128,21 +128,24 @@ public function lastInsnRef(BasicBlock bb) returns InsnRef {
     return { block: bb.label, index: bb.insns.length() - 1 };
 }
 
-public type PositionString readonly & [Position, string];
-
 public type Register readonly & record {|
     # Unique identifier within a function
     # Always >= 0
     int number;
     SemType semType;
-    PositionString? varName;
+    string? varName;
+    Position? pos;
 |};
 
-public function createRegister(FunctionCode code, SemType semType, PositionString? varName = ()) returns Register {
+public function createRegister(FunctionCode code, SemType semType, string? varName = (), Position? pos = ()) returns Register {
     int number = code.registers.length();
-    Register r = { number, semType, varName };
+    Register r = { number, semType, varName, pos };
     code.registers.push(r);
     return r;
+}
+
+public function getRegister(FunctionCode code, int registerNo) returns Register {
+    return code.registers[registerNo];
 }
 
 public type ArithmeticBinaryOp "+" | "-" | "*" | "/" | "%";
--- a/compiler/modules/bir/bir.bal
+++ b/compiler/modules/bir/bir.bal
@@ -128,17 +128,38 @@ public function lastInsnRef(BasicBlock bb) returns InsnRef {
     return { block: bb.label, index: bb.insns.length() - 1 };
 }
 
-public type Register readonly & record {|
+public distinct readonly class Register {
     # Unique identifier within a function
     # Always >= 0
-    int number;
-    SemType semType;
-    string? varName;
-|};
+    public final int number;
+    public final SemType semType;
+    function init(int number, SemType semType) {
+        self.number = number;
+        self.semType = semType;
+    }
+}
+
+public distinct readonly class NamedRegister {
+    *Register;
+    public final int number;
+    public final SemType semType;
+    public final string varName;
+    function init(int number, SemType semType, string varName) {
+        self.number = number;
+        self.semType = semType;
+        self.varName = varName;
+    }
+}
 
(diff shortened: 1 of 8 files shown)


compiler/modules/bir/bir.bal
```diff
@@ -127,17 +127,41 @@ public function lastInsnRef(BasicBlock bb) returns InsnRef {
     return { block: bb.label, index: bb.insns.length() - 1 };
 }
 
-public type Register readonly & record {|
+public distinct readonly class Register {
```
jclark: No. Leave it as a record.

Just change varName to have type `PositionString?`

```
public type PositionString readonly & [Position, string];
```

Subtyping is only relevant to codeGen. So in that module create a structural subtype

```
type PositionRegister record {|
  *bir:Register;
  PositionString varName;
|};
```

heshanpadmasiri: I considered doing this but if I did 
```
public type PositionString readonly & [Position, string];

public type Register readonly & record {|
    int number;
    SemType semType;
    PositionString? varName;
|};

public type PositionRegister readonly & record {|
    *Register;
    PositionString varName;
|};
```
I am getting an error at "*Register" saying "incompatible types: 'Register' is not a record. We can't also do the workaround we have used in `ast.bal` by replacing "*Register" with 
```
int number;
SemType semType;
```
because this don't allow me cast from `Register` to `PositionRegister`

jclark: It's probably the `readonly` that is causing the problem.

In any case, it's a complete non-starter to use a class here. So get rid of that first. If you can't make the subtype work, it's no big deal.

jclark (changes requested):

heshanpadmasiri:
fixed (removed register subtyping)

compiler/modules/front/codeGen.bal
```diff
@@ -103,9 +103,11 @@ class CodeGenContext {
         self.returnType = returnType;
     }
 
-    function createRegister(bir:SemType t, string? varName = ()) returns bir:Register {
+    function createRegister(bir:SemType t, bir:PositionString? varName = ()) returns bir:Register {
         bir:Register reg = bir:createRegister(self.code, t, varName);
-        self.registerVarNames[reg.number] = varName;
+        if varName != () {
+            self.registerVarNames[reg.number] = varName[1];
```
jclark: But you're accessing this array in validLoopAssignments, which needs the position.

Do we need the PositionString in the registerVarNames array or in the bir:Register or both?

heshanpadmasiri: added a new function `registerPosition` to get the register position (leaving `registerVarName` to return the register name). If we have a reference to the `bir:Register` having  PositionString there is enought. But in cases (such as `validLoopAssignments` that only has the register number we need something else to get the position).

jclark:
If we have registerVarNames containing a PositionString, then why do we need the Position in the Register?

registerVarNames needs to be renamed if it is a PositionString. Probably better to use a record than a tuple.

compiler/modules/bir/bir.bal
```diff
@@ -134,15 +134,20 @@ public type Register readonly & record {|
...
-public function createRegister(FunctionCode code, SemType semType, string? varName = ()) returns Register {
+public function createRegister(FunctionCode code, SemType semType, string? varName = (), Position? pos = ()) returns Register {
     int number = code.registers.length();
-    Register r = { number, semType, varName };
+    Register r = { number, semType, varName, pos };
     code.registers.push(r);
     return r;
 }
 
+public function getRegister(FunctionCode code, int registerNo) returns Register {
```
jclark, line 147: I don't usually abbreviate `number` to `no`: too much like `yes`. Use either `num` or `number`.

compiler/modules/front.syntax/ast.bal
```diff
@@ -36,6 +36,7 @@ public type FunctionDefn record {|
     Visibility vis;
     FunctionTypeDesc typeDesc;
     string[] paramNames;
+    Position[] paramPos;
```
jclark, line 39: We have too many inconsistent ways of doing the position for a name:

- name + namePos
- paramNames + paramPos arrays
- varName + varPos
- `[string, Position]` identity


"identity" is worse than "name" or "identifier".

jclark (changes requested):

heshanpadmasiri:
closed in favor of #700

PR #696, opened 2021-11-12
Specialize representations for int[] and float[]

### LINKED ISSUE #329: Implement specialized representations for `int[]` and `float[]`
This improves the initial implementation of arrays from #328 to provide specialized implementations for `int[]` and `float[]`.

This should use the idea in #204.

After this we can tackle #169

manuranga:
Fixes #329

### DIFF
--- a/compiler/modules/nback/init.bal
+++ b/compiler/modules/nback/init.bal
@@ -50,6 +50,7 @@ const TYPE_KIND_PRECOMPUTED = "precomputed";
 type TypeKindArrayOrMap TYPE_KIND_ARRAY|TYPE_KIND_MAP;
 type TypeKind TypeKindArrayOrMap|TYPE_KIND_RECORD|TYPE_KIND_PRECOMPUTED;
 
+type ListAccessors (llvm:FunctionDecl|llvm:ConstPointerValue)[];
 type InitModuleContext record {|
     llvm:Context llContext;
     llvm:Module llMod;
@@ -61,8 +62,7 @@ type InitModuleContext record {|
     table<SubtypeDefn> key(typeCode, semType) subtypeDefns = table [];
     InitTypes llTypes;
     map<llvm:FunctionDecl> typeTestFuncs = {};
-    llvm:FunctionDecl? listSetFunc = ();
-    llvm:FunctionDecl? listGetFunc = ();
+    ListAccessors? genericListAccessors = ();
     // subtype definitions cannot be completed before inherent types are complete,
     // because precomputed subtypes need to know all inherent types
     boolean inherentTypesComplete;
@@ -162,7 +162,13 @@ function addInherentTypeDefn(InitModuleContext cx, string symbol, t:SemType semT
 }
 
 function addArrayInherentTypeDefn(InitModuleContext cx, string symbol, int tid, t:SemType memberType) returns [llvm:StructType, llvm:ConstPointerValue] {
-    llvm:ConstValue initValue = cx.llContext.constStruct([llvm:constInt(LLVM_TID, tid), getListGetFunc(cx), getListSetFunc(cx), getMemberType(cx, memberType)]);
+    ListAccessors accessors = getListAccessors(cx, typeToListRepr(memberType));
+    llvm:Value[] initStructValues = [llvm:constInt(LLVM_TID, tid)];
+    foreach var funcDecl in accessors {
+        initStructValues.push(funcDecl);
+    }
+    initStructValues.push(getMemberType(cx, memberType));
+    llvm:ConstValue initValue = cx.llContext.constStruct(initStructValues);
     llvm:ConstPointerValue ptr = cx.llMod.addGlobal(llListDescType, symbol, initializer=initValue, isConstant=true);
     return [llListDescType, ptr];
 }
@@ -358,28 +364,35 @@ function createRecordSubtypeStruct(InitModuleContext cx, string[] fieldNames, t:
     };
 }
 
-function getListSetFunc(InitModuleContext cx) returns llvm:FunctionDecl {
-    llvm:FunctionDecl? existing = cx.listSetFunc;
-    if existing == () {
-        llvm:FunctionDecl decl = cx.llMod.addFunctionDecl(mangleRuntimeSymbol("list_set"), llListSetFuncType);
-        cx.listSetFunc = decl;
-        return decl;
+function getListAccessors(InitModuleContext cx, ListRepr repr) returns ListAccessors {
+    if repr.isSpecialized {
+        return createListAccessor(cx, repr.rtFuncSuffix);
     }
     else {
-        return existing;
+        ListAccessors? existing = cx.genericListAccessors;
+        if existing == () {
+            ListAccessors a = createListAccessor(cx, repr.rtFuncSuffix);
+            cx.genericListAccessors = a;
+            return a;
+        }
+        else {
+            return existing;
+        }
     }
 }
 
-function getListGetFunc(InitModuleContext cx) returns llvm:FunctionDecl {
-    llvm:FunctionDecl? existing = cx.listGetFunc;
-    if existing == () {
-        llvm:FunctionDecl decl = cx.llMod.addFunctionDecl(mangleRuntimeSymbol("list_get"), llListGetFuncType);
-        cx.listGetFunc = decl;
-        return decl;
+function createListAccessor(InitModuleContext cx, string?[] rtFuncSuffix) returns ListAccessors {
+    ListAccessors result = [];
+    foreach int i in 0 ..< rtFuncSuffix.length() {
+        string? funcSuffix = rtFuncSuffix[i];
+        if funcSuffix != () {
+            result.push(cx.llMod.addFunctionDecl(mangleRuntimeSymbol("list_" + funcSuffix), llListDescFuncs[i]));
+        }
+        else {
+            result.push(llvm:constNull(llListDescFuncPtrs[i]));
+        }
     }
-    else {
-        return existing;
-    }
+    return result;
 }
 
 function getSubtypeContainsFunc(InitModuleContext cx, TypeKind tk) returns llvm:FunctionDecl {
--- a/compiler/modules/nback/structure.bal
+++ b/compiler/modules/nback/structure.bal
@@ -67,6 +67,68 @@ final RuntimeFunction mappingConstructFunction = {
 
 const LLVM_INDEX = "i32";
 
+type ListRepr readonly & object {
+    llvm:Type memberType;
+    string?[] rtFuncSuffix;
+    int listDescGetIndex;
+    int listDescSetIndex;
+    boolean isSpecialized;
+    function buildMember(llvm:Builder builder, Scaffold scaffold, bir:Operand member, t:SemType memberType) returns llvm:Value|BuildError;
+    function buildMemberStore(llvm:Builder builder, Scaffold scaffold, llvm:Value value, bir:Register reg);
+};
+
+final ListRepr listGenericRepr = object {
+    llvm:Type memberType = LLVM_TAGGED_PTR;
+    string?[] rtFuncSuffix = ["generic_get_tagged", "generic_set_tagged", "generic_get_int", "generic_set_int", "generic_get_float", "generic_set_float"];
+    int listDescGetIndex = 1;
+    int listDescSetIndex = 2;
+    boolean isSpecialized = false;
+    function buildMember(llvm:Builder builder, Scaffold scaffold, bir:Operand member, t:SemType memberType) returns llvm:Value|BuildError {
+        return buildWideRepr(builder, scaffold, member, REPR_ANY, memberType);
+    }
+    function buildMemberStore(llvm:Builder builder, Scaffold scaffold, llvm:Value value, bir:Register reg) {
+        return buildStoreTagged(builder, scaffold, value, reg);
+    }
+};
+
+final ListRepr listIntArrayRepr = object {
+    llvm:Type memberType = LLVM_INT;
+    string?[] rtFuncSuffix = ["int_array_get_tagged", "int_array_set_tagged", "int_array_get_int", "int_array_set_int", (), "int_array_set_float"];
+    int listDescGetIndex = 3;
+    int listDescSetIndex = 4;
+    boolean isSpecialized = true;
+    function buildMember(llvm:Builder builder, Scaffold scaffold, bir:Operand member, t:SemType memberType) returns llvm:Value|BuildError {
+        return buildInt(builder, scaffold, <bir:IntOperand>member);
+    }
+    function buildMemberStore(llvm:Builder builder, Scaffold scaffold, llvm:Value value, bir:Register reg) {
+        return buildStoreInt(builder, scaffold, value, reg);
+    }
+};
+
+final ListRepr listFloatArrayRepr = object {
+    llvm:Type memberType = LLVM_DOUBLE;
+    string?[] rtFuncSuffix = ["float_array_get_tagged", "float_array_set_tagged", (), "float_array_set_int", "float_array_get_float", "float_array_set_float"];
+    int listDescGetIndex = 5;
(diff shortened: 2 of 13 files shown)


manuranga:
I forgot to specialize array compare. I'll work on it tomorrow.

runtime/balrt.h
```diff
@@ -81,6 +85,20 @@ typedef struct {
     GC TaggedPtr *members;
 } TaggedPtrArray;
 
+// We are assuming sizeof(void64_t) = 8
+typedef union {
+    int64_t i;
+    TaggedPtr ptr;
+    double f;
+// pr-todo: should I capitalize? picked this style to match standard types
+} void64_t;
```
jclark: Why do we need all this?

runtime/list.c
```diff
@@ -4,58 +4,77 @@
...
-
 ListPtr _bal_list_construct(ListDescPtr desc, int64_t capacity) {
     ListPtr lp = _bal_alloc(sizeof(struct List));
     lp->desc = desc;
     initGenericArray(&(lp->gArray), capacity, TAGGED_PTR_SHIFT);
     return lp;
 }
 
-static bool getFiller(ListDescPtr desc, TaggedPtr *valuePtr) {
+static bool getFiller(ListDescPtr desc, void64_t *valuePtr) {
```
jclark: Why are you changing this?

runtime/list.c
```diff
@@ -4,58 +4,77 @@
...
+
+TaggedPtr _bal_list_get(TaggedPtr p, int64_t index) {
+    return listGet(p, index, noOpTag);
+}
+
+static inline TaggedPtr tagInt(void64_t v) {
+    return intToTagged(v.i);
+}
+
+TaggedPtr _bal_list_int_get(TaggedPtr p, int64_t index) {
```
jclark: Simpler, better to have separate specialized versions.

compiler/modules/nback/init.bal
```diff
@@ -40,6 +40,15 @@ const TYPE_KIND_PRECOMPUTED = "precomputed";
 type TypeKindArrayOrMap TYPE_KIND_ARRAY|TYPE_KIND_MAP;
 type TypeKind TypeKindArrayOrMap|TYPE_KIND_RECORD|TYPE_KIND_PRECOMPUTED;
 
+// pr-todo: can use t:UT_TOP instead of `()`
+type ListAccessorType t:UT_INT|t:UT_FLOAT|();
```
jclark: I suspect things will work better with an enum corresponding to the possible representations.

```
enum ListRepresentation {
  LIST_GENERIC = "generic",
  LIST_INT_ARRAY = "int_array",
  LIST_FLOAT_ARRAY = "float_array"
}
```

Then
- have a  function that maps the inherent type to the representation
- you can construct the runtime function names using the ListRepresentation string and concatenation
- you can use maps rather than tables

compiler/modules/nback/init.bal
```diff
@@ -152,7 +160,9 @@ function addInherentTypeDefn(InitModuleContext cx, string symbol, t:SemType semT
 }
 
 function addArrayInherentTypeDefn(InitModuleContext cx, string symbol, int tid, t:UniformTypeBitSet bitSet) returns [llvm:StructType, llvm:ConstPointerValue] {
-    llvm:ConstValue initValue = cx.llContext.constStruct([llvm:constInt(LLVM_TID, tid), getListGetFunc(cx), getListSetFunc(cx), llvm:constInt(LLVM_BITSET, bitSet)]);
+    t:UniformTypeCode? utCode = t:uniformTypeCode(bitSet);
```
jclark: addArrayInherentTypeDefn only gets called once for each InherentType.  So it's only going to get called at most once for int array and at most once for float array.  So this call can unconditionally declare the get/setFuncs.  You don't need all these tables.  The int/float and generic cases are not parallel: the generic case will get called for multiple inherent types; the specialized case for one each.

compiler/modules/nback/structure.bal
```diff
@@ -83,21 +83,35 @@ function buildListConstruct(llvm:Builder builder, Scaffold scaffold, bir:ListCon
 
         // Cases that are not UniformTypeBitSet should have been filtered out before
         t:UniformTypeBitSet memberType = <t:UniformTypeBitSet>t:simpleArrayMemberType(scaffold.typeContext(), listType);
-        foreach int i in 0 ..< length {
-            builder.store(check buildWideRepr(builder, scaffold, insn.operands[i], REPR_ANY, memberType),
-                          builder.getElementPtr(array, [llvm:constInt(LLVM_INT, 0), llvm:constInt(LLVM_INT, i)], "inbounds"));
+        // pr-todo: should I replace if-else with a lookup table?
+        if memberType == t:INT {
+            check buildListArray(builder, scaffold, array, LLVM_INT, insn.operands, o => buildInt(builder, scaffold, <bir:IntOperand>o));
```
jclark: I don't think we have any closures in the code base at the moment: I have been avoiding them to facilitate self-hosting.

jclark: The function argument is duplicating the information in the fourth argument.

jclark (changes requested):
As far as the C stuff goes, the changes are far more extensive than I think are required. I designed it so that the shared part would be the growing of arrays. Other parts should be specialized. That's why we have _bal_array_grow and initGenericArray taking a shift.

On the Ballerina side, I think you need to materialize the two specialized representations better. In other words, we have two cases: specialized or generic. The specialized case should be handled by have a value (record, object, string or whatever) for that specialized representation.

jclark:
It's going to be simpler just to declare the ListDesc for `int[]` and `float[]` in list.c, then init.bal just generates an alias to that.

Edited: tids are going to complicate that.  Maybe we should predefine some TIDs for well-known types.

runtime/balrt.h
manuranga: This is so that list set can be reused for all three 64-bit types. Another option is `listSet` to take `sizeof` and do calculations on `void*` of `GenericArray`.

manuranga:
I updated the `init.bal` and `structure.bal`.

runtime/balrt.h
```diff
@@ -55,14 +55,18 @@
 #endif
 #endif
 
+#define IMMEDIATE_INT_MIN -(1L << (TAG_SHIFT - 1))
```
jclark: I would use `(int64_t)1` rather than 1L as we do elsewhere.

runtime/balrt.h
```diff
@@ -271,8 +293,28 @@ extern char *_bal_string_alloc(uint64_t lengthInBytes, uint64_t lengthInCodePoin
...
+extern TaggedPtr _bal_list_generic_tagged_get(TaggedPtr p, int64_t index);
+extern int64_t _bal_list_generic_int_get(TaggedPtr p, int64_t index);
+double _bal_list_generic_float_get(TaggedPtr p, int64_t index);
+extern PanicCode _bal_list_generic_tagged_set(TaggedPtr p, int64_t index, TaggedPtr val);
+extern PanicCode _bal_list_generic_int_set(TaggedPtr p, int64_t index, int64_t val);
+PanicCode _bal_list_generic_float_set(TaggedPtr p, int64_t index, double val);
+
+extern TaggedPtr _bal_list_int_array_tagged_get(TaggedPtr p, int64_t index);
+extern int64_t _bal_list_int_array_int_get(TaggedPtr p, int64_t index);
+extern float (*const _bal_list_int_array_float_get)(TaggedPtr p, int64_t index); // Reads can't fail. Will not be called.
```
jclark: Don't see the point of this.

jclark (commented):
Just looked at the C stuff so far, and it's looking great.

Will look in more detail later.

compiler/modules/nback/types.bal
```diff
@@ -17,13 +17,25 @@ final llvm:StructType llStructureDescType = llvm:structType([LLVM_TID]);
...
-                                              heapPointerType(llvm:arrayType(LLVM_TAGGED_PTR, 0))]); // *members
+final llvm:FunctionType[6] llListDescFuncs = [
+            llvm:functionType(LLVM_TAGGED_PTR, [LLVM_TAGGED_PTR, LLVM_INT]),
+            llvm:functionType(LLVM_PANIC_CODE, [LLVM_TAGGED_PTR, LLVM_INT, LLVM_TAGGED_PTR]),
+            llvm:functionType(LLVM_INT, [LLVM_TAGGED_PTR, LLVM_INT]),
+            llvm:functionType(LLVM_PANIC_CODE, [LLVM_TAGGED_PTR, LLVM_INT, LLVM_INT]),
+            llvm:functionType(LLVM_DOUBLE, [LLVM_TAGGED_PTR, LLVM_INT]),
+            llvm:functionType(LLVM_PANIC_CODE, [LLVM_TAGGED_PTR, LLVM_INT, LLVM_DOUBLE])];
+final llvm:StructType llListDescType = llvm:structType([LLVM_TID,
+                                                        llvm:pointerType(llListDescFuncs[0]),
```
jclark: Clearer to define a const for each index.

We ought to avoid the repetitive call to llvm:pointerType.

compiler/modules/nback/structure.bal
```diff
@@ -67,6 +67,78 @@ final RuntimeFunction mappingConstructFunction = {
...
+    int listDescSetIndex = 2;
+    function buildMember(llvm:Builder builder, Scaffold scaffold, bir:Operand member, t:SemType memberType) returns llvm:Value|BuildError {
+        return buildWideRepr(builder, scaffold, member, REPR_ANY, memberType);
+    }
+    function buildMemberStore(llvm:Builder builder, Scaffold scaffold, llvm:Value value, bir:Register reg) {
+        return buildStoreTagged(builder, scaffold, value, reg);
+    }
+}
+
+readonly class IntListRepr {
```
jclark: If you are only ever creating one of a class, then it's better to use an object-constructor-expr e.g.

final ListRepr listIntArrayRepr = object {
    ListReprKind kind = LIST_FLOAT_ARRAY
    // etc

};
```

File: compiler/modules/nback/structure.bal [hunk 11]
```diff
@@ -67,6 +67,78 @@ final RuntimeFunction mappingConstructFunction = {
...
+    int listDescSetIndex = 6;
+    function buildMember(llvm:Builder builder, Scaffold scaffold, bir:Operand member, t:SemType memberType) returns llvm:Value|BuildError {
+        return buildFloat(builder, scaffold, <bir:FloatOperand>member);
+    }
+    function buildMemberStore(llvm:Builder builder, Scaffold scaffold, llvm:Value value, bir:Register reg) {
+        return buildStoreFloat(builder, scaffold, value, reg);
+    }
+}
+
+final ListRepr LIST_GENERIC_REPR = new GenericListRepr();
```
jclark: We generally use all caps only for const (or things that would be const except for jBallerina/spec bugs).

compiler/modules/nback/init.bal
```diff
@@ -279,28 +286,37 @@ function addRecordSubtypeDefn(InitModuleContext cx, string symbol, string[] fiel
...
-function getListGetFunc(InitModuleContext cx) returns llvm:FunctionDecl {
-    llvm:FunctionDecl? existing = cx.listGetFunc;
-    if existing == () {
-        llvm:FunctionDecl decl = cx.llMod.addFunctionDecl(mangleRuntimeSymbol("list_get"), llListGetFuncType);
-        cx.listGetFunc = decl;
-        return decl;
-    }
-    else {
-        return existing;
+final readonly & string[] LIST_ACCESSOR_RETURN_TYPES = ["tagged", "int", "float"];
```
jclark: This sort of thing is fine in all caps, because it's only a JBUG that is stopping it being a const.

jclark (commented):
I like this all a lot more than the first version.

runtime/balrt.h
manuranga: In `init.bal` I blindly create `FunctionDecl` for each combination, so currently I rely on this symbol existing. Alternatively, I can put a `nullptr` in `init.bal` when `ty != "tagged" && ty !=  repr.kind`.

compiler/modules/nback/structure.bal
```diff
@@ -67,6 +67,78 @@ final RuntimeFunction mappingConstructFunction = {
...
+    LIST_GENERIC = "generic",
+    LIST_INT_ARRAY = "int_array",
+    LIST_FLOAT_ARRAY = "float_array"
+}
+
+type ListRepr readonly & object {
+    ListReprKind kind;
+    llvm:Type memberType;
+    int listDescGetIndex;
+    int listDescSetIndex;
```
manuranga, line 74: I could have introduced a separate object called `ListAccessorsRepr` and move `listDescGetIndex`, `listDescSetIndex` there. But I'd have to duplicate `buildMemberStore` as well.

Since (currently?) number of `ListRepr` happens to be same as number of accessors pairs, I can reuse `ListRepr`.

compiler/modules/nback/types.bal
manuranga: Not sure I understand this suggestion. I need `llListDescFuncs` as an array for creating FunctionDesc in a loop.
I can change this to `llvm:pointerType(llListDescFuncs[listGenericArrayRepr.listDescGetIndex -  1])` and so on.

Not sure how I can avoid `llvm:pointerType`.

manuranga: I'll add a function.

manuranga: I tried following, unfortunately it's a JBUG.

```ballerina
function llListDescFuncPtr(int listReprIndex) returns llvm:PointerType {
    return llvm:pointerType(llListDescFuncs[listReprIndex - 1]); // `-1` since TID occupies the first
}

final llvm:StructType llListDescType = llvm:structType([LLVM_TID,
                                                        llListDescFuncPtr(listGenericArrayRepr.listDescGetIndex),
                                                        llListDescFuncPtr(listGenericArrayRepr.listDescSetIndex),
                                                        llListDescFuncPtr(listIntArrayRepr.listDescGetIndex),
                                                        llListDescFuncPtr(listIntArrayRepr.listDescSetIndex),
                                                        llListDescFuncPtr(listFloatArrayRepr.listDescGetIndex),
                                                        llListDescFuncPtr(listFloatArrayRepr.listDescSetIndex),
                                                        LLVM_MEMBER_TYPE])
```

jclark: Use a loop to create an array of PointerType and initialize the array members in structType from that. You're supposed to be able to use `...x` to splice in the members of the array, but IIRC we haven't got that yet.

jclark: Point is there is one call to pointerType in the source not six.

runtime/balrt.h
jclark: The problem with what you have is that the types are bogus. When you refer to the symbol _bal_list_int_array_int_get in LLVM the type will be a pointer to a function. When you refer to the symbol _bal_list_int_array_float_get, you will get a pointer to a pointer to a function.

jclark:
Is this ready for review?

manuranga:
Yes

runtime/list.c
```diff
@@ -46,14 +46,55 @@ static bool getArrayFiller(MemberType memberType, TaggedPtr *valuePtr) {
...
+
+TaggedPtr _bal_list_float_array_get_tagged(TaggedPtr p, int64_t index) {
+    ListPtr lp = taggedToPtr(p);
+    GC FloatArray *ap = &(lp->fArray);
+    return floatToTagged(ap->members[index]);
+}
+
+double _bal_list_float_array_get_float(TaggedPtr p, int64_t index) {
+    // SUBSET should only be called when subset supports float subtype arrays
+    fprintf(stderr, "warning: access float array via float type desc func pointer");
```
jclark: Should have a `\n` at the end of this and the get_int one.

jclark:
It looks good, but I would like to do a call where you talk me through the changes to structure.bal.

manuranga:
Did the changes.

jclark (approved):

PR #699, opened 2021-11-15
Support decimal mul and div

KavinduZoysa:
## Purpose
Related to #289

### DIFF
--- a/runtime/decimal.c
+++ b/runtime/decimal.c
@@ -45,25 +45,47 @@ TaggedPtrPanicCode _bal_decimal_sub(TaggedPtr tp1, TaggedPtr tp2) {
     return finish(&d, &cx);    
 }
 
+TaggedPtrPanicCode _bal_decimal_mul(TaggedPtr tp1, TaggedPtr tp2) {
+    const decQuad *dq1 = taggedToDecQuad(tp1);
+    const decQuad *dq2 = taggedToDecQuad(tp2);
+    decQuad d;
+    decContext cx;
+    initContext(&cx);   
+    decQuadMultiply(&d, dq1, dq2, &cx);
+    return finish(&d, &cx); 
+}
+
 TaggedPtrPanicCode finish(decQuad *dq, decContext *cx) {
     TaggedPtrPanicCode result;
+    enum decClass class = decQuadClass(dq);
+    if (class == DEC_CLASS_POS_ZERO || 
+        class == DEC_CLASS_NEG_ZERO || 
+        class == DEC_CLASS_POS_SUBNORMAL || 
+        class == DEC_CLASS_NEG_SUBNORMAL) {
+        decQuadZero(dq);
+        result.panicCode = 0;
+        result.ptr = createDecimal(dq);
+        return result;
+    }
     uint32_t status = cx->status;
     if (status & DECIMAL_STATUS_FAIL) {
-        // XXX decide what panic codes to use 
-        result.panicCode = (status & DEC_Division_by_zero) ? PANIC_DIVIDE_BY_ZERO : PANIC_ARITHMETIC_OVERFLOW;
+        if (status & DEC_Division_by_zero) {
+            result.panicCode = PANIC_DIVIDE_BY_ZERO;
+        }
+        else if (status & DEC_Underflow) {
+            decQuadZero(dq);
+            result.panicCode = 0;
+            result.ptr = createDecimal(dq);
+        }
+        else {
+            result.panicCode = PANIC_ARITHMETIC_OVERFLOW;
+        }
     }
     else {
         result.panicCode = 0;
-        enum decClass class = decQuadClass(dq);
-        if (class == DEC_CLASS_POS_ZERO || 
-            class == DEC_CLASS_NEG_ZERO || 
-            class == DEC_CLASS_POS_SUBNORMAL || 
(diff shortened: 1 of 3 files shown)


runtime/decimal.c
```diff
@@ -54,7 +54,10 @@ TaggedPtrPanicCode finish(decQuad *dq, decContext *cx) {
     }
     else {
         result.panicCode = 0;
-        if (decQuadClass(dq) == DEC_CLASS_NEG_ZERO) {
```
jclark, line 57: You are callling decQuadClass 4 times!

Also why you are testing for DEC_CLASS_POS_ZERO?

I don't think you can test this until you have done mul or div. So please add those with some tests.

jclark (changes requested):

runtime/decimal.c
KavinduZoysa: > You are callling decQuadClass 4 times!
> 
Fixed

> Also why you are testing for DEC_CLASS_POS_ZERO?
> 
For a case like this `-1E-6143 + 1E-6143`, expected value in subset is `0`. But if we did not handle like this we are getting  `0E-6143`. That is why I checked +Zero case too, and added a test for that

KavinduZoysa: I will add mul into this PR.

runtime/decimal.c
```diff
@@ -45,21 +45,57 @@ TaggedPtrPanicCode _bal_decimal_sub(TaggedPtr tp1, TaggedPtr tp2) {
...
+        return result;
+    }
     uint32_t status = cx->status;
     if (status & DECIMAL_STATUS_FAIL) {
-        // XXX decide what panic codes to use 
-        result.panicCode = (status & DEC_Division_by_zero) ? PANIC_DIVIDE_BY_ZERO : PANIC_ARITHMETIC_OVERFLOW;
+        if (status & DEC_Division_by_zero) {
+            result.panicCode = PANIC_DIVIDE_BY_ZERO;
+        }
+        else if (status & DEC_Underflow) {
```
jclark, line 77: Why are you doing this? What are the values in this case? Won't the class be subnormal?

runtime/decimal.c
```diff
@@ -45,21 +45,57 @@ TaggedPtrPanicCode _bal_decimal_sub(TaggedPtr tp1, TaggedPtr tp2) {
     return finish(&d, &cx);    
 }
 
+TaggedPtrPanicCode _bal_decimal_mul(TaggedPtr tp1, TaggedPtr tp2) {
+    const decQuad *dq1 = taggedToDecQuad(tp1);
```
jclark: Let's inline the calls to taggedToDecQuad in all four functions.

jclark (changes requested):

runtime/decimal.c
KavinduZoysa: Please consider this example:
`mul(9.999999999999999999999999999999999E-6001, 1E-143)`
In decQuad module, the class of the result is +Normal and value of the result is `1.000000000000000000000000000000000E-6143` because it is rounded. Also, undeflowed and inexact flags are raised.

But in subset, the class of the result is +Zero and the value of the result is `0`. Also, undeflowed, subnormal, inexact and rounded flags are raised. 

In subset, subnormal values are not permited, underflow flag is raised and value becomes zero. 
In other case, rounding was applies to the result(9.999... -> 10.000...) which is a subnormal and value becomes +Normal (9.999...E-6144 -> 10.000...E-6144 ->1.000...E-6143).

runtime/decimal.c

runtime/decimal.c
jclark: Let me see if I understand.  You are saying:

1. Multiplication happens in two logical stages: perform the multiplication to a precision greater than allowed by decimal128 format, and then round to fit into a decimal128
2. It is possible for the result of the first stage to be subnormal, but then be rounded in the second stage to become normal
3. In this case, decNumber in subset mode will return 0.
4. We can detect this case in decQuad by checking the Underflow flag.

Is that right?

If so, that's a very interesting case.  I am glad you have discovered it.

But I don't think the Ballerina spec requires us to return 0. Not allowing subnormal values means that there are no Ballerina decimal values that are subnormal. It does not constrain internal intermediate results.   My concern is that this may be very difficult to implement in jBallerina.  Can this be done on top of BigDecimal?

KavinduZoysa: > Is that right?
> 

Yes, that is correct.

> But I don't think the Ballerina spec requires us to return 0. Not allowing subnormal values means that there are no Ballerina decimal values that are subnormal. It does not constrain internal intermediate results. My concern is that this may be very difficult to implement in jBallerina. Can this be done on top of BigDecimal?

@jclark, I have tried it here https://github.com/KavinduZoysa/decimal/blob/master/decimal.java. I am not sure whether this is the correct approach doing this using BigDecimal. In BigDecimal, setting context to DECIMAL128 only limit the number of digits to 34 and does not limit the value of exponent. So I think we can compare the result with minimum magnitute to detect the subnormal case.

jclark: That will, presumably, give the same result as testing whether class is subnormal. How can it handle the case where the intermediate result (before rounding) is subnormal but the result after rounding is normal?

jclark (approved):

PR #704, opened 2021-11-17
Support decimal neg and rem

KavinduZoysa:
## Purpose
Related to #289


KavinduZoysa:
~~Please note that these changes are done on top of #699~~

runtime/decimal.c
```diff
@@ -96,6 +104,9 @@ TaggedPtrPanicCode finish(decQuad *dq, decContext *cx) {
             result.panicCode = 0;
             result.ptr = createDecimal(dq);
         }
+        else if (status & DEC_Division_impossible) {
```
jclark: Why can't we just use PANIC_ARITHMETIC_OVERFLOW for this?

runtime/tests/decimal_arith.c
```diff
@@ -135,10 +135,48 @@ void testDiv() {
...
+    genAndValidateDecNeg("9.999999999999999999999999999999998E6144", "-9.999999999999999999999999999999998E+6144");
+}
+
+void genAndValidateDecRem(const char *decStr1, const char *decStr2, const char *rem, PanicCode code) {
+    TaggedPtrPanicCode tp = _bal_decimal_rem(_bal_decimal_const(decStr1), _bal_decimal_const(decStr2));
+    validate(tp, rem, code);
+}
+
+void testRem() {
+    genAndValidateDecRem("1", "1", "0", 0);
```
jclark, line 158: Need tests whether the second operand is negative.

jclark: Make it obvious that these are giving the same results as integer division by using eg `(-10%3).toString()`

I meant: "integer remainder"

jclark (changes requested):

runtime/tests/decimal_arith.c
KavinduZoysa: Added more tests

runtime/decimal.c
KavinduZoysa: I used the error given by the library, yes we can use PANIC_ARITHMETIC_OVERFLOW, because DEC_Division_impossible happens when the operations have too many digits.
Changed.

runtime/tests/decimal_arith.c
jclark: I want to be absolutely sure that we are not getting the IEEE Remainder operation. Which are the tests that show that? Are the decNumber docs clear on this?  See this for some concrete cases 

https://docs.microsoft.com/en-us/dotnet/api/system.math.ieeeremainder?view=net-6.0

jclark:
I'm not convinced decQuadRemainder has the semantics we need.  I read the docs for decDoubleRemainder here:

http://speleotrove.com/decimal/dnfloat.html

And Ballerina remainder here:

https://ballerina.io/spec/lang/2021R1/#section_6.23

In particular, the second paragraph of the decDoubleRemainder docs does not obviously correspond to what the Ballerina spec is saying.

runtime/decimal.c
```diff
@@ -57,6 +57,14 @@ TaggedPtrPanicCode _bal_decimal_div(TaggedPtr tp1, TaggedPtr tp2) {
     return finish(&d, &cx); 
 }
 
+TaggedPtrPanicCode _bal_decimal_rem(TaggedPtr tp1, TaggedPtr tp2) {
+    decQuad d;
+    decContext cx;
+    initContext(&cx);
+    decQuadRemainder(&d, taggedToDecQuad(tp1), taggedToDecQuad(tp2), &cx);
```
jclark, line 64: Why do you think this has the semantics that Ballerina requires?

jclark:
@manuranga Have a look at this https://github.com/ballerina-platform/nballerina/pull/704#issuecomment-974618725 and see what you think.

manuranga:
DecQuad provides two operations `decQuadRemainder` and `decQuadRemainderNear`. `decQuadRemainder` is the integer remainder and `decQuadRemainderNear` is IEEE remainder. Therefor `decQuadRemainder` is closer to what Ballerina specifies. Unfortunately it not exactly what we need since it raise `DEC_Division_impossible`, when the Ballerina spec says no exceptions.

jclark:
I've created a spec issue to record this problem https://github.com/ballerina-platform/ballerina-spec/issues/998

For now let's do this:

- introduce a separate panic code corresponding to the IEEE invalid operation exception: we check for this in `finish` using `status & DEC_IEEE_754_Invalid_operation`; these correspond to cases where we would get NaN if we weren't in the subset; call this `PANIC_INVALID_DECIMAL`; errors we get from decNumber should be one of invalid operation, divide by zero, overflow or underflow; each of the first three turn into different panic codes; underflow turns into zero
- mention this issue in the subset definition: we will do this later, when we get a bit further along

runtime/decimal.c
KavinduZoysa: Relevant tests are added from L:162 to L:168 in `testRem()` function.

runtime/decimal.c
```diff
@@ -57,6 +57,20 @@ TaggedPtrPanicCode _bal_decimal_div(TaggedPtr tp1, TaggedPtr tp2) {
...
+    return finish(&d, &cx);    
+}
+
+TaggedPtrPanicCode _bal_decimal_neg(TaggedPtr tp) {
```
jclark: When can this panic? I don't think it can.

runtime/decimal.c
```diff
@@ -57,6 +57,20 @@ TaggedPtrPanicCode _bal_decimal_div(TaggedPtr tp1, TaggedPtr tp2) {
...
+    decQuad d;
+    decQuadCopyNegate(&d, taggedToDecQuad(tp));
+    return finish(&d, NULL);
```
jclark: I don't like this passing in NULL to finish. Will this work?

```
const decQuad *dp = taggedToDecQuad(tp);
if (decQuadClass(dp) == DEC_CLASS_POS_ZERO) {
    return tp;
}
decQuad d;
decQuadCopyNegate(&d, taggedToDecQuad(tp));
return createDecimal(&d);
```

runtime/decimal.c
```diff
@@ -96,6 +113,9 @@ TaggedPtrPanicCode finish(decQuad *dq, decContext *cx) {
             result.panicCode = 0;
             result.ptr = createDecimal(dq);
         }
+        else if (status & DEC_IEEE_754_Invalid_operation) {
+            result.panicCode = PANIC_INVALID_DECIMAL;
+        }
         else {
```
jclark, line 110: This should only happen with DEC_Overflow, right? The code doesn't capture that.

jclark (changes requested):

runtime/decimal.c
KavinduZoysa: Yes, this cannot panic, But to handle -Zero case and be consistent with other cases  I passed result and cx(which is NULL) to finish method.

I changed it as per the below comment.

runtime/decimal.c
KavinduZoysa: This is logically correct, but the representation of the result fails at this example : `0.0`. Since we return the tp at `0.0`, we are getting `0.0` but in the subset, we should get `0`.
I am going to do this like,
```
TaggedPtr _bal_decimal_neg(TaggedPtr tp) {
    const decQuad *dq = taggedToDecQuad(tp);
    decQuad d;
    if (decQuadClass(dq) == DEC_CLASS_POS_ZERO) {
        decQuadZero(&d);
    } 
    else {
        decQuadCopyNegate(&d, dq);
    }
    return createDecimal(&d);
}
```

runtime/decimal.c
KavinduZoysa: Yes, checked the `DEC_Overflow` and moved `DEC_IEEE_754_Invalid_operation` into the else part.

manuranga: At lease we need a comment saying `// status & DEC_IEEE_754_Invalid_operation = true` here.

Is the `if` at line 73 redundant now?

KavinduZoysa: Did you mean this check?
```
    if (decQuadClass(dq) == DEC_CLASS_POS_ZERO) {
        decQuadZero(&d);
    } 
```

manuranga: I meant `if (status & DECIMAL_STATUS_FAIL) {`. Is that check redundant now?

KavinduZoysa: I think we need to have it for the readability. In `status` these are not only the flags that raise. Except for error flags, we may get inexact and rounded flags. So I think it is better to have logic like this

```
if (Dec_error) {
// handle each error
}
else {
// No error flag, return value
}
```

runtime/decimal.c
```diff
@@ -107,6 +119,18 @@ TaggedPtrPanicCode finish(decQuad *dq, decContext *cx) {
     return result;
 }
 
+TaggedPtr _bal_decimal_neg(TaggedPtr tp) {
+    const decQuad *dq = taggedToDecQuad(tp);
+    decQuad d;
+    if (decQuadClass(dq) == DEC_CLASS_POS_ZERO) {
+        decQuadZero(&d);
```
jclark: Better to return `tp` in this case.

jclark (changes requested):

runtime/decimal.c
KavinduZoysa: @jclark, If the input of `_bal_decimal_neg` is `0.0`, we have to return `0`. Please check the 2nd point of `Simplified number set` in doc http://speleotrove.com/decimal/dax3274.html
If we return `tp`, in this case we will get `0.0`.

KavinduZoysa: Please ignore my comment. `0.0` is not a valid case at this point, it should be handled before passing to this function.

Changed to return `tp`.

jclark (approved):

PR #708, opened 2021-11-18
Fix constBitCast

### LINKED ISSUE #705: constBitCast fails with named struct when struct body has not been set
I need to call constBitCast on a named struct when I haven't yet set the body of the struct.

I get a crash here:

```
[fail] testCompileEU#'future/record2-uv.bal':

		    {ballerina/lang.map}KeyNotFound {"message":"cannot find key '%Type0'"}

		    	at ballerina.lang.map.0:get(map.bal:65)
		    	wso2.nballerina.print:getStructName(llvm.bal:251)
		    	wso2.nballerina.print.llvm.0:typeToString(llvm.bal:1444)
		    	wso2.nballerina.print.llvm.0:typeToString(llvm.bal:1438)
		    	wso2.nballerina.print.llvm.0:bitCastArgs(llvm.bal:1673)
		    	wso2.nballerina.print:constBitCast(llvm.bal:190)
```

heshanpadmasiri:
## Purpose

- Fix constBitCast fails with named struct when struct body has not been set
- Make handling of named struct types more efficient by storing the type name in type itself.

Resolves #705

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/print.llvm/llvm.bal
+++ b/compiler/modules/print.llvm/llvm.bal
@@ -207,7 +207,7 @@ public class Context {
         if self.namedStructTypes.hasKey(structName) {
             panic err:illegalArgument("this context already has a struct type by that name");
         }
-        StructType ty = { elementTypes: [] };
+        StructType ty = { elementTypes: [], name: structName };
         self.namedStructTypes[structName] = [ty, false];
         self.namedStructTypeBody[structName] = [];
         return ty;
@@ -215,16 +215,13 @@ public class Context {
 
     // Corresponds to LLVMStructSetBody
     public function structSetBody(StructType namedStructTy, Type[] elementTypes) {
-        foreach var entry in self.namedStructTypes.entries() {
-            var data = entry[1];
-            string name = entry[0];
-            StructType ty = data[0];
-            if ty === namedStructTy {
-                self.namedStructTypeBody[name] = elementTypes;
-                return;
-            }
+        string? tyName = namedStructTy.name;
+        if tyName is string && self.namedStructTypes.hasKey(tyName) {
+            self.namedStructTypeBody[tyName] = elementTypes;
         }
-        panic err:illegalArgument("no such named struct type");
+        else {
+            panic err:illegalArgument("no such named struct type");
+        }
     }
 
     // Corresponds to LLVMConstPtrToInt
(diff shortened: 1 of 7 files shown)


compiler/modules/print.llvm/llvm.bal
```diff
@@ -1442,27 +1440,31 @@ function typeToString(RetType ty, Context context, boolean forceInline=false) re
         }
     }
     else if ty is StructType {
-        var data = context.getStructName(ty);
+        string? tyName = ty.name;
+        Type[] elementTypes = ty.elementTypes;
+        if tyName != () {
+            elementTypes = context.getNamedStructBody(ty);
+        }
         if !forceInline {
```
jclark, line 1448: This ought to go before the assignment to elementTypes (since it does not depend on that)

compiler/modules/print.llvm/llvm.bal
```diff
@@ -242,18 +239,19 @@ public class Context {
...
     }
 
-    function getStructName(StructType ty) returns [string, Type[]?]? {
-        foreach var entry in self.namedStructTypes.entries() {
-            var data = entry[1];
-            if data[0] === ty {
+    function getNamedStructBody(StructType ty) returns Type[] {
+        string? tyName = ty.name;
+        if tyName is string {
+            var data = self.namedStructTypes[tyName];
```
jclark, line 245: You are over-using var.  If the type isn't too complicated, write the type not `var`: it makes the code easier to understand.

jclark (commented):
I'll merge this to get the build working. But please fix these points in a follow-up pull request.

I also don't understand how you are dealing with named structs. Can you explain your implementation approach?

heshanpadmasiri:
> I'll merge this to get the build working. But please fix these points in a follow-up pull request.
> 
> I also don't understand how you are dealing with named structs. Can you explain your implementation approach?

For "normal" struct types the `StructType` record has all the details we need to work with it. But for "named" struct types in addition to having a name (thus it behave differently when we output) we need the fallowing while keeping the `StructType` record itself readonly.

1. ability to change element types after the creation of `StructType`
2. keep track of if the type is used (so we can avoid unused struct types in the output)

This is achieved by actually keeping the element types (and whether it is used) in the context. (For named struct type [`elementTypes` array](https://github.com/ballerina-platform/nballerina/blob/2f1e5da2cf09c5a664bd07946b1e638a02faa9cf/compiler/modules/print.llvm/common.bal#L42) is always ignored). So whenever we need the element types we first check if it is a named struct type and if so we get the element types from the context (this tells the context that type is actually used). Only if it is not a named struct type we care about the `elementTypes` array. This means as far the internals of llvm module is concerned there are actually two different struct types and this probably leading to lot of confusion. @jclark would it be better if we make this implicit separation more explicit by splinting the `StructType` internally while keeping the same `StructType` public interface. (draft of the idea :#709)

PR #709, opened 2021-11-18
Rework named struct type

heshanpadmasiri:
## Purpose
Currently we actually have two different kinds of struct types

1. Those created using `structType` function (These don't have a name and their element types can't be changed after creation) 
2. Those created used `structCreateNamed` function in context (These have a name and their element types can be changed after creation)

In addition to differences in how we create them they are different based on how we output them. 1st type always inline all the element types where as second one only uses the name. While current implementation of `StructType` is sufficient to deal with requirements of type 1, requirements of type 2 is achieved using the Context (This is because we need `StructType` itself to be read-only  so we can't change the element types after creation). Context keep track of element types (So for the 2nd type `elementTypes` array is meaningless) and whenever we need the element types we get them via the context. In addition to keeping track of element types context also keep track of whether a given named struct type is used or not (To make sure only used types will end up in the output).

This implicit separation of `StructType` inside the llvm module leads to various unclear subtleties that can be avoided by explicitly spiting the `StructType` (While maintaining all the public interfaces as `StructType` such that users don't need to know the internal implementation)

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/print.llvm/common.bal
+++ b/compiler/modules/print.llvm/common.bal
@@ -38,9 +38,14 @@ public function arrayType(Type ty, int elementCount) returns ArrayType {
 }
 
 // Corresponds to llvm::StructType
-public type StructType readonly & record {
+public type StructType UnnamedStructType|NamedStructType;
+
+public type UnnamedStructType readonly & record {
     Type[] elementTypes;
-    string? name = ();
+};
+
+public type NamedStructType readonly & record {
+    string name;
 };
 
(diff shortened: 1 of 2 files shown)


jclark:
Maybe the value returned by structCreateNamed shouldn't be readonly?

Edited: going to be difficult to make that work, but it's pretty weird to be able to do setBody on something that's readonly. For concurrency reasons, we need the type of all final module-level variables to be readonly.

PR #712, opened 2021-11-19
Introduce fixed length array type

### LINKED ISSUE #502: Implement fixed length array types
Logically equivalent to a tuple.

rdhananjaya:
## Purpose
Fixes #502 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/types/list.bal
```diff
@@ -126,8 +160,61 @@ function listSubtypeIsEmpty(Context cx, SubtypeData t) returns boolean {
     return isEmpty;    
 }
 
+function memberAt(ListMemberType m, int i) returns SemType {
```
jclark: The scope of this is the whole module: every module-level symbol in list.bal should start with `list`.

compiler/modules/front.syntax/ast.bal
```diff
@@ -447,6 +447,7 @@ public type ListTypeDesc record {|
     Position endPos;
     TypeDesc[] members;
     TypeDesc rest;
+    SingletonTypeDesc? length = ();
```
jclark: `length` should be a SimpleConstExpr.

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -69,8 +69,14 @@ function parsePostfixTypeDesc(Tokenizer tok) returns TypeDesc|err:Syntax {
         }
         else if tok.current() == "[" {
             check tok.advance();
+            Token? c = tok.current();
+            SingletonTypeDesc? length = ();
+            // todo: this only allow int literals to be within brackets, should be allow any numeric literal and then show appropriate error later?
```
jclark: You've got to allow a reference to a const, which you have to resolve during resolveTypes.

compiler/modules/front/resolveTypes.bal
```diff
@@ -148,9 +148,16 @@ function resolveTypeDesc(ModuleSymbols mod, s:ModuleLevelDefn modDefn, int depth
             }
             t:ListDefinition d = new;
             td.defn = d;
-            t:SemType[] members = from var x in td.members select check resolveTypeDesc(mod, modDefn, depth + 1, x);
-            t:SemType rest = check resolveTypeDesc(mod, modDefn, depth + 1, td.rest);
-            return d.define(env, members, rest);
+            var tdLen = td.length?.value;
```
jclark: Don't use `var` in a case like this: it makes the code harder to understand.

This bit is really ugly and repetitive.

compiler/testSuite/semtype/fixed-length-array-t.bal
```diff
@@ -0,0 +1,29 @@
...
+type ArrayOfIntArray int[][];
+
+// @type ArrayOfInt5 < ArrayOfIntArray
+// @type Int5 <> ArrayOfInt5
+type ArrayOfInt5 int[][5];
+
+// @type Array5OfInt5 < ArrayOfInt5
+// @type Array5OfInt5 < ArrayOfIntArray
+type Array5OfInt5 int[5][5];
+
```
manuranga, line 23: don't we need examples with `int[5][]`

compiler/modules/front.syntax/astString.bal
```diff
@@ -284,7 +284,10 @@ function typeDescToWords(Word[] w, TypeDesc td, boolean|BinaryTypeOp wrap = fals
         if wrap != false {
             w.push("(");
         }
-        typeDescToWords(w, td.rest, true);
+        TypeDesc? rest = td.rest;
+        if rest != () {
+            typeDescToWords(w, rest, true);
+        }
         w.push(CLING);
         w.push("[", "]");
```
manuranga: I think you'll need to serialize length here, please add some test cases to `testParser.json`.

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -69,8 +69,19 @@ function parsePostfixTypeDesc(Tokenizer tok) returns TypeDesc|err:Syntax {
         }
         else if tok.current() == "[" {
             check tok.advance();
+            SimpleConstExpr? length = ();
```
manuranga:
```suggestion
            SimpleConstExpr? length;
```

manuranga: I think it's better if below two lines are also non-init declarations.

compiler/modules/front.syntax/tests/testPos.bal
```diff
@@ -462,8 +462,11 @@ function validateTypeDescPos(TypeDesc td, Tokenizer tok, Position parentStartPos
             check validateTypeDescPos(member, tok, td.startPos, td.endPos);
             childNodePos.push([member.startPos, member.endPos]);
         }
-        check validateTypeDescPos(td.rest, tok, td.startPos, td.endPos);
-        childNodePos.push([td.rest.startPos, td.rest.endPos]);
+        TypeDesc? rest = td.rest;
+        if (rest != ()) {
```
manuranga: extra parans.

compiler/modules/front/resolveTypes.bal
```diff
@@ -148,9 +148,20 @@ function resolveTypeDesc(ModuleSymbols mod, s:ModuleLevelDefn modDefn, int depth
             }
             t:ListDefinition d = new;
             td.defn = d;
+            s:SimpleConstExpr? size = td.length;
+            int length = 0;
+            if size != () {
+                [t:SemType, t:Value] [_, value] = check resolveConst(mod, modDefn, size, t:INT);
+                length = <int>value.value;
+                length = length - 1; // the member in members array is alredy counted toward member length, we need to adjust for that.
```
manuranga: Better to inline `- 1` to above.

compiler/modules/front/resolveTypes.bal
```diff
@@ -148,9 +148,20 @@ function resolveTypeDesc(ModuleSymbols mod, s:ModuleLevelDefn modDefn, int depth
             }
             t:ListDefinition d = new;
             td.defn = d;
+            s:SimpleConstExpr? size = td.length;
+            int length = 0;
+            if size != () {
+                [t:SemType, t:Value] [_, value] = check resolveConst(mod, modDefn, size, t:INT);
```
manuranga: Better to pick another name, how about `resolved` ?

compiler/modules/front.syntax/astString.bal
rdhananjaya: ahh didn't notice that, I'll fix this and add test cases like `int[x] v = xxx` to `testParser.json`

compiler/modules/front.syntax/astString.bal
```diff
@@ -282,9 +282,20 @@ function typeDescToWords(Word[] w, TypeDesc td, boolean|BinaryTypeOp wrap = fals
...
-        typeDescToWords(w, td.rest, true);
+        TypeDesc? rest = td.rest;
+        if rest != () {
+            typeDescToWords(w, rest, true);
+        }
         w.push(CLING);
-        w.push("[", "]");
+        SimpleConstExpr? length = td.length;
+        if length != () {
+            w.push("[", CLING);
```
manuranga: You may not need `CLING` here, `[` auto clings.

compiler/modules/front.syntax/astString.bal
```diff
@@ -282,9 +282,20 @@ function typeDescToWords(Word[] w, TypeDesc td, boolean|BinaryTypeOp wrap = fals
...
+            exprToWords(w, length);
+            w.push(CLING, "]");
+        }
+        else {
+            w.push("[", "]");
```
manuranga: I would factor this out of if-else

compiler/modules/front/const.bal
```diff
@@ -79,24 +79,29 @@ function resolveConstDefn(ModuleSymbols mod, s:ConstDefn defn) returns s:Resolve
...
+        return resolvedConst;
+    }
+}
+
+function resolveConst(ModuleSymbols mod, s:ModuleLevelDefn defn, s:Expr origExpr, t:SemType? expectedType) returns s:ResolvedConst|FoldError {
+    ConstFoldContext cx = new ConstFoldContext(defn, mod);
+    s:Expr expr = check foldExpr(cx, expectedType, origExpr);
+    if expr is s:ConstValueExpr {
+        if expectedType == () || t:containsConst(expectedType, expr.value) {
+            s:ResolvedConst r = [t:singleton(expr.value), { value: expr.value }];
```
manuranga: need spaces around `[`

rdhananjaya: Would something like this work here?
```ballerina
            return <s:ResolvedConst> [ 
                t:singleton(expr.value), 
                { value: expr.value } 
                ];
```

manuranga: I meant
```suggestion
            s:ResolvedConst r = [ t:singleton(expr.value), { value: expr.value } ];
```

jclark: That's not the style I have used: I put spaces inside curly braces not square or round. Yes, I know this is arbitrary.

manuranga: Ahh yes. please disregard.

rdhananjaya: The line breaks was because it felt like too many spaces.

compiler/modules/front.syntax/astString.bal
rdhananjaya: As in extract a function using this `if-else` block?

manuranga: No, I meant putting `w.push("[");` before `if` and `w.push("]");` after, and getting rid of `else`.

compiler/modules/types/core.bal
```diff
@@ -802,8 +802,8 @@ public function listAtomicSimpleArrayMemberType(ListAtomicType? atomic) returns
     return ();   
 }
 
-final ListAtomicType LIST_ATOMIC_TOP = { members: [], rest: TOP };
-final ListAtomicType LIST_ATOMIC_READONLY = { members: [], rest: READONLY };
+final ListAtomicType LIST_ATOMIC_TOP = { members: [], repeatLastMember: 0, rest: TOP };
+final ListAtomicType LIST_ATOMIC_READONLY = { members: [],  repeatLastMember: 0, rest: READONLY };
```
jclark: Extra space before repeatLastMember

compiler/modules/types/list.bal
```diff
@@ -88,13 +92,18 @@ function readOnlyListAtomicType(ListAtomicType ty) returns ListAtomicType {
...
     };   
 }
 
 public function tuple(Env env, SemType... members) returns SemType {
     ListDefinition def = new;
-    return def.define(env, members, NEVER);
+    return def.define(env, members, 0, NEVER);
+}
+
+public function isFixedLengthList(ListAtomicType t) returns boolean {
```
jclark: Instead of this, have

```
public function listAtomicMinLength(ListAtomicType t) returns int {
    return t.members.length() + t.repeatLastMember;
}
```

and use below also.

compiler/modules/types/list.bal
```diff
@@ -126,17 +135,73 @@ function listSubtypeIsEmpty(Context cx, SubtypeData t) returns boolean {
     return isEmpty;    
 }
 
+function listMemberAt(SemType[] members, int i) returns SemType {
+    return members[int:min(i, members.length() - 1)];
+}
+
+function fixLength(int newAllowedIndex, SemType[] members, int repeatCount, SemType withNewMember) returns int {
```
jclark: The scope of this function is the whole module. Functions in this file should be named starting with `list`.

This needs a comment saying what it is doing.

compiler/modules/types/list.bal
```diff
@@ -275,15 +351,16 @@ function listAtomicMemberType(ListAtomicType atomic, int? key) returns SemType {
...
+        else if key < atomic.members.length() + atomic.repeatLastMember {
+            return listMemberAt(atomic.members, key);
         }
         return atomic.rest;
     }
     SemType m = atomic.rest;
     foreach var ty in atomic.members {
         m = union(m, ty);
     }
+
```
jclark: Added line.

compiler/modules/front.syntax/astString.bal
rdhananjaya: Oh, I didn't see that at that time, fixed, thanks.

compiler/modules/types/list.bal
```diff
@@ -2,12 +2,16 @@
 
 public type ListAtomicType readonly & record {|
     SemType[] members;
+    // Repeat the last member of `members` this number of times more.
+    // { members: [int], repeatLastMember: 2, rest: NEVER } is same as { members: [int, int, int], repeatLastMember: 0, rest: NEVER }
+    // Must be zero, if members array is empty.
+    int repeatLastMember;
```
jclark: Looking at the code below, you need often to treat the combination of `members` and `repeatLastMember` as a unit: these two are representing what used to be represented by just members.  So I think everything will be simpler, if you do this:

```
public type ListAtomicTypeMembers readonly & record {|
   SemType[] initial;
   int repeatLastCount;
|};

public type ListAtomicType readonly & record {|
   ListAtomicTypeMembers members;
   SemType rest;
|};
```

compiler/modules/types/list.bal
jclark: With ListAtomicTypeMembers, this would be `listMembersMinLength(ListAtomicTypeMembers) returns int`

compiler/modules/types/list.bal
jclark: And a better name that conveys what it is doing.

compiler/modules/front/const.bal
```diff
@@ -79,24 +79,31 @@ function resolveConstDefn(ModuleSymbols mod, s:ConstDefn defn) returns s:Resolve
...
-            else {
-                return err:semantic(`initializer of ${defn.name} is not a subtype of the declared type`, s:defnLocation(defn));
-            }
+        s:ResolvedConst resolvedConst = check resolveConst(mod, defn, defn.expr, expectedType);
+        defn.resolved = resolvedConst;
+        return resolvedConst;
+    }
+}
+
+function resolveConst(ModuleSymbols mod, s:ModuleLevelDefn defn, s:Expr origExpr, t:SemType? expectedType) returns s:ResolvedConst|FoldError {
```
jclark: Call this `resolveConstExpr`. The argument should be `expr` not `origExpr`.

jclark (changes requested):

jclark:
> With ListAtomicTypeMembers, this would be `listMembersMinLength(ListAtomicTypeMembers) returns int` 

If it's a function that takes a ListAtomicTypeMembers, then it should be listMembersLength rather than listMembersMinLength.

compiler/modules/types/list.bal
```diff
@@ -126,17 +142,65 @@ function listSubtypeIsEmpty(Context cx, SubtypeData t) returns boolean {
...
 }
 
+function listMemberAt(SemType[] members, int i) returns SemType {
+    return members[int:min(i, members.length() - 1)];
+}
+
+function listDecompressMembersForSet(int setIndex, SemType[] members, int repeatCount, SemType withNewMember) returns int {
+    int memberLen = members.length();
+    boolean lastMemberRepeats = repeatCount != 0;
+    if lastMemberRepeats {
```
rdhananjaya: We could merge this `if-else` block and the `if-else` block below this. I felt that keeping the early return bits separate will make the method cleaner.

compiler/modules/types/list.bal
```diff
@@ -332,6 +285,43 @@ function listInhabited(Context cx, SemType[] m, int repeatLastCount, SemType res
...
+
+function listMemberAt(ListAtomicTypeMembers members, int i) returns SemType {
+    return members.initial[int:min(i, members.initial.length() - 1)];
+}
+
+function isSameListAcomicTypeStructure(ListAtomicTypeMembers v1, ListAtomicTypeMembers v2) returns boolean {
+    return v1.repeatLastCount == v2.repeatLastCount && v1.initial.length() == v2.initial.length();
+}
+
+function listDecompressMembersForSet(ListAtomicTypeMembers members, int setIndex, SemType withNewMember) {
```
jclark: This is just performing a set-member operation of `ListAtomicTypeMembers`; `listMemberAt` is performing a get-member. The naming of the two functions should reflect this.

rdhananjaya: 1. This doesn't set when the `setIndex` is below the length of `initial` list. Can we say this is the set-member? Would you like me to merge the `set` behavior into this function?

2. How about `listMemberGet`?

jclark: > This doesn't set when the setIndex is below the length of initial list.

Can we change things so it does?

rdhananjaya: I think we need to pass the `rest` type to this as we sometime need to fill with `rest` type too.

rdhananjaya: Will update the PR with that change

compiler/modules/types/list.bal
```diff
@@ -82,19 +90,27 @@ public class ListDefinition {
     }       
 }
 
+function createListAtomicType(SemType[] members, int repeatLastMember, SemType rest) returns ListAtomicType {
```
jclark: The first two arguments are a ListAtomicTypeMembers.

compiler/modules/types/list.bal
```diff
@@ -32,8 +40,8 @@ public class ListDefinition {
         }
     }
 
-    public function define(Env env, SemType[] members, SemType rest) returns ComplexSemType {
-        ListAtomicType rwType = { members: members.cloneReadOnly(), rest };
+    public function define(Env env, SemType[] members, int repeatLastMember, SemType rest) returns ComplexSemType {
+        ListAtomicType rwType = createListAtomicType(members, repeatLastMember, rest);
```
jclark: The first two arguments as a ListAtomicTypeMembers.

compiler/modules/types/list.bal
```diff
@@ -151,31 +167,43 @@ function listFormulaIsEmpty(Context cx, Conjunction? pos, Conjunction? neg) retu
...
                 p = p.next; 
                 lt = cx.listAtomType(d);
-                int newLen = int:max(members.length(), lt.members.length());
-                if members.length() < newLen {
+                int prevLen = listMembersLength(members);
+                int currentLen = listMembersLength(lt.members);
+                int newLen = int:max(prevLen, currentLen);
+
+                // intersect similar compressed representations without decompressing
+                if isSameListAcomicTypeStructure(members, lt.members) {
```
jclark: Spelling: Acomic

compiler/modules/types/list.bal
```diff
@@ -257,6 +284,50 @@ function listInhabited(Context cx, SemType[] members, SemType rest, Conjunction?
     }
 }
 
+function shallowCopyListAtomicType(ListAtomicTypeMembers members) returns ListAtomicTypeMembers {
```
jclark: Should have consistent naming pattern for functions (like this) operating on ListAtomicTypeMembers.

compiler/modules/types/list.bal
```diff
@@ -257,6 +284,50 @@ function listInhabited(Context cx, SemType[] members, SemType rest, Conjunction?
...
+    return { initial: shallowCopyTypes(members.initial), repeatLastCount: members.repeatLastCount };
+}
+
+function listMemberGet(ListAtomicTypeMembers members, int index) returns SemType {
+    return members.initial[int:min(index, members.initial.length() - 1)];
+}
+
+function isSameListAcomicTypeStructure(ListAtomicTypeMembers v1, ListAtomicTypeMembers v2) returns boolean {
```
jclark: Should use consistent naming patterns for functions providing operations on ListAtomicTypeMembers.

compiler/modules/types/list.bal
```diff
@@ -238,11 +265,11 @@ function listInhabited(Context cx, SemType[] members, SemType rest, Conjunction?
...
         foreach int i in 0 ..< len {
-            SemType ntm = i < negLen ? nt.members[i] : nt.rest;
-            SemType d = diff(members[i], ntm);
+            SemType ntm = i < negLen ? listMemberGet(nt.members, i) : nt.rest;
+            SemType d = diff(listMemberGet(members, i), ntm);
             if !isEmpty(cx, d) {
-                SemType[] s = shallowCopyTypes(members);
-                s[i] = d;
+                ListAtomicTypeMembers s = shallowCopyListAtomicType(members);
+                s.initial[i] = d;
```
jclark: I would replace these two lines by `ListAtomicMembers s = listAtomicMemberReplace(members, i, d)`: you are creating a new ListAtomicMembers like `members` but with the i-th member replaced by d.

jclark (changes requested):

compiler/modules/types/list.bal
rdhananjaya: Would `listMembersShallowCopy` work here? We already have `shallowCopyTypes` and `shallowCopyStrings`.

jclark:
> Would `listMembersShallowCopy` work here? We already have `shallowCopyTypes` and `shallowCopyStrings`. 

Use the same prefix as for the other functions on ListAtomicMembers followed by either "Copy" or "ShallowCopy".

This is slightly different from the other two cases.  In this case, we are introducing a data-type specific to this file, and providing functions on it. In the other case, we are providing functions on a more generic data-type used generally.

compiler/modules/types/list.bal
```diff
@@ -188,29 +182,52 @@ function listFormulaIsEmpty(Context cx, Conjunction? pos, Conjunction? neg) retu
     return !listInhabited(cx, members, rest, neg);
 }
 
+function listIntersectWith(FixedLengthArray members, SemType rest, ListAtomicType lt) returns [FixedLengthArray, SemType]? {
+    int ltLen = lt.members.fixedLength;
+    int newLen = int:max(members.fixedLength, ltLen);
+    if members.fixedLength < newLen {
```
rdhananjaya, line 205: @jclark As an optimization we can handle `members.initial.length() == lt.members.initial.length()` case without pre-filling `members`.  Should we?

jclark: Do whatever is least code first, and put in a comment about possible future optimization.

rdhananjaya:
@jclark I'm working on adapting this to ordered types. What should the result of `listAtomicMemberType` [1] be for a `int[0]` when `key = ()`? Is it semtype nil or semtype `int`?

[1] - https://github.com/ballerina-platform/nballerina/pull/712/files#diff-62808c0208460483ed51b0c9dad7a0ed348b50daa9630db5b7f88f01910039b1R337

manuranga:
@rdhananjaya Please note that we are simplifying ordering in #756, better to do on top it.

jclark:
`int[0]` is the same as `[]` so it would have to be NEVER.

compiler/modules/types/list.bal
```diff
@@ -1,13 +1,23 @@
...
 public type ListAtomicType readonly & record {|
-    SemType[] members;
+    readonly & FixedLengthArray members;
     SemType rest;
 |};
 
+// Represent a fixed length semtype member list similar to a tuple.
+// The length of the list is `fixedLength`, the last member of the `initial` is repeated to achive this semantic.
+// { initial: [int], fixedLength: 3, } is same as { initial: [int, int, int], fixedLength: 3 }
+// { initail: [string, int], fixedLength: 100 } means `int` is repeated 99 times to get a list of 100 members.
```
jclark: Spelling

compiler/modules/types/list.bal
```diff
@@ -1,13 +1,23 @@
...
+// `fixedLength` must be at least `initial.length()`
```
jclark: Also need to say that if initial.length() is 0 then fixedLength must be 0.

compiler/modules/types/list.bal
```diff
@@ -257,6 +272,55 @@ function listInhabited(Context cx, SemType[] members, SemType rest, Conjunction?
...
+    arr.fixedLength = newLen;
+}
+
+function fixedArrayGet(FixedLengthArray members, int index) returns SemType {
+    int memberLen = members.initial.length();
+    int i = int:min(index, memberLen - 1);
+    return members.initial[i];
+}
+
+function listMemberAt(FixedLengthArray fixedArray, SemType rest, int index) returns SemType {
```
jclark: Let's move this function to just before all the fixedLength ones.

compiler/modules/types/list.bal
```diff
@@ -257,6 +272,55 @@ function listInhabited(Context cx, SemType[] members, SemType rest, Conjunction?
...
+    }
+        
+    SemType lastMember = members.initial[initCount - 1]; 
+    foreach int i in initCount ..< setIndex + 1 {
+        members.initial.push(lastMember);
+    }
+    members.initial[setIndex] = m;
+}
+
+function fixedArrayWith(FixedLengthArray array, int index, SemType newMember) returns FixedLengthArray {
```
jclark: I guess you are not using `replace` because it sounds like it is mutating. I am not wild about `with`. If we cannot find a good name, then maybe better to inline it, since it is for simple and only called once.

jclark (changes requested):
I have only looked at `list.bal` so far, but that's now looking really nice.  We are going to update docs/type-subset.md also.

compiler/modules/front.syntax/ast.bal
```diff
@@ -462,7 +462,8 @@ public type ListTypeDesc record {|
     Position startPos;
     Position endPos;
     TypeDesc[] members;
-    TypeDesc rest;
+    TypeDesc? rest;
```
jclark, line 463: We need to have separate ArrayTypeDesc and TupleTypeDesc types. These are too different now.

compiler/modules/front/resolveTypes.bal
```diff
@@ -142,14 +142,32 @@ function resolveTypeDesc(ModuleSymbols mod, s:ModuleLevelDefn modDefn, int depth
     if td is s:ListTypeDesc {
         t:ListDefinition? defn = td.defn;
         if defn == () {
+            // todo: Ask James, what to do about this not implemented thing?
```
jclark: We still need to give unimplemented error.

compiler/modules/types/list.bal
```diff
@@ -32,7 +42,7 @@ public class ListDefinition {
         }
     }
 
-    public function define(Env env, SemType[] members, SemType rest) returns ComplexSemType {
+    public function define(Env env, FixedLengthArray members, SemType rest) returns ComplexSemType {
         ListAtomicType rwType = { members: members.cloneReadOnly(), rest };
```
jclark, line 47: Somewhere we need to deal with normalizing the FixedLengthArray by chopping duplicates off the end of the `initial` array.  Feels like this should be done here with a fixedLengthNormalize function.

jclark (changes requested):

compiler/modules/types/list.bal
rdhananjaya: Inlined in `list.bal` and `listProj.bal`

compiler/modules/front.syntax/ast.bal
```diff
@@ -455,14 +455,19 @@ public type TypeDefn record {|
...
+    *PositionFields;
     TypeDesc[] members;
-    TypeDesc rest;
+    TypeDesc? rest;
+    t:ListDefinition? defn = ();
+|};
+
+public type ArrayTypeDesc record {|
+    *PositionFields;
+    TypeDesc elementTypeDesc;
```
jclark: In Ballerina, we use `member` not `element`. I would call this `member` to be consistent with `members` in `TupleTypeDesc`.

compiler/modules/front.syntax/ast.bal
```diff
@@ -455,14 +455,19 @@ public type TypeDefn record {|
...
+    SimpleConstExpr?[] arrayLen = [];
```
jclark: The name needs improving. I don't think `array` adds value, since this is ArrayTypeDesc. We also try to keep the name of array-valued field to be plural. Another quibble is that nil isn't a length.

I suggest `dimensions` following the spec https://ballerina.io/spec/lang/master/#section_5.4.1.1

compiler/modules/front/resolveTypes.bal
```diff
@@ -148,8 +148,33 @@ function resolveTypeDesc(ModuleSymbols mod, s:ModuleLevelDefn modDefn, int depth
...
+        if defn == () {
+            t:ListDefinition d = new;
+            td.defn = d;
+            t:SemType t =  check resolveTypeDesc(mod, modDefn, depth + 1, td.elementTypeDesc);
+            foreach s:SimpleConstExpr? len in td.arrayLen.reverse() {
+                if len == () {
+                    t = d.define(env, { initial: [], fixedLength: 0 }, t);
+                }
+                else {
+                    [t:SemType, t:Value] [_, resolved] = check resolveConstExpr(mod, modDefn, len, t:INT);
```
jclark: Need to give an error if !mod.allowAllTypes.

docs/type-subset.md
```diff
@@ -23,7 +23,7 @@ intersection-td :=
 
 postfix-td := 
   primary-td
-  | postfix-td "[" "]"
+  | postfix-td "[" array-len "]"
```
jclark: This isn't right with new semantics for dimensions. Adapt what I have done here https://ballerina.io/spec/lang/master/#section_5.4.1.1

compiler/testSuite/12-nested/order-v.bal
```diff
@@ -11,8 +11,7 @@ public function main() {
     EmptyTuple e2 = [];
     io:println(e1 < e2); // @output false
 
-
     boolean[0] b0 = [];
     float[0] f0 = [];
```
jclark: Fixed length arrays not part of the subset yet. They will get added with tuples in subset14.

compiler/modules/types/list.bal
```diff
@@ -82,19 +92,37 @@ public class ListDefinition {
...
+    }
+    SemType last = initial[i];
+    i -= 1;
+    while i >= 0 {
+        if last != initial[i] {
+            break;
+        }
+        i -= 1;
+    }
+    return { initial: initial.slice(0, i+2), fixedLength: array.fixedLength };
```
jclark: Spaces around `+`.

jclark (changes requested):

compiler/modules/types.json/parse.bal
```diff
@@ -114,7 +114,7 @@ function parseCompoundType(t:Env env, Binding? b, string k, json[] jlist, Path p
             else {
                 rest = members.pop();
             }
-            return def.define(env, members, rest);
+            return def.define(env,{ initial: [], fixedLength: 0 }, rest);
```
jclark: Missing space,

compiler/modules/types.json/parse.bal
```diff
@@ -98,7 +98,7 @@ function parseCompoundType(t:Env env, Binding? b, string k, json[] jlist, Path p
             }
             t:ListDefinition def = new;
             t:SemType[] members = check parseTypes(env, consDefBinding(jlist, def, b), jlist, parent, 1);
-            return def.define(env, members, t:NEVER);
+            return def.define(env, { initial: members, fixedLength: 0 }, t:NEVER);
```
jclark: fixedLength: members.length()

jclark:
In ListDefinition.define, how about giving defaults to all except the Env argument: initial=[], fixedLength=(), rest=t:NEVER. Then when calling use named arguments to specify the ones that are needed.

compiler/modules/front/resolveTypes.bal
```diff
@@ -148,8 +148,36 @@ function resolveTypeDesc(ModuleSymbols mod, s:ModuleLevelDefn modDefn, int depth
...
+            foreach s:SimpleConstExpr? len in td.dimensions.reverse() {
+                if len == () {
+                    t = d.define(env, rest = t);
+                }
+                else {
+                    if !mod.allowAllTypes {
+                        return err:unimplemented("fixed length array types not implemented", s:locationInDefn(modDefn, td.startPos));
+                    }
+                    [t:SemType, t:Value] [_, resolved] = check resolveConstExpr(mod, modDefn, len, t:INT);
+                    t = d.define(env, initial = [t], fixedLength = <int>resolved.value, rest = t:NEVER);
```
jclark: Can default `rest`.

compiler/modules/front/resolveTypes.bal
```diff
@@ -148,8 +148,36 @@ function resolveTypeDesc(ModuleSymbols mod, s:ModuleLevelDefn modDefn, int depth
...
+        else {
+            return defn.getSemType(env);
+        }   
+    }    
+    if td is s:ArrayTypeDesc {
+        t:ListDefinition? defn = td.defn;
+        if defn == () {
+            t:ListDefinition d = new;
+            td.defn = d;
+            t:SemType t =  check resolveTypeDesc(mod, modDefn, depth + 1, td.member);
```
jclark: Extra space

compiler/modules/front/resolveTypes.bal
jclark: The cast bothers me a bit, but I don't see a great solution.

compiler/testSuite/future/order-uv.bal
```diff
@@ -11,8 +11,7 @@ public function main() {
     EmptyTuple e2 = [];
     io:println(e1 < e2); // @output false
 
-
     boolean[0] b0 = [];
     float[0] f0 = [];
-    io:println(b1 < f0); // @output false
+    io:println(b0 < f0); // @output false
```
jclark, line 16: Why is this file changed?

compiler/testSuite/semtype/fixed-length-array-t.bal
```diff
@@ -0,0 +1,61 @@
+type IntArray int[];
```
jclark, line 1: I would like to have tests for `int[X][Y][Z]` meaning `((int[Z])[Y])[X]`

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -68,10 +68,20 @@ function parsePostfixTypeDesc(Tokenizer tok) returns TypeDesc|err:Syntax {
             td = bin;
         }
         else if tok.current() == "[" {
-            check tok.advance();
-            Position endPos = check tok.expectEnd("]");
-            ListTypeDesc list = { startPos, endPos, members: [], rest: td };
-            td = list;
+            SimpleConstExpr?[] dimensions = [];
+            Position? endPos = (); // Init with nil as we can't prove to the compiler that first iteration will always run.
```
jclark: Can we use `while true` and check for tok.current() at the end, so we can avoid the cast?

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -68,10 +68,20 @@ function parsePostfixTypeDesc(Tokenizer tok) returns TypeDesc|err:Syntax {
...
+                check tok.advance();
+                if tok.current() == "]" {
+                    dimensions.push(());
+                } 
+                else {
+                    dimensions.push(check parseSimpleConstExpr(tok));
+                }
+                endPos = check tok.expectEnd("]");
+            }
+            ArrayTypeDesc array = { startPos, endPos: <int>endPos, member: td , dimensions };
```
jclark: If we keep the cast, should be Position not int.

compiler/modules/types/listProj.bal
```diff
@@ -52,31 +52,28 @@ function listProjPath(Context cx, int k, Conjunction? pos, Conjunction? neg) ret
                 Atom d = p.atom;
                 p = p.next; 
                 lt = cx.listAtomType(d);
-                int newLen = int:max(members.length(), lt.members.length());
-                if members.length() < newLen {
+                int newLen = int:max(members.fixedLength, lt.members.fixedLength);
```
jclark: Can we use listIntersectWith here?

docs/type-subset.md
```diff
@@ -47,6 +47,8 @@ error-td := "error" [type-param]
 // XXX handle "..." in tuple-td and record-td
 tuple-td := "[" opt-td-list "]"
 
+array-dimension := "[" [simple-const-expr] "]"
```
jclark: I would put this immediately after postfix-td

compiler/modules/front/resolveTypes.bal
```diff
@@ -148,8 +148,36 @@ function resolveTypeDesc(ModuleSymbols mod, s:ModuleLevelDefn modDefn, int depth
...
+            foreach s:SimpleConstExpr? len in td.dimensions.reverse() {
+                if len == () {
+                    t = d.define(env, rest = t);
+                }
+                else {
+                    if !mod.allowAllTypes {
+                        return err:unimplemented("fixed length array types not implemented", s:locationInDefn(modDefn, td.startPos));
+                    }
+                    [t:SemType, t:Value] [_, resolved] = check resolveConstExpr(mod, modDefn, len, t:INT);
```
jclark: Localize the cast by adding to const.bal a function resolveConstIntExpr that returns int|FoldError.

jclark (changes requested):

compiler/testSuite/future/order-uv.bal
rdhananjaya: `b1` in line:17 is undefined

compiler/modules/types/listProj.bal
rdhananjaya: oh yes, I'll replace these with `listIntersectWith`

compiler/modules/front.syntax/parseTypeDesc.bal
rdhananjaya: Because we have a `if` block before the initialization of `endPos` jBallerina is going to report an un-initialized variable.

jclark: This is a jBallerina issue, so as discussed let's have a JBUG comment

compiler/modules/types/list.bal
```diff
@@ -151,31 +180,14 @@ function listFormulaIsEmpty(Context cx, Conjunction? pos, Conjunction? neg) retu
...
+                var intersected = listIntersectWith(members, rest, lt);
+                if intersected is () {
+                    return true;
                 }
-                rest = intersect(rest, lt.rest);
+                [members, rest] = intersected;
             }
         }
-        foreach var m in members {
+        foreach var m in members.initial {
```
jclark: We have this both here and in listProj, so let's factor it into a function typeListAllEmpty in common.bal.

jclark (changes requested):

compiler/modules/types/list.bal
rdhananjaya: typeListAnyEmpty?

rdhananjaya: Since we operate on the level of `FixedLengthArray`, shouldn't this function take a `FixedLengthArray` as an argument, in that case we will have to keep the factored function in `list.bal` I think.

jclark: That's fine too. And yes, should be any not every.

PR #716, opened 2021-11-22
Add a equality cycles  test

### LINKED ISSUE #574: Make `==` deal with cycles
This is required for #571.

manuranga:
Fixes #574

### DIFF
--- a/compiler/testSuite/11-nested/eqcycle2-v.bal
+++ b/compiler/testSuite/11-nested/eqcycle2-v.bal
@@ -1,0 +1,21 @@
+import ballerina/io;
+
+type Part Part[];
+
+public function main() {
+    Part yin = [];
+    Part yang = [];
+    yin[0] = yang;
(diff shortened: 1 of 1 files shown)


compiler/testSuite/11-nested/eqcycle2-v.bal
```diff
@@ -0,0 +1,21 @@
...
+
+    Part father = [];
+    Part son = [];
+    Part sprit = [];
+    father[0] = son;
+    son[0] = sprit;
+    sprit[0] = father;
+    io:println(father == son); // @output true
+
+    io:println(father == yin); // @output true
```
manuranga, line 20: Is this correct ? makes sense based on shape, but constructed differently.

jclark: Yes: `==` is testing same shape.

Issue #664, opened 2021-11-03
Modify Makefiles to enable build directory to be separate from source directory


jclark:
jBallerina is adding support for controlling `target` in slbeta4, so we can now do this properly.

https://github.com/ballerina-platform/ballerina-lang/issues/32295

PR #720, opened 2021-11-23
Migrate to jBallerina swan lake beta4

### LINKED ISSUE #644: Fix breaking changes with jBallerina Swan Lake Beta4

manuranga:
Fixes  #644 and migrate to beta4

### DIFF
--- a/compiler/Makefile
+++ b/compiler/Makefile
@@ -1,18 +1,19 @@
 COMPILER_SRC=$(wildcard *.bal) $(wildcard modules/*/*.bal)
-COMPILER_JAR=target/bin/nballerina.jar
+COMPILER_JAR=../build/bin/nballerina.jar
 BAL?=bal
 BAL_BUILD_FLAGS ?= --offline
 
 all: $(COMPILER_JAR)
 
 test:
-	$(BAL) build $(BAL_BUILD_FLAGS)
+	$(BAL) test $(BAL_BUILD_FLAGS)
 
 $(COMPILER_JAR): $(COMPILER_SRC)
-	$(BAL) build $(BAL_BUILD_FLAGS) --skip-tests
+	$(BAL) build --target-dir ../build $(BAL_BUILD_FLAGS)
 
 clean:
 	$(BAL) clean
+	rm -rf $(COMPILER_OUT)
 
 .PHONY: all test clean
 
--- a/llvm_jni/build.gradle
+++ b/llvm_jni/build.gradle
@@ -86,7 +86,7 @@ testBuild.dependsOn(copyNbal, copyJNILLVM, copyDriver, copyDriverOutput, copyLoc
 
 task balCompile(type: Exec) {
         workingDir tmpBuildDir
-        commandLine 'bal', 'build', '--offline' , '--skip-tests'
+        commandLine 'bal', 'build', '--offline'
 }
 balCompile.dependsOn(copyNbal, copyJNILLVM, copyDriver, copyDriverOutput, copyLocalDependencies, copyBalConfig)
 
@@ -115,7 +115,7 @@ cleanBuildArtificats.mustRunAfter(buildWithRuntime)
 
 task buildNBalCompilerBase(type: Exec) {
         workingDir compilerDir
-        commandLine 'bal', 'build', '--skip-tests'
+        commandLine 'bal', 'build'
 }
 
 
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   test:
     env:
-      JBAL_VERSION: swan-lake-beta3
+      JBAL_VERSION: swan-lake-beta4
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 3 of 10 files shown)


manuranga:
It's odd that this PR is passing, since I didn't remove `--skip-tests`. I am looking into it. @KavinduZoysa when you tried migration I saw you removed it, did it fail back then?

jclark:
It just gives a warning if you use `--skip-tests`.

KavinduZoysa:
I removed it because skipping tests is the default case. Sorry I do not remember what I got with `--skip-tests`.

manuranga:
Answering my own question, it is an error, but `$(COMPILER_JAR)` target is never invoked, that is why we did't face any errors, but testerina tests were not running. Now corrected.

compiler/Makefile
```diff
@@ -5,11 +5,13 @@ BAL_BUILD_FLAGS ?= --offline
 
 all: $(COMPILER_JAR)
 
-test:
+# need to run `bal build` after `bal test` since `bal test` removes the jar
```
manuranga: Another option is to trigger `bal build` from `nballerina/test/Makefile`. But since in future (hopefully) `bal test` will produce a jar. So we can mock that behavior for now.

llvm_jni/build.gradle
```diff
@@ -86,7 +86,7 @@ testBuild.dependsOn(copyNbal, copyJNILLVM, copyDriver, copyDriverOutput, copyLoc
 
 task balCompile(type: Exec) {
         workingDir tmpBuildDir
-        commandLine 'bal', 'build', '--offline' , '--skip-tests'
```
manuranga, line 89: These seems to be not in use, I removed just in case. @heshanpadmasiri Please do a cleanup when you get a chance.

compiler/Makefile
```diff
@@ -1,18 +1,22 @@
...
 BAL?=bal
 BAL_BUILD_FLAGS ?= --offline
 
 all: $(COMPILER_JAR)
 
-test:
-	$(BAL) build $(BAL_BUILD_FLAGS)
+# `bal test` dosn't produce a jar, but `test` target of `test/Makefile` needs it
+# therefor this depends on `$(COMPILER_JAR)`
+test: $(COMPILER_JAR)
```
jclark: Better to handle this in the top-level Makefile: the `test` target of this make does not depend on the compiler jar.

Here's how I think it should work to minimize total elapsed time with lots of cores:

Top-level `make all` should do `make all` in compiler and runtime in parallel.

Top-level `make test` should do sequentially
1. `make all`
2. `make test` in runtime, compiler and test in parallel

I also think it's better for `make test` to use the different directory (something like `compiler/testTarget`) -- that way both `bal build` and `bal test` will produce something that `../test` can use if we want to run things manually.

jclark (changes requested):

compiler/Makefile
manuranga: Ok, let me change it that way. But if beta5 starts wiring the jar during `bal test`, we'll have to change back.
`bal test --target-dir` don't seem to work, I'll create an issue.
It might be good to change the default target dir via `[build-options]` in `Ballerina.toml`, I'll experiment and see.

jclark: If `bal test --target-dir` doesn't work, then we will have to use `bal build --target-dir`.  Only problem is that means we usually will have to use `make` rather than calling `bal build`.  `make` in /test should use the jar produced by `make all` in /compiler, so that `make test` in /compiler can run in parallel with `make` in /test

manuranga: I deviated slightly form the above suggestion. Top-level `make test` will run `make all` only on compiler and runtime. Otherwise all tests get compiled first. It didn't feel right to spending time compiling higher subsets when there is an error in the lower.

I couldn't used `[build-options]` since that will be used by the `bal test` and clean the dir.

manuranga: Current plan from jBallerina team (@azinneera) is to fix the jar deletion issue in beta5 but not the ` --target-dir` missing in `bal test` issue [(#33954)](https://github.com/ballerina-platform/ballerina-lang/issues/33954). Since `[build-options]` is already working for all, including `bal test`, I believe this is good enough for us.

We'll not be able to have different build dirs for `test` and `build`, but as long as `test` doesn't override the jar, it shouldn't be an issue.

jclark: Does `[build-options]` have to be specified in a file in the source directory?  If so, that's not really enough, since we want to be able to build in a user-specified build directory without changing anything in the source directory.

manuranga: Yes, it need to be specified in `Ballerina.toml`. I imagined being able to specify a fixed dir outside source is enough for us to implement #664. Are you planing something beyond that?

jclark: Build systems (e.g. cmake, autotools) usually allow the user to build into a directory they choose at build time, without touching the source directory.  This means in particular you can have multiple build trees from the same source tree. So, no, it's not enough for #664.  When we do #664, we should do it once, properly.

PR #721, opened 2021-11-23
Support decimal cmp and exact_eq

KavinduZoysa:
## Purpose
Related to #289

### DIFF
--- a/runtime/decimal.c
+++ b/runtime/decimal.c
@@ -1,3 +1,5 @@
+#include <string.h>
+
 #include "balrt.h"
 #include "third-party/decNumber/decQuad.h"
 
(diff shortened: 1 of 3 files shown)


runtime/decimal.c
```diff
@@ -107,6 +109,42 @@ TaggedPtrPanicCode finish(decQuad *dq, decContext *cx) {
...
+    decQuad dc1;
+    decQuad dc2;
+    decQuadCanonical(&dc1, taggedToDecQuad(tp1));
+    decQuadCanonical(&dc2, taggedToDecQuad(tp2));
+    if (memcmp(&dc1, &dc2, DECQUAD_Bytes) == 0) {
+        return true;
+    }
+    else {
+        return false;
+    }
```
manuranga: `return memcmp(&dc1, &dc2, DECQUAD_Bytes) == 0;`

runtime/decimal.c
```diff
@@ -107,6 +109,37 @@ TaggedPtrPanicCode finish(decQuad *dq, decContext *cx) {
     return result;
 }
 
+bool _bal_decimal_exact_eq(TaggedPtr tp1, TaggedPtr tp2) {
+    decQuad dc1;
```
jclark: Should be d1, d2 for consistency with other functions.

runtime/decimal.c
```diff
@@ -107,6 +109,37 @@ TaggedPtrPanicCode finish(decQuad *dq, decContext *cx) {
...
+    decQuadCanonical(&dc2, taggedToDecQuad(tp2));
+    return memcmp(&dc1, &dc2, DECQUAD_Bytes) == 0;
+}
+
+int64_t _bal_decimal_cmp(TaggedPtr tp1, TaggedPtr tp2) {
+    decQuad d;
+    decContext cx;
+    initContext(&cx);  
+    decQuadCompare(&d, taggedToDecQuad(tp1), taggedToDecQuad(tp2), &cx);
+    // The return value of decQuadCompare can be 1, 0, -1 or NaN.
```
jclark: Why are we using decClass rather than simply returning the result of decQuadCompare?

jclark (changes requested):

runtime/decimal.c

runtime/decimal.c
KavinduZoysa: @jclark, I am returning `int64_t` to be consistent with other runtime cmp functions.
Since `decQuadCompare` returning decQuad as its' result, I used decClass.

Is that okay to return result of `decQuadCompare` which is a decQuad?

runtime/decimal.c
```diff
@@ -129,6 +131,37 @@ TaggedPtr _bal_decimal_neg(TaggedPtr tp) {
...
+    // NaN is returned only if lhs or rhs is a NaN.
+    // In _bal_decimal_cmp, both arguments are valid decimals.
+    // Therefore NaN is not an expected output.
+    // Since the result is a decQuad, decQuadClass is used to
```
jclark: Move the last two lines of the comment to be under the first line and combind

```
The return value of decQuadCompare can be 1, 0, -1 or NaN;
since this is represented as a decQuad, decQuadClass is used to
identify 1, 0, -1.
```

jclark (changes requested):
I understand now. The comment confused me.

runtime/decimal.c

PR #724, opened 2021-11-25
Support float to decimal conversion

KavinduZoysa:
## Purpose
Related to #289


runtime/decimal.c
```diff
@@ -107,6 +109,28 @@ TaggedPtrPanicCode finish(decQuad *dq, decContext *cx) {
     return result;
 }
 
+TaggedPtrPanicCode _bal_decimal_from_float(double val) {
+    TaggedPtrPanicCode result;
+    if (isnan(val) || isinf(val)) {
+        result.panicCode = PANIC_INVALID_CONVERSION;
```
jclark: We don't need a separate error here.  nan should produce PANIC_INVALID_DECIMAL; inf should produce PANIC_ARITHMETIC_OVERFLOW.

jclark (changes requested):

runtime/decimal.c

runtime/decimal.c
```diff
@@ -129,6 +131,32 @@ TaggedPtr _bal_decimal_neg(TaggedPtr tp) {
...
+        result.panicCode = PANIC_INVALID_DECIMAL;
+        return result;
+    }
+    if (isinf(val)) {
+        result.panicCode = PANIC_ARITHMETIC_OVERFLOW;
+        return result;
+    }
+    result.panicCode = 0;
+    decQuad d;
+    if (val == 0) {
```
jclark: Better to use `0.0` here rather than relying on usual C conversions.

jclark (changes requested):
Looks good.

runtime/tests/decimal_arith.c
```diff
@@ -178,6 +178,36 @@ void testRem() {
...
 }
 
+void validateFloatToDec(double val, const char *decStr, PanicCode code) {
+    validate(_bal_decimal_from_float(val), decStr, code);
+}
+
+void testFloatToDec() {
+    validateFloatToDec(0, "0", 0);
+    validateFloatToDec(0.0, "0", 0);
+    validateFloatToDec(1, "1.0", 0);
```
jclark: It is better to use .0 in all of these so the C type of the literal is double not int.

runtime/tests/decimal_arith.c
```diff
@@ -178,6 +178,36 @@ void testRem() {
...
+    validateFloatToDec(1234567890, "1234567890.0", 0);
+    validateFloatToDec(1234567890123456, "1234567890123456.0", 0);
+    validateFloatToDec(1234567890123456.1, "1234567890123456.0", 0);
+    validateFloatToDec(1234567890123456.2, "1234567890123456.3", 0);
+    validateFloatToDec(1234567890123456.9, "1234567890123457.0", 0);
```
jclark: Why is this correct?

runtime/decimal.c
```diff
@@ -129,6 +131,32 @@ TaggedPtr _bal_decimal_neg(TaggedPtr tp) {
...
+    if (val == 0.0) {
+        decQuadZero(&d);
+        result.ptr = createDecimal(&d);
+        return result;
+    }
+    char str[EMYG_DTOA_BUFFER_LEN];
+    emyg_dtoa_non_special(val, str);
+    decContext cx;
+    initContext(&cx);
+    decQuadFromString(&d, str, &cx);
```
jclark, line 211: We need to think about what the right precision is and whether this will give it to us.

jclark (changes requested):

jclark:
There's an issue we need to consider. What is the right precision for the result and how do we get that precision?

KavinduZoysa:
> There's an issue we need to consider. What is the right precision for the result and how do we get that precision?

@jclark, since the precision of double is 16 digits(Table 3.2, page 8 https://irem.univ-reunion.fr/IMG/pdf/ieee-754-2008.pdf), can't we use that? 
~~The output of `dtoa` method always has 16-17 digits which is equal to the precision of double.~~
Edited:
The output of `dtoa` method always has 16-17 digits, when the number of digits exceeds the double precision.

jclark:
> The output of `dtoa` method always has 16-17 digits which is equal to the precision of double.

Really?  I thought it had the minimum number of digits needed to round-trip.

jclark:
Can we test what gcc does with _Decimal128?

KavinduZoysa:
I have edited my comment.

jclark:
I do not understand your edited comment.

In any case, I think the precision should be the number of digits needed to round trip.  For example, a Ballerina float value equal to 2 (or 2.0 or 2.00), should turn into decimal with 1 significant digit (i.e. decimal to string) should be `2`. I don't think it would be user friendly to turn it into a decimal with the precision being the maximum precision of float, since that would lead to `2` being printed as `2.0000000000000000`.

KavinduZoysa:
I have attached some outputs of _Decimal128

| double | _Decimal64 | _Decimal128 |
| ------------- | ------------- | ------------- |
| 123  | 123 | 123 |
| 1234567890123456  | 1234567890123456 | 1234567890123456 |
| 12345678901234567  | 1.234567890123457e+16 | 12345678901234568 |
| DBL_MAX(1.7976931348623157e+308) | 1.797693134862316e+308 | 1.797693134862315708145274237317044e+308 |

According to these results, _Decimal128 exceeds the maximum precision of double only if the input has more than 16 digits.

KavinduZoysa:
> In any case, I think the precision should be the number of digits needed to round trip. For example, a Ballerina float value equal to 2 (or 2.0 or 2.00), should turn into decimal with 1 significant digit (i.e. decimal to string) should be `2`. I don't think it would be user friendly to turn it into a decimal with the precision being the maximum precision of float, since that would lead to `2` being printed as `2.0000000000000000`.

I understood now. As I understood, the current implementation does the same thing as you explained here. If the input double value is `2`, `dtoa` function converts it to C string "2.0", decNumber convert it to `2.0`.

runtime/tests/decimal_arith.c
KavinduZoysa: If the result is `0`, we should represent it as `0` according to the subset. All other places are changed to have `.0`.

KavinduZoysa:
> There's an issue we need to consider. What is the right precision for the result and how do we get that precision?

The method `emyg_dtoa_non_special` does not exceed the max precision(16-17 digits). It gives the digits which si less than or equal max precision.

runtime/decimal.c
KavinduZoysa: https://github.com/ballerina-platform/nballerina/pull/724#issuecomment-981325798

runtime/tests/decimal_conv.c
```diff
@@ -51,6 +48,36 @@ void testDecToFloat() {
     validatDecToFloat("-1.7976931348623156e+309", -INFINITY);
 }
 
+void validateFloatToDec(double val, const char *decStr, PanicCode code) {
+    validate(_bal_decimal_from_float(val), decStr, code);
+}
+
+void testFloatToDec() {
+    validateFloatToDec(0.0, "0", 0);
+    validateFloatToDec(1.0, "1.0", 0);
```
jclark: The output should not have `.0`.

jclark (changes requested):

runtime/tests/decimal_conv.c
KavinduZoysa: @jclark, in [subset](http://speleotrove.com/decimal/dax3274.html), it says,
```
If the coefficient in a result has the value zero, then the sign is set to 0 and (unless the operation is quantize) the exponent is set to 0.
```

So as I understood, we can trim the `.0` only if the value is zero.
I checked this, by generating the decString giving `1.0` in the decNumber library which is compiled for subset and I got `1.0` as the output.

Also the output of `dtoa` function is 1.0, if the input double is 1.0.

jclark: We decided that the correct output for float to decimal was to have the minimum number digits needed to round-trip the value.  Therefore there should never a trailing `.0`.

KavinduZoysa: Changed

runtime/tests/decimal_conv.c
```diff
@@ -68,7 +68,38 @@ void testIntToDec() {
...
+    validateFloatToDec(1.00000, "1", 0);
+    validateFloatToDec(1234567891.0, "1234567891", 0);
+    validateFloatToDec(1234567890123456.0, "1234567890123456", 0);
+    validateFloatToDec(1234567890123456.1, "1234567890123456", 0);
+    validateFloatToDec(1234567890123456.2, "1234567890123456.3", 0);
+    validateFloatToDec(1234567890123456.9, "1234567890123457", 0);
+    validateFloatToDec(1.7976931348623157e+308, "1.7976931348623157E+308", 0);
+    validateFloatToDec(1.7976931348623156e+308, "1.7976931348623155E+308", 0);
+    validateFloatToDec(1.7976931348623157e+307, "1.7976931348623158E+307", 0);
+    validateFloatToDec(1.7976931348623157e+308*10, "", PANIC_ARITHMETIC_OVERFLOW);
```
jclark: Don't see the point of all these: we're not testing double. We just need to test double +ve/-ve infinity. So use something that is obviously +ve/-ve infinity 1.0/0.0, -1.0/0.0. Also need to test NaN.

runtime/tests/decimal_conv.c
```diff
@@ -68,7 +68,38 @@ void testIntToDec() {
...
+}
+
+void testFloatToDec() {
+    validateFloatToDec(0.0, "0", 0);
+    validateFloatToDec(1.0, "1", 0);
+    validateFloatToDec(1.00000, "1", 0);
+    validateFloatToDec(1234567891.0, "1234567891", 0);
+    validateFloatToDec(1234567890123456.0, "1234567890123456", 0);
+    validateFloatToDec(1234567890123456.1, "1234567890123456", 0);
+    validateFloatToDec(1234567890123456.2, "1234567890123456.3", 0);
```
jclark: Why is it correct to have the `.3` here?

jclark (changes requested):

runtime/tests/decimal_conv.c
KavinduZoysa: Changed

runtime/tests/decimal_conv.c
KavinduZoysa: Basically, this is the string output we get from `dtoa` function. But this does not violate the round-trip behavior. I have changed the tests to check round-trip.

PR #725, opened 2021-11-25
Support decimal to float conversion

KavinduZoysa:
## Purpose
Related to #289

### DIFF
--- a/runtime/decimal.c
+++ b/runtime/decimal.c
@@ -1,5 +1,10 @@
+<<<<<<< HEAD
 #include <string.h>
 
+=======
+#include <math.h>
+#include <float.h>
+>>>>>>> Add decimal to float conversion
 #include "balrt.h"
 #include "third-party/decNumber/decQuad.h"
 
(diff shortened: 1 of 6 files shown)


runtime/decimal.c
```diff
@@ -129,6 +131,21 @@ TaggedPtr _bal_decimal_neg(TaggedPtr tp) {
...
+    decQuadToString(d, dblStr);
+    double dbl;
+    // It is garanteed that dblStr has the correct format,
+    // because decQuadToString returns scientific notation.
+    // Therefore it is not needed to handle the EOF.
+    sscanf(dblStr, "%lf", &dbl);
+    if (isfinite(dbl)) {
+        return dbl;
+    }
+    return decQuadClass(d) == DEC_CLASS_POS_NORMAL ? DBL_MAX : -DBL_MAX;
```
KavinduZoysa: Instead of `isinf`, decQuadClass is used here, because `isinf` is cannot identify -inf over +inf in macos.

runtime/decimal.c
```diff
@@ -129,6 +131,21 @@ TaggedPtr _bal_decimal_neg(TaggedPtr tp) {
...
 
+double _bal_decimal_to_float(TaggedPtr tp) {
+    char dblStr[DECQUAD_String];
+    const decQuad *d = taggedToDecQuad(tp);
+    decQuadToString(d, dblStr);
+    double dbl;
+    // It is garanteed that dblStr has the correct format,
+    // because decQuadToString returns scientific notation.
+    // Therefore it is not needed to handle the EOF.
+    sscanf(dblStr, "%lf", &dbl);
```
jclark: Why aren't you using `strtod`?

runtime/decimal.c
jclark: Why do you need to do something special if `isfinite` is false?

jclark (changes requested):

jclark:
The tests here should be in a separate file decimal_conv.c.

runtime/decimal.c
KavinduZoysa: It is not needed, Removed.

runtime/decimal.c
KavinduZoysa: Changed to use `strtod`

KavinduZoysa:
> The tests here should be in a separate file decimal_conv.c.

Done

PR #740, opened 2021-11-29
Support int to decimal conversion

KavinduZoysa:
## Purpose
Related to #289

### DIFF
--- a/runtime/decimal.c
+++ b/runtime/decimal.c
@@ -171,15 +171,17 @@ TaggedPtr _bal_decimal_from_int(int64_t val) {
     decQuad d;
     if (INT32_MIN <= val && val <= INT32_MAX) {
         decQuadFromInt32(&d, (int32_t)val);
-        return createDecimal(&d);
+    } 
+    else {
+#define STR_CONVERT(x) #x
+#define STR(x) STR_CONVERT(x)
+#define INT64_MAX_LEN sizeof(STR(INT64_MIN))
+        char intStr[INT64_MAX_LEN];
+        sprintf(intStr, "%" PRId64, val);
+        decContext cx;
+        initContext(&cx);
+        decQuadFromString(&d, intStr, &cx);
     }
-    // The INT64_MIN(-9223372036854775808) has the maximum string length.
-    // It has 20 characters. Adding +1 for NULL terminator.
-    char intStr[21];
-    sprintf(intStr, "%" PRId64, val);
-    decContext cx;
-    initContext(&cx);
-    decQuadFromString(&d, intStr, &cx);
     return createDecimal(&d);
 }
 
--- a/runtime/decimal.c
+++ b/runtime/decimal.c
@@ -167,6 +167,22 @@ double _bal_decimal_to_float(TaggedPtr tp) {
     return strtod(dblStr, NULL);
 }
 
+TaggedPtr _bal_decimal_from_int(int64_t val) {
+    decQuad d;
+    if (INT32_MIN <= val && val <= INT32_MAX) {
+        decQuadFromInt32(&d, (int32_t)val);
+        return createDecimal(&d);
+    }
+    // The INT64_MIN(-9223372036854775808) has the maximum string length.
+    // It has 20 characters. Adding +1 for NULL terminator.
+    char intStr[21];
+    sprintf(intStr, "%" PRId64, val);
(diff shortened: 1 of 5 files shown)


runtime/decimal.c
```diff
@@ -167,6 +167,20 @@ double _bal_decimal_to_float(TaggedPtr tp) {
     return strtod(dblStr, NULL);
 }
 
+TaggedPtr _bal_decimal_from_int(int64_t val) {
+    decQuad d;
+    if (val <= INT32_MAX && val >= INT32_MIN) {
+        decQuadFromInt32(&d, val);
+        return createDecimal(&d);
+    }
+    char intStr[21]; // max length + 1
```
jclark: How do you know 21 is enough? Use an expression that shows it.

runtime/decimal.c
```diff
@@ -167,6 +167,20 @@ double _bal_decimal_to_float(TaggedPtr tp) {
     return strtod(dblStr, NULL);
 }
 
+TaggedPtr _bal_decimal_from_int(int64_t val) {
+    decQuad d;
+    if (val <= INT32_MAX && val >= INT32_MIN) {
```
jclark: Write this as in maths like:

```
INT32_MIN <= val && val <= INT32_MAX
```

runtime/decimal.c
```diff
@@ -167,6 +167,20 @@ double _bal_decimal_to_float(TaggedPtr tp) {
...
+        decQuadFromInt32(&d, val);
```
jclark: Explicit cast `val` to `int32_t`.

jclark (changes requested):

runtime/decimal.c
KavinduZoysa: Added a comment

runtime/decimal.c

runtime/decimal.c

runtime/decimal.c
```diff
@@ -167,6 +167,22 @@ double _bal_decimal_to_float(TaggedPtr tp) {
...
 
+TaggedPtr _bal_decimal_from_int(int64_t val) {
+    decQuad d;
+    if (INT32_MIN <= val && val <= INT32_MAX) {
+        decQuadFromInt32(&d, (int32_t)val);
+        return createDecimal(&d);
+    }
+    // The INT64_MIN(-9223372036854775808) has the maximum string length.
+    // It has 20 characters. Adding +1 for NULL terminator.
+    char intStr[21];
```
jclark: I want the code to be clearly correct. How do I see how many digits are in `9223372036854775808`? How do I know that is -2^63?

What I want is `21` replaced by an expression that is clearly >=the number of digits needed.

For example, I know that one decimal digit can represents at least 3 binary bits.

So I know that ceiling(64/3) + 1 for sign + 1 for zero is enough.

Or you can have a macro that does sizeof on the stringification of INT64_MIN.

jclark (changes requested):

runtime/decimal.c

runtime/decimal.c
```diff
@@ -167,6 +171,20 @@ double _bal_decimal_to_float(TaggedPtr tp) {
     return strtod(dblStr, NULL);
 }
 
+TaggedPtr _bal_decimal_from_int(int64_t val) {
+    decQuad d;
+    if (INT32_MIN <= val && val <= INT32_MAX) {
+        decQuadFromInt32(&d, (int32_t)val);
+        return createDecimal(&d);
+    }
+    char intStr[INT64_MAX_LEN];
```
jclark: Better to do `else` here and then do `createDecimal` once.

runtime/decimal.c
```diff
@@ -9,6 +9,10 @@ typedef GC decQuad *DecimalPtr;
 // XXX need to investigate DEC_Underflow
 #define DECIMAL_STATUS_FAIL DEC_Errors
 
+#define STR_CONVERT(x) #x
+#define STR(x) STR_CONVERT(x)
+#define INT64_MAX_LEN sizeof(STR(INT64_MIN))
```
jclark: Move these defines to just before you use INT64_MAX_LEN.

jclark (changes requested):

runtime/decimal.c

runtime/decimal.c

PR #741, opened 2021-11-29
Support decimal to int conversion

KavinduZoysa:
## Purpose
Related to #289

### DIFF
--- a/runtime/balrt.h
+++ b/runtime/balrt.h
@@ -281,6 +281,11 @@ typedef struct {
     PanicCode panicCode;
 } TaggedPtrPanicCode;
 
+typedef struct {
+    int64_t val;
+    bool overflow;
+} DecToIntResult;
+
 #define ALIGN_HEAP 8
 
 // Don't declare functions here if they are balrt_inline.c
@@ -358,6 +363,7 @@ extern bool _bal_decimal_exact_eq(TaggedPtr tp1, TaggedPtr tp2);
 extern double _bal_decimal_to_float(TaggedPtr tp);
 extern TaggedPtr _bal_decimal_from_int(int64_t val);
 extern TaggedPtrPanicCode _bal_decimal_from_float(double val);
+extern DecToIntResult _bal_decimal_to_int(TaggedPtr tp);
 
 // Library mangling
 #define BAL_ROOT_NAME(sym) _B04root ## sym
--- a/runtime/balrt.h
+++ b/runtime/balrt.h
@@ -282,9 +282,9 @@ typedef struct {
 } TaggedPtrPanicCode;
 
 typedef struct {
-    int64_t val;
+    int64_t value;
     bool overflow;
-} DecToIntResult;
+} IntWithOverflow;
 
 #define ALIGN_HEAP 8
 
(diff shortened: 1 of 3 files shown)


runtime/tests/decimal_conv.c
```diff
@@ -51,6 +51,44 @@ void testDecToFloat() {
...
+    validateDecToIntNonOverflow("-0.5", 0);
+    validateDecToIntNonOverflow("-1.5", -2);
+    validateDecToIntNonOverflow("-2.5", -2);
+    validateDecToIntNonOverflow("1e2", 100);
+    validateDecToIntNonOverflow("1.5e2", 150);
+    validateDecToIntNonOverflow("-1.5e2", -150);
+    validateDecToIntNonOverflow("1.51e2", 151);
+    validateDecToIntNonOverflow("1.513e2", 151);
+    validateDecToIntNonOverflow("1.515e2", 152);
+    validateDecToIntNonOverflow("9223372036854775807", 9223372036854775807);
```
jclark, line 134: Should test 9223372036854775807.4 and similar values.

runtime/balrt.h
```diff
@@ -281,6 +281,11 @@ typedef struct {
     PanicCode panicCode;
 } TaggedPtrPanicCode;
 
+typedef struct {
+    int64_t val;
+    bool overflow;
+} DecToIntResult;
```
jclark: This isn't specific to Decimal.  Call it something like `IntWithOverflow`.

runtime/balrt.h
```diff
@@ -281,6 +281,11 @@ typedef struct {
     PanicCode panicCode;
 } TaggedPtrPanicCode;
 
+typedef struct {
+    int64_t val;
```
jclark: Don't abbreviate `val` to `value` when the other field is unabbreviated and longer than `value`.

runtime/decimal.c
```diff
@@ -167,6 +168,35 @@ double _bal_decimal_to_float(TaggedPtr tp) {
     return strtod(dblStr, NULL);
 }
 
+DecToIntResult _bal_decimal_to_int(TaggedPtr tp) {
+    decQuad dQuantize;
+    decQuad dZero;
+    decQuadZero(&dZero);
+    decContext cx;
+    initContext(&cx);
+    decQuadQuantize(&dQuantize, taggedToDecQuad(tp), &dZero, &cx);
```
jclark, line 225: This is nice!

runtime/decimal.c
```diff
@@ -167,6 +168,35 @@ double _bal_decimal_to_float(TaggedPtr tp) {
...
+    DecToIntResult res;
+    if (cx.status & DEC_Invalid_operation) {
+        // The invalid operation flag is raised 
+        // when 34 digits is not enough to represent quantized decQuad.
+        // This situation can be considered as an overflow scenario.
```
jclark: Are we sure this handles rounding correctly, particularly at the limits?  Say something in the comment about this.

jclark (changes requested):

runtime/tests/decimal_conv.c
KavinduZoysa: Added more tests

runtime/balrt.h

runtime/balrt.h

runtime/decimal.c
KavinduZoysa: @jclark, did you mean handling the rounding when the `DEC_Invalid_operation` is raised?

KavinduZoysa: Updated the comment

PR #742, opened 2021-11-30
Implement relational operators on nested lists

### LINKED ISSUE #706: Fix relational operators on lists to handle arrays with arbitrary member types
At the moment, the implementation assumes that member type is a bitset (in the front-end, it uses simpleArrayMemberType).

manuranga:
Fixes #706

### DIFF
--- a/compiler/modules/bir/bir.bal
+++ b/compiler/modules/bir/bir.bal
@@ -306,14 +306,32 @@ public type ConvertToFloatInsn readonly & record {|
     Register operand;
 |};
 
-public type OrderType UniformOrderType|OptOrderType|ArrayOrderType;
-public type UniformOrderType t:UT_FLOAT|t:UT_INT|t:UT_BOOLEAN|t:UT_STRING;
-public type OptOrderType readonly & record {|
-    UniformOrderType opt;
-|};
-public type ArrayOrderType readonly & [OptOrderType];
+public type OrderType BasicOrderType|OptOrderType|ArrayOrderType;
 
-# This does ordered comparision
+public type BasicOrderType t:UT_FLOAT|t:UT_INT|t:UT_BOOLEAN|t:UT_STRING;
+
+public const NO_ORDER_TYPE = false;
+
+public type EmptyOrderType readonly & record {|
+   NO_ORDER_TYPE opt;
+|};
+
+public type NonEmptyOrderType ArrayOrderType|BasicOrderType;
+public type OptOrderType readonly & record {|
+    NonEmptyOrderType|NO_ORDER_TYPE opt;
+|};
+
+// Only used for t1 < t2 and t1 > t2  where t1, t2 are of static type `()`
+public final OptOrderType EMPTY_ORDER_TYPE = { opt: NO_ORDER_TYPE };
+
+// JBUG can't use [OrderType]
+public type ArrayOrderType readonly & record {| OrderType|NO_ORDER_TYPE memberOrderType; |};
+
+// Only used for t1 < t2 and t1 > t2 where t1, t2 are of static type `[]`
+public final ArrayOrderType EMPTY_TUPLE_ORDER_TYPE = { memberOrderType: NO_ORDER_TYPE };
+
+
+# This does ordered comparison
 # Equality and inequality are done by equal
 public type CompareInsn readonly & record {|
     *InsnBase;
--- a/compiler/modules/front/codeGen.bal
+++ b/compiler/modules/front/codeGen.bal
@@ -88,12 +88,18 @@ type LoopContext record {|
     int[] onContinueAssignments = [];
 |};
 
+type OrderTypeMemo record {|
+    readonly t:SemType semType;
+    bir:OrderType? orderType = ();
+|};
+
 class CodeGenContext {
     final ModuleSymbols mod;
     final s:SourceFile file;
     final s:FunctionDefn functionDefn;
     final bir:FunctionCode code;
     final t:SemType returnType;
+    final table<OrderTypeMemo> key(semType) orderTypeMemo = table[];
     LoopContext? loopContext = ();
 
     function init(ModuleSymbols mod, s:FunctionDefn functionDefn, t:SemType returnType) {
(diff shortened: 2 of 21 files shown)


compiler/modules/bir/bir.bal
```diff
@@ -308,12 +308,18 @@ public type ConvertToFloatInsn readonly & record {|
...
-    UniformOrderType opt;
+    ArrayOrderType|UniformOrderType|NoOrderType opt;
 |};
-public type ArrayOrderType readonly & [OptOrderType];
 
-# This does ordered comparision
+public final OptOrderType nilOrderType = { opt: NoOrderType };
+
+// JBUG can't use [OrderType]
+public type ArrayOrderType readonly & record{| OrderType memberOrderType; |};
```
jclark: Space after `record`.

compiler/modules/nback/compare.bal
```diff
@@ -165,24 +183,62 @@ final readonly & table<TaggedCompareFunction> key(op) compareFunctions = table [
...
         panic error("failed to find runtime compare function");
     }
     else {
-        TaggedCompareResultTransform transform = taggedCompareResultTransforms.get(insn.op);
-        buildStoreBoolean(builder, scaffold, builder.iCmp(transform.predicate, compareResult, llvm:constInt(LLVM_INT, transform.compareResult)), insn.result);
+        buildCompareStore(builder, scaffold, insn, lhs, rhs, comparator);
     }
 }
 
+function buildCompareSpecializedIntLit(llvm:Builder builder, Scaffold scaffold, bir:CompareInsn insn, llvm:Value lhs, llvm:Value rhs) {
```
jclark: `Lit` in the name makes no sense to me. (If it's short for "literal", that's a syntactic concept.)

compiler/modules/nback/compare.bal
```diff
@@ -165,24 +183,62 @@ final readonly & table<TaggedCompareFunction> key(op) compareFunctions = table [
...
+    llvm:BasicBlock bbExact = scaffold.addBasicBlock();
+    llvm:BasicBlock bbInexact = scaffold.addBasicBlock();
+    llvm:BasicBlock bbJoin = scaffold.addBasicBlock();
+    // SUBSET lhs and rhs must have static type `int[]`
+    llvm:Value lIsExact = builder.iCmp("ne",
```
jclark: Too much repeated code here.

compiler/modules/nback/compare.bal
```diff
@@ -165,24 +183,62 @@ final readonly & table<TaggedCompareFunction> key(op) compareFunctions = table [
...
+    builder.br(bbJoin);
+    builder.positionAtEnd(bbInexact);
+    buildCompareStore(builder, scaffold, insn, lhs, rhs, arrayGenericCompareFunction);
+    builder.br(bbJoin);
+    builder.positionAtEnd(bbJoin);
+}
+
+function buildCompareStore(llvm:Builder builder, Scaffold scaffold, bir:CompareInsn insn, llvm:Value lhs, llvm:Value rhs, RuntimeFunction comparator) {
+    TaggedCompareResultTransform transform = taggedCompareResultTransforms.get(insn.op);
+    buildStoreBoolean(builder, scaffold, builder.iCmp(transform.predicate, <llvm:Value>builder.call(scaffold.getRuntimeFunctionDecl(comparator), [lhs, rhs]), llvm:constInt(LLVM_INT, transform.compareResult)), insn.result);
```
jclark: Split this over multiple lines.

compiler/modules/front/codeGen.bal
```diff
@@ -1889,40 +1889,56 @@ function typedOperand(bir:Operand operand) returns TypedOperand? {
...
     if lot == rot {
         return lot;
     }
     return ();
 }
 
-function promoteToOptOrderType(bir:OrderType? ot) returns bir:OrderType? {
-    if ot is bir:UniformOrderType {
-        return { opt: ot };
+function promoteToMatchingOptOrderType(bir:OrderType? lot, bir:OrderType? rot) returns [bir:OrderType?, bir:OrderType?] {
```
jclark: I would call this `promoteOrderTypePair` and the next one `promoteOrderType`.

jclark: Or add in `Opt`, but make them differ in the presence of `Pair`.

compiler/modules/front/codeGen.bal
```diff
@@ -1889,40 +1889,56 @@ function typedOperand(bir:Operand operand) returns TypedOperand? {
...
+        bir:OrderType? lOpt = promoteToOptOrderType(lot);
+        return [lOpt, lOpt];
     }
-    else {
-        return ot;
+    if lot == bir:nilOrderType {
+        bir:OrderType? rOpt = promoteToOptOrderType(rot);
+        return [rOpt, rOpt];
+    }
+    // Both are bir:nilOrderType case handled before
```
jclark: Elsewhere in the code I have dealt with this sort of situation using the following pattern:

```
match [lot is bir:OptOrderType, rot is bir:OptOrderType] {
    [true, true] => {

    }
    [true, false] => {

    }
}
```

runtime/balrt.h
```diff
@@ -10,10 +10,13 @@
 #define UT_MASK 0x1F
 #define TAG_SHIFT 56
 
+#define NIL NULL
+
 #define COMPARE_UN -1
 #define COMPARE_LT 0
 #define COMPARE_EQ 1
 #define COMPARE_GT 2
+#define COMPARE(l, r) ((l > r) ? COMPARE_GT : ((l < r) ? COMPARE_LT : COMPARE_EQ))
```
jclark: Need to parenthesize the `l` and the `r` in the macro body.

Do you use the COMPARE macro other than in list.c? If not, define it there.

runtime/list.c
```diff
@@ -301,3 +301,83 @@ bool _bal_array_subtype_contains(UniformSubtypePtr stp, TaggedPtr p) {
...
+
+typedef int64_t(*TaggedValueComparator)(TaggedPtr, TaggedPtr);
+
+static READONLY TaggedValueComparator getArrayComparator(MemberType memberType) {
+    uint32_t bitSet;
+    if ((memberType & 1) == 0) {
+        return &_bal_array_generic_compare;
+    }
+    else {
+        // Clear out the nil bit since comparators can handle both nilable and non-nilble of a given type
```
jclark: Spelling is `nillable`

runtime/list.c
```diff
@@ -301,3 +301,83 @@ bool _bal_array_subtype_contains(UniformSubtypePtr stp, TaggedPtr p) {
...
+    if (lhs == rhs) {
+        return COMPARE_EQ;
+    }
+    ListPtr lhsLp = taggedToPtr(lhs);
+    ListPtr rhsLp = taggedToPtr(rhs);
+    int64_t lhsLen = lhsLp->iArray.length;
+    int64_t rhsLen = rhsLp->iArray.length;
+    int64_t length = (lhsLen <= rhsLen) ? lhsLen : rhsLen;
+
+    GC IntArray *lhsAp = &(lhsLp->iArray);
```
jclark: Better to make these two variables have type `int64_t*`

runtime/balrt.h
```diff
@@ -10,10 +10,13 @@
 #define UT_MASK 0x1F
 #define TAG_SHIFT 56
 
+#define NIL NULL
```
jclark: Define it as `((TaggedPtr)0`

runtime/list.c
```diff
@@ -301,3 +301,83 @@ bool _bal_array_subtype_contains(UniformSubtypePtr stp, TaggedPtr p) {
...
+            return COMPARE_GT;
+        }
+        else {
+            return COMPARE_LT;
+        }
+    }
+    return COMPARE(lhsLen, rhsLen);
+}
+
+typedef int64_t(*TaggedValueComparator)(TaggedPtr, TaggedPtr);
```
jclark: Space after `int64_t`.

runtime/list.c
```diff
@@ -301,3 +301,83 @@ bool _bal_array_subtype_contains(UniformSubtypePtr stp, TaggedPtr p) {
...
+            case (1 << TAG_BOOLEAN):
+                return &taggedBooleanCompare;
+            case (1 << TAG_INT):
+                return &taggedIntCompare;
+            case (1 << TAG_FLOAT):
+                return &taggedFloatCompare;
+            case (1 << TAG_STRING):
+                return &taggedStringCompare;
+        }
+        __builtin_unreachable();
```
jclark: Elsewhere I have defined macros for this sort of thing.

runtime/list.c
```diff
@@ -301,3 +301,83 @@ bool _bal_array_subtype_contains(UniformSubtypePtr stp, TaggedPtr p) {
...
+            return COMPARE(lhsLen, rhsLen);
+        }
+        else{
+            ldp = rhsLdp;
+        }
+    }
+    else {
+        ldp = lhsLdp;
+    }
+    return arrayCompare(lhs, rhs, getArrayComparator(ldp->memberType));
```
jclark, line 382: Why is `arrayCompare` not in `list.c`?

jclark (commented):
All looks pretty reasonable. I will have another look in more detail tomorrow.

runtime/list.c
manuranga: Will do. BTW if for some reason (bug) code reaches here, it doesn't panic, do we need something that actually crashes the program instead `__builtin_unreachable`?

compiler/modules/front/codeGen.bal
manuranga: That looks nice, but I am relaying on the type narrowing here.

runtime/list.c
manuranga: It was pre-existing and some refactoring needs to be done, it would make this pr difficult to understand, let me do it under #744.

runtime/balrt.h
manuranga: Will do rest under #744.

compiler/modules/front/codeGen.bal
manuranga: It worked out by mapping `false` to `nil`. BTW why can't we have `typed-binding-pattern` in `match-pattern`?

manuranga: Actually, plain pattern matching worked.

compiler/modules/bir/bir.bal
```diff
@@ -308,12 +308,20 @@ public type ConvertToFloatInsn readonly & record {|
...
+
+const NoOrderType = false;
 public type OptOrderType readonly & record {|
-    UniformOrderType opt;
+    ArrayOrderType|UniformOrderType|NoOrderType opt;
 |};
-public type ArrayOrderType readonly & [OptOrderType];
 
-# This does ordered comparision
+public final OptOrderType nilOrderType = { opt: NoOrderType };
```
jclark: For this an emptyTypeOrderType, the only reason they are not `const` is that jBallerina does not support it yet, so let's name them like consts in ALL_UPPER_CASE.

compiler/modules/front/codeGen.bal
```diff
@@ -88,12 +88,18 @@ type LoopContext record {|
...
+    bir:OrderType? orderType = ();
+|};
+
 class CodeGenContext {
     final ModuleSymbols mod;
     final s:SourceFile file;
     final s:FunctionDefn functionDefn;
     final bir:FunctionCode code;
     final t:SemType returnType;
+    final table<OrderTypeMemo> key(semType) orderTyMemo = table[];
```
jclark: If the type is `OrderTypeMemo` then the field should be `orderTypeMemo`.

compiler/modules/front/codeGen.bal
```diff
@@ -1897,40 +1903,69 @@ function typedOperand(bir:Operand operand) returns TypedOperand? {
...
+        bir:ArrayOrderType? rArrOrdTy = r == () ? () : { memberOrderType: r } ;
+        return [ lArrOrdTy, rArrOrdTy ];
+    }
+    return [lot, rot];
 }
 
-function operandIsNil(bir:Operand operand) returns boolean {
-    if operand is bir:Register {
-        return operand.semType === t:NIL;
+function promoteOrderType(bir:OrderType? ot) returns bir:OrderType? {
```
jclark: Why isn't this returning an OptOrderType?

Seems like what this is doing is going from an OrderType to an OptOrderType, so maybe name should something like `optionalizeOrderType`.

compiler/modules/front/codeGen.bal
```diff
@@ -1939,39 +1974,63 @@ final readonly & bir:UniformOrderType[] UNIFORM_ORDER_TYPES = [t:UT_BOOLEAN, t:U
...
+            return bir:nilOrderType;
+        }
+        else if tc is bir:UniformOrderType {
             return tc;
         }
     }
     return ();
 }
 
+function toOrderType(CodeGenContext cx, t:SemType operandTy) returns bir:OrderType? {
```
jclark: operandOrderType goes from an Operand to an OrderType. This looks like it's going from a SemType to an OrderType, in which case a consistent name would be `semTypeOrderType`.

compiler/modules/front/codeGen.bal
```diff
@@ -1939,39 +1974,63 @@ final readonly & bir:UniformOrderType[] UNIFORM_ORDER_TYPES = [t:UT_BOOLEAN, t:U
...
+    }
+    else {
+        result = operandUniformOrderType(operandTy);
+    }
+    memo.orderType = result;
+    return result;
+}
+
 function operandUniformOrderType(t:SemType operandType) returns bir:OrderType? {
+    if operandType == t:NIL {
```
jclark: UNIFORM_ORDER_TYPES should be defined just before this. I don't think the name here conveys what this is doing.

compiler/modules/front/codeGen.bal
```diff
@@ -1939,39 +1974,63 @@ final readonly & bir:UniformOrderType[] UNIFORM_ORDER_TYPES = [t:UT_BOOLEAN, t:U
...
+            bir:OrderType? memberOrderType = toOrderType(cx, memberTy);
+            result = memberOrderType == () ? () : { memberOrderType };
+        }
+        else {
+            result = ();
+        }
+    }
+    else if t:containsNil(operandTy) {
+        // Cast is valid since we exclude NIL resulting non-opt order type.
+        var opt = <bir:UniformOrderType|bir:ArrayOrderType|()> toOrderType(cx, t:diff(operandTy, t:NIL));
```
jclark: In bir, define a type for UniformOrderType|ArrayOrderType (maybe `NonNilOrderType`), and use that in OptOrderType and this becomes `<NonNilOrderType?>`.

compiler/modules/front/codeGen.bal
jclark: It will be clearer to have one funciton that is responsible only for memo-izing. Then put the computation of orderType into a separate function.

compiler/modules/front/codeGen.bal
```diff
@@ -1939,39 +1974,63 @@ final readonly & bir:UniformOrderType[] UNIFORM_ORDER_TYPES = [t:UT_BOOLEAN, t:U
...
+    // XXX We assume recursive types can't be OrderType, reconsider after adding tuples.
+    cx.orderTyMemo.add(memo);
+
+    bir:OrderType? result;
+    if operandTy == t:NIL {
+        result = bir:nilOrderType;
+    }
+    t:UniformTypeBitSet arrTy = t:LIST;
+    if t:isSubtypeSimple(operandTy, arrTy) {
+        t:SemType? memberTy = t:listMemberType(cx.mod.tc, operandTy);
```
jclark: Clearer to inline arrTy here.

compiler/modules/bir/bir.bal
```diff
@@ -308,12 +308,20 @@ public type ConvertToFloatInsn readonly & record {|
 
 public type OrderType UniformOrderType|OptOrderType|ArrayOrderType;
 public type UniformOrderType t:UT_FLOAT|t:UT_INT|t:UT_BOOLEAN|t:UT_STRING;
+
+const NoOrderType = false;
```
jclark: All caps.

jclark (changes requested):

compiler/modules/bir/bir.bal
```diff
@@ -308,12 +308,20 @@ public type ConvertToFloatInsn readonly & record {|
 
 public type OrderType UniformOrderType|OptOrderType|ArrayOrderType;
 public type UniformOrderType t:UT_FLOAT|t:UT_INT|t:UT_BOOLEAN|t:UT_STRING;
```
jclark: I think `BasicOrderType` would be a slightly better name.

compiler/modules/front/codeGen.bal
```diff
@@ -1939,39 +1974,63 @@ final readonly & bir:UniformOrderType[] UNIFORM_ORDER_TYPES = [t:UT_BOOLEAN, t:U
...
+        result = opt == () ? () : { opt };
+    }
+    else {
+        result = operandUniformOrderType(operandTy);
+    }
+    memo.orderType = result;
+    return result;
+}
+
 function operandUniformOrderType(t:SemType operandType) returns bir:OrderType? {
```
jclark: Use Basic rather than Uniform here to go with suggested change in bir.

Then call this semTypeBasicOrderType to go with suggested name change in toOrderType.

jclark (changes requested):

compiler/modules/front/codeGen.bal
```diff
@@ -1897,86 +1903,137 @@ function typedOperand(bir:Operand operand) returns TypedOperand? {
...
+        bir:ArrayOrderType rArrOrdTy = { memberOrderType: r } ;
+        return [ lArrOrdTy, rArrOrdTy ];
+    }
+    return [lot, rot];
 }
 
-function operandIsNil(bir:Operand operand) returns boolean {
-    if operand is bir:Register {
-        return operand.semType === t:NIL;
+function optionalizeOrderType(bir:OrderType ot) returns bir:OrderType {
```
jclark: This should return an OptOrderType. Then you would not need the cast.

compiler/modules/front/codeGen.bal
```diff
@@ -1897,86 +1903,137 @@ function typedOperand(bir:Operand operand) returns TypedOperand? {
...
-        return ot;
+    if rot == bir:EMPTY_TUPLE_ORDER_TYPE {
+        return [lot, lot];
     }
+    if lot is bir:ArrayOrderType && rot is bir:ArrayOrderType {
+        // Below casts are valid since memberOrderType is bir:NO_ORDER_TYPE case handled above
+        var [l, r] = optionalizeOrderTypePair(<bir:OrderType>lot.memberOrderType, <bir:OrderType>rot.memberOrderType);
+        bir:ArrayOrderType lArrOrdTy = { memberOrderType: l } ;
+        bir:ArrayOrderType rArrOrdTy = { memberOrderType: r } ;
+        return [ lArrOrdTy, rArrOrdTy ];
```
jclark: Too many spaces.

compiler/modules/front/codeGen.bal
```diff
@@ -1897,86 +1903,137 @@ function typedOperand(bir:Operand operand) returns TypedOperand? {
...
+        }
+        return ();
     }
-    bir:OrderType? lot = operandOrderType(cx, left);
-    bir:OrderType? rot = operandOrderType(cx, right);
-    if lot == rot {
-        return lot;
+}
+
+function optionalizeOrderTypePair(bir:OrderType lot, bir:OrderType rot) returns [bir:OrderType, bir:OrderType] {
```
jclark: This function does not always optionalize, as far as I can see. Here you are finding the order type that is common to a pair.

compiler/modules/front/codeGen.bal
```diff
@@ -1897,86 +1903,137 @@ function typedOperand(bir:Operand operand) returns TypedOperand? {
...
-    lot = promoteToOptOrderType(lot);
-    rot = promoteToOptOrderType(rot);
-    if lot == rot {
-        return lot;
+    if rot == bir:EMPTY_ORDER_TYPE {
+        bir:OrderType lOpt = optionalizeOrderType(lot);
+        return [lOpt, lOpt];
+    }
+    match [lot, rot] {
+        // Below casts are valid since lot.opt/rot.opt is bir:NO_ORDER_TYPE case handled above
```
jclark: These casts are not really satisfactory. Problem here is that EMPTY_ORDER_TYPE is not const, so narrowing cannot do its thing. So better to work around lack of const records in bir like this:

```
public type EmptyOrderType readonly & record {|
   NO_ORDER_TYPE opt;
|};

```

and then do `is bir:EmptyOrderType`, then the casts would be unnecessary (except if there's a JBUG), right?

compiler/modules/front/const.bal
```diff
@@ -369,8 +369,7 @@ function foldBinaryRelationalExpr(FoldContext cx, t:SemType? expectedType, s:Bin
         else if left is boolean && right is boolean {
             return foldedBinaryConstExpr(booleanRelationalEval(expr.relationalOp, left, right), t:BOOLEAN, leftExpr, rightExpr);
         }
-        else if (left == () && right is int|float|string|boolean)
-                || (right == () && left is int|float|string|boolean) {
+        else if left == () || right == () {
```
jclark, line 372: This isn't going to do the right thing for `() <= ()`.

jclark (changes requested):

jclark:
Let's merge this in (after fixing the issue in const.bal) and then refactor.

compiler/modules/front/const.bal
```diff
@@ -369,10 +369,10 @@ function foldBinaryRelationalExpr(FoldContext cx, t:SemType? expectedType, s:Bin
         else if left is boolean && right is boolean {
             return foldedBinaryConstExpr(booleanRelationalEval(expr.relationalOp, left, right), t:BOOLEAN, leftExpr, rightExpr);
         }
-        else if (left == () && right is int|float|string|boolean)
-                || (right == () && left is int|float|string|boolean) {
-            // () behaves like NaN
-            return foldedBinaryConstExpr(false, t:BOOLEAN, leftExpr, rightExpr);
+        else if left == () || right == () {
+            // () behaves like NaN, EQ iff `left = right = ()`
+            boolean result = expr.relationalOp is "<="|">=" ? left == right : false;
```
jclark, line 374: We will always have `left == right`, so this can be simplified.

PR #743, opened 2021-11-30
Format Diagnostic with caret

### LINKED ISSUE #695: Format Diagnostic with caret
The formatting of a diagnostic should be improved to show the range by printing the line with carets underneath (as in LLVM or GCC).

When the range is a single position, it should get the end of the range by looking for the end of the token.

This will need some additional methods on the File type.

This follows on from #693.

heshanpadmasiri:
## Purpose
Extend diagnostic format to handle cases where a single position is given by extending the carets to the end of token (if token is `"` then it will extend to cover the whole string)
Resolves #695

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/comm.diagnostic/diagnostic.bal
+++ b/compiler/modules/comm.diagnostic/diagnostic.bal
@@ -26,8 +26,7 @@ public type File readonly & object {
     public function filename() returns string;
     public function directory() returns string?;
     public function lineColumn(Position pos) returns LineColumn;
-    public function lineContent(Position pos) returns string;
-    public function tokenLength(Position startPos) returns int;
+    public function lineContent(Position|Range range) returns [string, string, string];
 };
 
 public type Location readonly & record {|
@@ -123,32 +122,24 @@ public function format(Diagnostic d) returns string[] {
     line += ": error: " + d.message;
     lines.push(line);
     Range|Position range = loc.range;
-    lines.push(range is Position ? loc.file.lineContent(range) : loc.file.lineContent(range.startPos));
+    [string, string, string] sourceLine = range is Position ? loc.file.lineContent(range) : loc.file.lineContent(range.startPos);
+    lines.push("".'join(...sourceLine));
     lines.push(caretLine(loc.file, range));
     return lines;
 }
 
 function caretLine(File file, Range|Position range) returns string {
-    int startColumn;
-    int caretLen;
-    if range is Position {
-        startColumn = file.lineColumn(range)[1];
-        caretLen = file.tokenLength(range);
-    }
-    else {
-        int startLine;
-        [startLine, startColumn] = file.lineColumn(range.startPos);
-        var [endLine, endColumn] = file.lineColumn(range.endPos);
-        if startLine == endLine {
-            caretLen = endColumn - startColumn + 1;
+    var [prefix, content, _] = file.lineContent(range);
+    string[] paddingContent = [];
+    foreach string:Char c in prefix {
+        if c == "\t" {
+            paddingContent.push(c);
         }
         else {
-            // multiline err
-            string line = file.lineContent(range.startPos);
-            caretLen = line.length() - startColumn;
+            paddingContent.push(" ");
         }
     }
-    string padding = lib:stringRepeat(" ", startColumn);
-    string carets = lib:stringRepeat("^", caretLen);
+    string padding = "".'join(...paddingContent);
+    string carets = lib:stringRepeat("^", content.length());
     return padding + carets;
 }
--- a/compiler/modules/comm.diagnostic/diagnostic.bal
+++ b/compiler/modules/comm.diagnostic/diagnostic.bal
@@ -27,6 +27,7 @@ public type File readonly & object {
     public function directory() returns string?;
     public function lineColumn(Position pos) returns LineColumn;
     public function lineContent(Position pos) returns string;
+    public function tokenLength(Position startPos) returns int;
 };
 
 public type Location readonly & record {|
@@ -128,17 +129,16 @@ public function format(Diagnostic d) returns string[] {
 }
 
 function caretLine(File file, Range|Position range) returns string {
-    string carets;
     int startColumn;
+    int caretLen;
     if range is Position {
         startColumn = file.lineColumn(range)[1];
-        carets = "^";
+        caretLen = file.tokenLength(range);
     }
     else {
         int startLine;
         [startLine, startColumn] = file.lineColumn(range.startPos);
         var [endLine, endColumn] = file.lineColumn(range.endPos);
-        int caretLen;
         if startLine == endLine {
             caretLen = endColumn - startColumn + 1;
         }
@@ -147,8 +147,9 @@ function caretLine(File file, Range|Position range) returns string {
             string line = file.lineContent(range.startPos);
             caretLen = line.length() - startColumn;
         }
-        carets = lib:stringRepeat("^", caretLen);
     }
     string padding = lib:stringRepeat(" ", startColumn);
+    string carets = lib:stringRepeat("^", caretLen);
+
     return padding + carets;
 }
--- a/compiler/modules/comm.diagnostic/diagnostic.bal
+++ b/compiler/modules/comm.diagnostic/diagnostic.bal
@@ -122,8 +122,7 @@ public function format(Diagnostic d) returns string[] {
     line += ": error: " + d.message;
     lines.push(line);
     Range|Position range = loc.range;
-    [string, string, string] sourceLine = range is Position ? loc.file.lineContent(range) : loc.file.lineContent(range.startPos);
-    lines.push("".'join(...sourceLine));
+    lines.push("".'join(...loc.file.lineContent(range)));
     lines.push(caretLine(loc.file, range));
     return lines;
 }
--- a/compiler/modules/comm.diagnostic/diagnostic.bal
+++ b/compiler/modules/comm.diagnostic/diagnostic.bal
@@ -150,6 +150,5 @@ function caretLine(File file, Range|Position range) returns string {
     }
     string padding = lib:stringRepeat(" ", startColumn);
     string carets = lib:stringRepeat("^", caretLen);
-
     return padding + carets;
 }
--- a/compiler/modules/front.syntax/scan.bal
+++ b/compiler/modules/front.syntax/scan.bal
@@ -218,7 +218,7 @@ function scanLineFragIndex(ScannedLine line, int codePointIndex) returns [int, i
     return [fragCodeIndex, fragmentIndex];
 }
 
-function scanLineContent(ScannedLine line) returns string[] {
+function scanLineFragments(ScannedLine line) returns string[] {
     string[] lineContent = [];
     readonly & FragCode[] fragCodes = line.fragCodes;
     int fragmentIndex = 0;
@@ -241,7 +241,7 @@ function scanLineContent(ScannedLine line) returns string[] {
 }
 
 function scanLineToString(ScannedLine line) returns string {
-    return "".'join(...scanLineContent(line));
+    return "".'join(...scanLineFragments(line));
 }
 
 function unicodeEscapeValue(string fragment) returns string|error {
--- a/compiler/modules/front.syntax/scan.bal
+++ b/compiler/modules/front.syntax/scan.bal
@@ -224,33 +224,30 @@ function fragCodeLength(ScannedLine line, int fragCodeIndex, int fragmentIndex) 
     }
 }
 
-function fragmentToString(ScannedLine line, int fragCodeIndex, int fragmentIndex) returns string {
-    FragCode code = line.fragCodes[fragCodeIndex];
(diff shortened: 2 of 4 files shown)


compiler/modules/front.syntax/token.bal
```diff
@@ -497,6 +497,20 @@ public readonly class SourceFile {
     function scannedLine(int lineNumber) returns ScannedLine {
         return self.lines[lineNumber - 1];
     }
+
+    public function tokenLength(Position startPos) returns int {
+        var [lineIndex, columnIndex] = unpackPosition(startPos);
+        ScannedLine line = self.scannedLine(lineIndex);
+        var [fragCodeIndex, fragmentIndex] = scanLineFragIndex(line, columnIndex);
+        if line.fragCodes[fragCodeIndex] == FRAG_STRING_OPEN {
```
jclark: There can be arbitrarily many fragments in the string.

compiler/modules/front.syntax/token.bal
```diff
@@ -497,6 +497,20 @@ public readonly class SourceFile {
...
+            if fragCodeIndex + 2 < line.fragCodes.length() && line.fragCodes[fragCodeIndex + 2] == FRAG_STRING_CLOSE {
+                return line.fragments[fragmentIndex].length() + 2;
+            }
+            // no closing quotes in this line
+            return line.fragments[fragmentIndex].length() + 1;
+        }
```
jclark, line 572: You also need an approach for dealing with FRAG_GREATER_THAN.

compiler/modules/comm.diagnostic/diagnostic.bal
```diff
@@ -27,6 +27,7 @@ public type File readonly & object {
     public function directory() returns string?;
     public function lineColumn(Position pos) returns LineColumn;
     public function lineContent(Position pos) returns string;
+    public function tokenLength(Position startPos) returns int;
```
jclark: A better way to approach this is to change `lineContent` to:

```
public function lineContent(Position|Range range) returns string[3];
```

This returns the line containing the start of the range split into 3 segments:
1. The segment before the range
2. The segment containing the range (where the range is a token when the argument is a Position)
3. The segment after the range

This will enable us to deal with tabs properly (which we don't currently): generate what you call the `padding` string by replacing every character _other than tab_ by a space character in the first segment.

jclark (changes requested):

compiler/modules/front.syntax/token.bal
heshanpadmasiri: fixed

compiler/modules/comm.diagnostic/diagnostic.bal

compiler/modules/front.syntax/token.bal

compiler/modules/comm.diagnostic/diagnostic.bal
```diff
@@ -122,33 +122,24 @@ public function format(Diagnostic d) returns string[] {
     line += ": error: " + d.message;
     lines.push(line);
     Range|Position range = loc.range;
-    lines.push(range is Position ? loc.file.lineContent(range) : loc.file.lineContent(range.startPos));
+    [string, string, string] sourceLine = range is Position ? loc.file.lineContent(range) : loc.file.lineContent(range.startPos);
```
jclark: Why aren't you just passing loc.range to lineContent? They are both `Range|Position`.

compiler/modules/front.syntax/token.bal
```diff
@@ -487,9 +487,70 @@ public readonly class SourceFile {
         return unpackPosition(pos);
     }
 
-    public function lineContent(Position pos) returns string {
-        int lineNum = self.lineColumn(pos)[0];
-        return scanLineToString(self.scannedLine(lineNum));
+    // range is expected to;the start of a token
+    public function lineContent((Position|d:Range) range) returns [string, string, string] {
```
jclark: You should be able to implement this in one iteration over the ScannedLine.  I don't think scanLineToString and scanLineFragIndex are the right primitives for this.

This needs more thought...

jclark (changes requested):

compiler/modules/comm.diagnostic/diagnostic.bal

compiler/modules/front.syntax/token.bal

compiler/modules/front.syntax/token.bal
```diff
@@ -487,9 +487,90 @@ public readonly class SourceFile {
         return unpackPosition(pos);
     }
 
-    public function lineContent(Position pos) returns string {
-        int lineNum = self.lineColumn(pos)[0];
-        return scanLineToString(self.scannedLine(lineNum));
+    // range is expected to be the start of a token
```
jclark: range will be at the start of a fragment but not necessarily the start of the token

compiler/modules/front.syntax/token.bal
```diff
@@ -487,9 +487,90 @@ public readonly class SourceFile {
...
+    public function lineContent((Position|d:Range) range) returns [string, string, string] {
+        if range is Position {
```
jclark: Why is the code for the Position case completely difference from the code for the Range case?

compiler/modules/front.syntax/token.bal
```diff
@@ -487,9 +487,90 @@ public readonly class SourceFile {
         return unpackPosition(pos);
     }
 
-    public function lineContent(Position pos) returns string {
-        int lineNum = self.lineColumn(pos)[0];
-        return scanLineToString(self.scannedLine(lineNum));
+    // range is expected to be the start of a token
+    public function lineContent((Position|d:Range) range) returns [string, string, string] {
```
jclark: Move the real work of this out into a module-level function in token.bal.

I think it will simplify things if we have a function (in scan.bal) that takes a scanLine and returns a `string[]` one for each member of fragCodes, such that scanLineToString is just the concatenation of those strings.

Then you can just iterate over this array until you the total length is what you need.

You can get rid of advanceFragmentIndex and fragmentToString.

jclark (changes requested):
lineContent still needs improvement.

compiler/modules/front.syntax/token.bal
heshanpadmasiri: comment fixed (logic already assumed it is the beginning of a fragment)

compiler/modules/front.syntax/token.bal
heshanpadmasiri: If we have a range all we have to do is, turn the line in to a string and slice it because we know the start and end positions of all three prefix, content, and suffix. But with a position what we know is where the prefix ends. We have to determine where the content ends and suffix begins.

compiler/modules/front.syntax/token.bal

compiler/modules/front.syntax/token.bal
jclark: OK.

But they are both starting off by doing the same thing:
- unpacking the first position
- getting the ScannedLine for the line it's on
- calling scanLineContent

At that point they diverge, and there should be two separate, parallel module-level functions that continue from there. The functions will have the same parameters, except that one of them will have a Position for the end.

jclark: But it's better to keep them more similar. For the case where there is no end Position, we can have a function

```
function tokenEndCodePointIndex(string[] lineContent, FragCode[] fragCodes, int tokenStartCodePointIndex) returns int;
```

Implement this using two functions similar to linePrefix and lineToken but have them return the codePointIndex of the end of the prefix and the end of the token respectively.

Order functions top down.

jclark (changes requested):

compiler/modules/front.syntax/scan.bal
```diff
@@ -218,28 +218,30 @@ function scanLineFragIndex(ScannedLine line, int codePointIndex) returns [int, i
...
-            string:Char? token = fragFixed[<int>code];
-            lineContent.push(<string>token);
+            lineContent.push(<string>fragFixed[<int>code]);
         }
     }
-    return "".'join(...lineContent);
+    return lineContent;
+}
+
+function scanLineToString(ScannedLine line) returns string {
```
jclark: Don't think we will need this.

compiler/modules/front.syntax/scan.bal
```diff
@@ -218,28 +218,30 @@ function scanLineFragIndex(ScannedLine line, int codePointIndex) returns [int, i
     return [fragCodeIndex, fragmentIndex];
 }
 
-function scanLineToString(ScannedLine line) returns string {
+function scanLineContent(ScannedLine line) returns string[] {
```
jclark: I think I would call this scanLineFragments.

(This is a bit confusing with the fragments member of ScannedLine, but we should probably fix the name of this member, since it's really only the variable fragments.)

heshanpadmasiri:
add new function `lineTokensInRange` (similar to `lineToken` used for position case) to get the content part. All (`linePrefix`, `lineToken` and `lineTokensInRange`) will return the next index of `string[]` array (of fragments) so that next function can start where they stopped.

jclark:
I was suggesting to make the two cases similar, so the only difference is how to find the end position.

heshanpadmasiri:
> I was suggesting to make the two cases similar, so the only difference is how to find the end position.

fixed with `lineTokenEndIndex` function

compiler/modules/front.syntax/token.bal
```diff
@@ -487,9 +487,16 @@ public readonly class SourceFile {
         return unpackPosition(pos);
     }
 
-    public function lineContent(Position pos) returns string {
-        int lineNum = self.lineColumn(pos)[0];
-        return scanLineToString(self.scannedLine(lineNum));
+    // range is expected to be the start of a fragment
+    public function lineContent((Position|d:Range) range) returns [string, string, string] {
+        int lineNum = range is Position ? self.lineColumn(range)[0] : self.lineColumn(range.startPos)[0];
```
jclark: Add a startPos variable and pass it to lineColumn and linePrefix.

compiler/modules/front.syntax/token.bal
```diff
@@ -499,6 +528,57 @@ public readonly class SourceFile {
...
+        if nCodePoints >= startCodePointIndex {
+            break;
+        }
+        nCodePoints += fragments[fragmentIndex].length();
+        fragmentIndex += 1;
+    }
+    return fragmentIndex;
+}
+
+// Return code-point index of end of token that starts in
```
jclark: Comment is incomplete.

compiler/modules/front.syntax/token.bal
```diff
@@ -499,6 +528,57 @@ public readonly class SourceFile {
...
+function lineTokenEndIndex(string[] fragments, FragCode[] fragCodes, int startCodePointIndex, int startFragmentIndex) returns int {
```
jclark: The name here isn't good. Let's merge this into `tokenEndCodePointInde`x`.

compiler/modules/front.syntax/token.bal
```diff
@@ -499,6 +528,57 @@ public readonly class SourceFile {
...
+                }
+                return startCodePointIndex + 2; // >>
+            }
+            return startCodePointIndex + 1; // >
+        }
+    }
+    return startCodePointIndex + fragments[startFragmentIndex].length();
+}
+
+function stringTokenEndIndex(string[] fragments, FragCode[] fragCodes, int startCodePointIndex, int startFragmentIndex) returns int {
```
jclark: Call this `stringTokenEndCodePointIndex`.

compiler/modules/front.syntax/token.bal
```diff
@@ -499,6 +528,57 @@ public readonly class SourceFile {
     }
 }
 
+function tokenEndCodePointIndex(string[] fragments, FragCode[] fragCodes, int startCodePointIndex) returns int {
+    int fragmentIndex = fragmentCountUpTo(fragments, startCodePointIndex);
+    return lineTokenEndIndex(fragments, fragCodes, startCodePointIndex, fragmentIndex);
+}
+
+function fragmentCountUpTo(string[] fragments, int startCodePointIndex) returns int {
```
jclark: Let's call the second argument `codePointIndex`.

jclark (changes requested):

compiler/modules/front.syntax/token.bal

compiler/modules/front.syntax/token.bal

compiler/modules/front.syntax/token.bal

compiler/modules/front.syntax/token.bal
heshanpadmasiri: function merged to `tokenEndCodePointIndex`

compiler/modules/front.syntax/token.bal

compiler/modules/front.syntax/scan.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/scan.bal
heshanpadmasiri: function removed

PR #748, opened 2021-12-02
Pass BIR insn position through to LLVM

### LINKED ISSUE #735: Pass BIR insn position through to LLVM
Implement DEBUG_FULL value for Options.debugLevel, which passes through position of every BIR instruction to LLVM.

heshanpadmasiri:
## Purpose
Pass BIR insn position through to LLVM
Resolves #735

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/compile.bal
+++ b/compiler/compile.bal
@@ -71,9 +71,9 @@ class CompileContext {
 }
 
 // basename is filename without extension
-function compileBalFile(string filename, string basename, string? outputBasename, nback:Options nbackOptions, OutputOptions outOptions) returns CompileError? {
+function compileBalFile(string filename, string basename, string? outputBasename, string? outDir, nback:Options nbackOptions, OutputOptions outOptions) returns CompileError? {
     CompileContext cx = new(basename, outputBasename, nbackOptions, outOptions);
-    front:ResolvedModule mod = check processModule(cx, DEFAULT_ROOT_MODULE_ID, [ {filename, directory: check sourceDir(filename, outOptions.outDir) } ], cx.outputFilename());
+    front:ResolvedModule mod = check processModule(cx, DEFAULT_ROOT_MODULE_ID, [ {filename, directory: check sourceDir(filename, outDir) } ], cx.outputFilename());
     check mod.validMain();
     check generateInitModule(cx, mod);
 }
--- a/compiler/compile.bal
+++ b/compiler/compile.bal
@@ -73,7 +73,7 @@ class CompileContext {
 // basename is filename without extension
 function compileBalFile(string filename, string basename, string? outputBasename, nback:Options nbackOptions, OutputOptions outOptions) returns CompileError? {
     CompileContext cx = new(basename, outputBasename, nbackOptions, outOptions);
-    front:ResolvedModule mod = check processModule(cx, DEFAULT_ROOT_MODULE_ID, [ {filename} ], cx.outputFilename());
+    front:ResolvedModule mod = check processModule(cx, DEFAULT_ROOT_MODULE_ID, [ {filename, directory: check sourceDir(filename, outOptions.outDir) } ], cx.outputFilename());
     check mod.validMain();
     check generateInitModule(cx, mod);
 }
@@ -167,4 +167,13 @@ function filterFuncs(front:ModuleExports defns) returns map<bir:FunctionSignatur
         }
     }
     return result;
+}
+
+function sourceDir(string filename, string? outDir) returns string|file:Error {
+    string sourceDir = check file:parentPath(filename);
+    if sourceDir == "" {
+        // file is in the current directory
+        sourceDir = file:getCurrentDir();
+    }
+    return file:relativePath(outDir ?: file:getCurrentDir(), sourceDir);
 }
--- a/compiler/compile.bal
+++ b/compiler/compile.bal
@@ -71,9 +71,9 @@ class CompileContext {
 }
 
 // basename is filename without extension
-function compileBalFile(string filename, string basename, string? outputBasename, string? outDir, nback:Options nbackOptions, OutputOptions outOptions) returns CompileError? {
+function compileBalFile(string filename, string basename, string? outputBasename, nback:Options nbackOptions, OutputOptions outOptions) returns CompileError? {
     CompileContext cx = new(basename, outputBasename, nbackOptions, outOptions);
-    front:ResolvedModule mod = check processModule(cx, DEFAULT_ROOT_MODULE_ID, [ {filename, directory: check sourceDir(filename, outDir) } ], cx.outputFilename());
+    front:ResolvedModule mod = check processModule(cx, DEFAULT_ROOT_MODULE_ID, [ {filename} ], cx.outputFilename());
     check mod.validMain();
     check generateInitModule(cx, mod);
 }
@@ -167,13 +167,4 @@ function filterFuncs(front:ModuleExports defns) returns map<bir:FunctionSignatur
         }
     }
     return result;
-}
-
-function sourceDir(string filename, string? outDir) returns string|file:Error {
-    string sourceDir = check file:parentPath(filename);
-    if sourceDir == "" {
-        // file is in the current directory
-        sourceDir = file:getCurrentDir();
-    }
-    return file:relativePath(outDir ?: file:getCurrentDir(), sourceDir);
 }
--- a/compiler/main.bal
+++ b/compiler/main.bal
@@ -13,7 +13,6 @@ public type Options record {
     boolean showTypes = false;
     int? debugLevel;
     // outDir also implies treating each file as a separate module
-    string? outDir = ();
     string? expectOutDir = ();
     string? gc = ();
     string? htmlError = ();
@@ -45,8 +44,7 @@ public function main(string[] filenames, *Options opts) returns error? {
     }
     nback:Options nbackOptions = {
         gcName: check nback:validGcName(opts.gc),
-        debugLevel: check nback:validDebugLevel(opts.debugLevel),
-        outDir: opts.outDir
+        debugLevel: check nback:validDebugLevel(opts.debugLevel)
     };
     int errorFileCount = 0;
     d:Printer dPrinter;
--- a/compiler/main.bal
+++ b/compiler/main.bal
@@ -13,6 +13,7 @@ public type Options record {
     boolean showTypes = false;
     int? debugLevel;
     // outDir also implies treating each file as a separate module
+    string? outDir = ();
     string? expectOutDir = ();
     string? gc = ();
     string? htmlError = ();
@@ -58,7 +59,7 @@ public function main(string[] filenames, *Options opts) returns error? {
     foreach string filename in filenames {
         var [basename, ext] = basenameExtension(filename);
         if ext == SOURCE_EXTENSION {
-            CompileError? err = compileBalFile(filename, basename, check chooseOutputBasename(basename, opts.outDir), nbackOptions, opts);
+            CompileError? err = printDiagnostic(compileBalFile(filename, basename, check chooseOutputBasename(basename, opts.outDir), opts.outDir, nbackOptions, opts));
             if err is err:Diagnostic {
                 errorFileCount += 1;
                 dPrinter.print(err.detail());
--- a/compiler/main.bal
+++ b/compiler/main.bal
@@ -59,7 +59,7 @@ public function main(string[] filenames, *Options opts) returns error? {
     foreach string filename in filenames {
         var [basename, ext] = basenameExtension(filename);
         if ext == SOURCE_EXTENSION {
-            CompileError? err = printDiagnostic(compileBalFile(filename, basename, check chooseOutputBasename(basename, opts.outDir), opts.outDir, nbackOptions, opts));
+            CompileError? err = compileBalFile(filename, basename, check chooseOutputBasename(basename, opts.outDir), nbackOptions, opts);
             if err is err:Diagnostic {
                 errorFileCount += 1;
                 dPrinter.print(err.detail());
--- a/compiler/main.bal
+++ b/compiler/main.bal
@@ -45,7 +45,8 @@ public function main(string[] filenames, *Options opts) returns error? {
     }
     nback:Options nbackOptions = {
         gcName: check nback:validGcName(opts.gc),
-        debugLevel: check nback:validDebugLevel(opts.debugLevel)
+        debugLevel: check nback:validDebugLevel(opts.debugLevel),
+        outDir: opts.outDir
     };
     int errorFileCount = 0;
     d:Printer dPrinter;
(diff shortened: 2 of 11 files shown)


jclark:
After doing this, can we set breakpoints using GDB based on line numbers?

compiler/modules/nback/scaffold.bal
```diff
@@ -172,6 +174,8 @@ class Scaffold {
 
     function stackGuard() returns llvm:PointerValue => self.mod.stackGuard;
 
+    function getDebugLevel() returns DebugLevel => self.debugLevel;
```
jclark: Instead of this, it's better to change the third argument to setDebugLocation to allow an additional value meaning set the debug location if the debugging level is full.

compiler/modules/nback/scaffold.bal
```diff
@@ -115,6 +115,7 @@ class Scaffold {
     private final llvm:FunctionDefn llFunc;
     private final DISubprogram? diFunc;
     private DILocation? noLineLocation = ();
+    private final DebugLevel debugLevel;
```
jclark: This field should be in ModuleDi and should be boolean debugFull. debugLevel 0 is indicated by have a null ModuleDI.

jclark (changes requested):

compiler/modules/nback/scaffold.bal
heshanpadmasiri: fixed

compiler/modules/nback/scaffold.bal
heshanpadmasiri: fixed

heshanpadmasiri:
> After doing this, can we set breakpoints using GDB based on line numbers?

yes. But we are setting the file path in debug information to be same as what users gives us. So if the user gives us a relative path and give a outdir, file path is not valid from that outdir. May I change this to use absolute path?

jclark:
> > After doing this, can we set breakpoints using GDB based on line numbers?
> 
> yes. But we are setting the file path in debug information to be same as what users gives us. So if the user gives us a relative path and give a outdir, file path is not valid from that outdir. May I change this to use absolute path?

The trouble with that is that it would include an absolute directory in the .ll files that we commit to the repo, which is not desirable.

If the user gives a relative path to the outdir, then file path should be made relative to that outdir.

compiler/modules/nback/module.bal
```diff
@@ -83,13 +83,14 @@ function createTypeUsage(table<UsedSemType> usedSemTypes) returns TypeUsage {
     return { types: types.cloneReadOnly(), uses: uses.cloneReadOnly() };
 }
 
-function createModuleDI(llvm:Module mod, bir:File[] partFiles) returns ModuleDI {
+function createModuleDI(llvm:Module mod, bir:File[] partFiles, DebugLevel debugLevel) returns ModuleDI {
```
jclark: Pass debugFull as an argument to this.

compiler/modules/nback/scaffold.bal
```diff
@@ -235,10 +236,13 @@ class Scaffold {
         return d:location(self.file, pos);
     }
 
-    function setDebugLocation(llvm:Builder builder, bir:Position pos, "file"? fileOnly = ()) {
+    function setDebugLocation(llvm:Builder builder, bir:Position pos, "file"? fileOnly = (), DebugLevel? ifLevel = ()) {
```
jclark: Combine fileOnly and ifLevel into a single argument. Please think carefully about the right name for the arguments and values.

jclark (changes requested):

heshanpadmasiri:
> > > After doing this, can we set breakpoints using GDB based on line numbers?
> > 
> > 
> > yes. But we are setting the file path in debug information to be same as what users gives us. So if the user gives us a relative path and give a outdir, file path is not valid from that outdir. May I change this to use absolute path?
> 
> The trouble with that is that it would include an absolute directory in the .ll files that we commit to the repo, which is not desirable.
> 
> If the user gives a relative path to the outdir, then file path should be made relative to that outdir.

fixed

compiler/modules/nback/module.bal
heshanpadmasiri: fixed

compiler/modules/nback/scaffold.bal
heshanpadmasiri: fixed

compiler/modules/nback/scaffold.bal
```diff
@@ -106,8 +106,13 @@ type ModuleDI record {|
     DIFile[] files;
     DICompileUnit compileUnit;
     DISubroutineType funcType;
+    boolean debugFull;
 |};
 
+type DebugLocationConfiguration record {|
+    "fileOnly"? location = ();
+    "debugFull"? ifDebugLevel = ();
+|};
```
jclark: I haven't made myself clear. It never makes sense to say both

- this is only when debugFull is true, and
- only include file information

So there are only three possibilities.  We need *one* argument that can take three possible values, not two arguments that can each take two values.

The question is what to call this one argument and its values.

jclark (changes requested):

compiler/modules/nback/scaffold.bal
jclark: The way to do this cleanly is to make the argument to setDebugLocation describe what the debug information is being provided for, which is one of

* error constructor - filename for backtrace (because position has to come from the error)
* call instruction - backtrace
* other instruction - debug

Better to use integers for these, so you can test for multiple ones using a comparison.

You need to think above what happens for the first case, when we have debugFull.

heshanpadmasiri: fixed

compiler/modules/nback/scaffold.bal
```diff
@@ -106,8 +106,16 @@ type ModuleDI record {|
     DIFile[] files;
     DICompileUnit compileUnit;
     DISubroutineType funcType;
+    boolean debugFull;
 |};
 
+// Debug location will always be added
+public const int ERROR_CONSTRUCTOR = 0; // only include file name not position
```
jclark: These constants need a prefix e.g. `DEBUG_`.

compiler/modules/nback/scaffold.bal
```diff
@@ -106,8 +106,16 @@ type ModuleDI record {|
...
+public const int CALL_INSTRUCTION = 1;
+// Debug locattion for these will be added only in full debug
+public const int OTHER_INSTRUCTION = 2;
+
+public type DebugLocationTarget ERROR_CONSTRUCTOR|CALL_INSTRUCTION|OTHER_INSTRUCTION;
```
jclark: It's not the `Target`. Maybe `Origin`.

The names of the constants and names of the type need to be harmonious and the 3 names need to be consistent. INSTRUCTION doesn't add value: everything comes from an instruction.

compiler/modules/nback/scaffold.bal
```diff
@@ -235,16 +243,15 @@ class Scaffold {
...
-    function setDebugLocation(llvm:Builder builder, bir:Position pos, "file"? fileOnly = ()) {
+    function setDebugLocation(llvm:Builder builder, bir:Position pos, DebugLocationTarget target) {
         DISubprogram? diFunc = self.diFunc;
         if diFunc != () {
             ModuleDI di = <ModuleDI>self.mod.di;
             DILocation loc;
-            if fileOnly == () {
-                var [line, column] = self.file.lineColumn(pos);
-                loc = di.builder.createDebugLocation(self.mod.llContext, line, column, self.diFunc);
+            if target > CALL_INSTRUCTION && !di.debugFull {
```
jclark: I am sure we can do better than this...

jclark (changes requested):

compiler/modules/nback/scaffold.bal
heshanpadmasiri: fixed

compiler/modules/nback/scaffold.bal
heshanpadmasiri: fixed

compiler/modules/nback/scaffold.bal
heshanpadmasiri: changed such that in full debug all locations will have the position (this have no effect on the panic error message)

compiler/modules/nback/scaffold.bal
```diff
@@ -106,8 +106,16 @@ type ModuleDI record {|
     DIFile[] files;
     DICompileUnit compileUnit;
     DISubroutineType funcType;
+    boolean debugFull;
 |};
 
+// Debug location will always be added
+public const int DEBUG_ORIGIN_CONSTRUCTOR = 0;
```
jclark: `DEBUG_ORIGIN_ERROR_CONSTRUCT`

compiler/modules/nback/scaffold.bal
```diff
@@ -235,16 +243,16 @@ class Scaffold {
         return d:location(self.file, pos);
     }
 
-    function setDebugLocation(llvm:Builder builder, bir:Position pos, "file"? fileOnly = ()) {
+    function setDebugLocation(llvm:Builder builder, bir:Position pos, DebugLocationOrigin origin) {
```
jclark: We can default origin to other.

compiler/modules/nback/scaffold.bal
```diff
@@ -235,16 +243,16 @@ class Scaffold {
...
         DISubprogram? diFunc = self.diFunc;
         if diFunc != () {
```
jclark: Take advantage of narrowing after return. Use an `if` statement to identify all the cases where we do not need to set a debug location, and return immediately if we do not. Then use if to test whether it's just a file.

compiler/main.bal
```diff
@@ -44,7 +44,8 @@ public function main(string[] filenames, *Options opts) returns error? {
     }
     nback:Options nbackOptions = {
         gcName: check nback:validGcName(opts.gc),
-        debugLevel: check nback:validDebugLevel(opts.debugLevel)
+        debugLevel: check nback:validDebugLevel(opts.debugLevel),
+        outDir: opts.outDir
```
jclark: Why is this the right place to do this? Surely the locations in the BIR should have the right directory.

jclark (changes requested):

compiler/main.bal
heshanpadmasiri: locations in BIR have the path to file relative to where we run the compiler from but debug information needs it relative to outDir

compiler/modules/nback/scaffold.bal
heshanpadmasiri: fixed

compiler/modules/nback/scaffold.bal
heshanpadmasiri: fixed

compiler/modules/nback/scaffold.bal
heshanpadmasiri: fixed

compiler/main.bal
jclark: How is it useful for BIR to have locations relative to where we run the compiler?  The point of BIR is to create output. So the BIR locations should be relative to the output directory. Please propose appropriate design changes in the issue.

compiler/modules/nback/scaffold.bal
```diff
@@ -235,16 +243,12 @@ class Scaffold {
         return d:location(self.file, pos);
     }
 
-    function setDebugLocation(llvm:Builder builder, bir:Position pos, "file"? fileOnly = ()) {
+    function setDebugLocation(llvm:Builder builder, bir:Position pos, DebugLocationOrigin origin = DEBUG_ORIGIN_OTHER) {
         DISubprogram? diFunc = self.diFunc;
-        if diFunc != () {
-            ModuleDI di = <ModuleDI>self.mod.di;
+        ModuleDI? di = self.mod.di;
+        if diFunc != () && di != () && (origin < DEBUG_ORIGIN_OTHER || di.debugFull) {
```
jclark: This is still not nice.  Let's LiveCode it together.

compiler/main.bal
jclark: Why is outDir still here?

jclark (changes requested):

heshanpadmasiri:
Fixed `outDir`

compiler/output.bal
```diff
@@ -8,6 +8,7 @@ const OUTPUT_EXTENSION = ".ll";
 
 type OutputOptions record {
     string? target = ();
+    string? outDir = ();
```
jclark: This does not belong here since it is not used by outputModule.

jclark (changes requested):

compiler/output.bal
heshanpadmasiri: fixed

compiler/compile.bal
```diff
@@ -71,9 +71,9 @@ class CompileContext {
 }
 
 // basename is filename without extension
-function compileBalFile(string filename, string basename, string? outputBasename, nback:Options nbackOptions, OutputOptions outOptions) returns CompileError? {
+function compileBalFile(string filename, string basename, string? outputBasename, string? outDir, nback:Options nbackOptions, OutputOptions outOptions) returns CompileError? {
```
jclark: There's a basic problem here that we haven't yet considered. We need to make this work for files in imported modules. This is handled in processModule/subModuleSourceParts. But for imported modules, it looks like `directory` is being used for the module directory.  We need a design for this.

jclark (approved):
Let's discuss this in the issue.

jclark:
Let's do this. In this issue, just pass through the debug info to LLVM.

Then let's deal with the output directory problem in a separate issue #761. It's really a separate issue: we already have it with backtraces.

heshanpadmasiri:
removed passing output directory

compiler/modules/comm.diagnostic/diagnostic.bal
```diff
@@ -23,7 +23,12 @@ public type Range readonly & record {|
 public type LineColumn readonly & [int, int];
 
 public type File readonly & object {
+    // If the filename is relative, it is relative to the compilation directory
```
jclark: These two comments are not correct for imports. Remove the changes to this file.

compiler/modules/nback/scaffold.bal
```diff
@@ -106,8 +106,16 @@ type ModuleDI record {|
     DIFile[] files;
     DICompileUnit compileUnit;
     DISubroutineType funcType;
+    boolean debugFull;
 |};
 
+// Debug location will always be added
+public const int DEBUG_ORIGIN_ERROR_CONSTRUCT = 0;
+public const int DEBUG_ORIGIN_CALL = 1;
+// Debug locattion for these will be added only in full debug
```
jclark: Spelling

jclark (changes requested):

compiler/modules/comm.diagnostic/diagnostic.bal

compiler/modules/nback/scaffold.bal
heshanpadmasiri: fixed

PR #749, opened 2021-12-05
Support inclusive-record-type-descriptor syntax

### LINKED ISSUE #737: Support inclusive-record-type-descriptor syntax
Support the `record { }` syntax. This should be explicit in the AST and then resolved into something equivalent to `record {| anydata...; |}`.

Need to make sure preparse understands `{` and `}` bracketing.

This is part of #573.

heshanpadmasiri:
## Purpose
Support inclusive-record-type-descriptor syntax
Resolves #737

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/front.syntax/ast.bal
```diff
@@ -472,7 +472,17 @@ public type FieldDesc record {|
...
-public type MappingTypeDesc record {|
+public type MappingTypeDesc MapTypeDesc|RecordTypeDesc;
+
+public type MapTypeDesc record {|
+    *PositionFields;
+    TypeDesc typeParam;
+    t:MappingDefinition? defn = ();
+|};
+
+public type RecordTypeDesc ExclusiveRecordTypeDesc|InclusiveRecordTypeDesc;
```
jclark: I think it's simpler to handle this by changing `rest` have type TypeDesc|boolean, where true is used for an inclusive descriptor.

jclark (changes requested):

compiler/modules/front.syntax/ast.bal

compiler/modules/front.syntax/ast.bal
```diff
@@ -477,7 +477,7 @@ public type MappingTypeDesc record {|
     Position startPos;
     Position endPos;
     FieldDesc[] fields;
-    TypeDesc? rest;
+    TypeDesc|boolean rest; // true if inclusive record type false if rest is nil
```
jclark: The comment is inaccurate.
`true` means the record rest comes implicitly from the use of an inclusive-record-type-descriptor. `false` means there is no record-rest.

Actually I think on reflection it is better for this to be `TypeDesc|INCLUSIVE_RECORD_TYPE_DESC|()`, with

```
public const INCLUSIVE_RECORD_TYPE_DESC = true;
```

compiler/modules/front/resolveTypes.bal
```diff
@@ -172,13 +172,16 @@ function resolveTypeDesc(ModuleSymbols mod, s:ModuleLevelDefn modDefn, int depth
...
-            s:TypeDesc? restTd = td.rest;
+            s:TypeDesc|boolean restTd = td.rest;
             t:SemType rest;
-            if restTd == () {
-                rest = t:NEVER;
+            if restTd is s:TypeDesc {
+                rest = check resolveTypeDesc(mod, modDefn, depth + 1, restTd);
+            }
+            else if restTd == true {
+                rest = t:ANY;
```
jclark: No -- read the spec.

compiler/modules/front.syntax/ast.bal

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: fixed

compiler/modules/nback/scaffold.bal
```diff
@@ -470,7 +470,7 @@ function semTypeRepr(t:SemType ty) returns Repr {
     if w == t:NEVER {
         panic err:impossible("allocate register with never type");
     }
-    int supported = t:NIL|t:BOOLEAN|t:INT|t:FLOAT|t:STRING|t:LIST|t:MAPPING|t:ERROR;
+    int supported = t:NIL|t:BOOLEAN|t:INT|t:FLOAT|t:DECIMAL|t:STRING|t:LIST|t:MAPPING|t:XML|t:TABLE|t:ERROR;
```
heshanpadmasiri: @jclark I am not sure whether this is the correct way to support any data (In the original form it will cause the panic "unimplemented type")

heshanpadmasiri: @jclark adding table here (`if (w | supported | t:DECIMAL| t:TABLE) == t:TOP || (w & supported) == w {`) had no effect because w (`1111000000011110111111`) already has all the bits of t:Table (`100000000000000010000`)

jclark: So why is it getting an unimplemented type error then? (t:DECIMAL and t:TABLE need **not** to be in supported).

What you have here is not right.

compiler/modules/front.syntax/astString.bal
```diff
@@ -244,30 +244,41 @@ function typeDescToWords(Word[] w, TypeDesc td, boolean|BinaryTypeOp wrap = fals
         return;
     }
     else if td is MappingTypeDesc {
-        TypeDesc? rest = td.rest;
+        TypeDesc|INCLUSIVE_RECORD_TYPE_DESC|() rest = td.rest;
         if td.fields.length() > 0 {
-            w.push("record", "{|");
+            w.push("record");
+            if rest != true {
```
jclark: Swap it round and say `w.push(rest == INCLUSIVE_RECORD_TYPE_DESC ? "{" : "{|");`

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -388,3 +402,18 @@ function parseRecordTypeDesc(Tokenizer tok, Position startPos) returns MappingTy
     check tok.advance();
     return { startPos, endPos, fields, rest };
 }
+
+function parseInclusiveRecordTypeDesc(Tokenizer tok, Position startPos) returns MappingTypeDesc|err:Syntax {
+    check tok.advance();
+    FieldDesc[] fields = [];
+    while tok.current() != "}" {
+        TypeDesc td = check parseTypeDesc(tok);
+        Position fieldStartPos = tok.currentStartPos();
```
jclark: Make this share code with parseExclusiveRecordTypeDesc.

jclark (changes requested):

compiler/modules/nback/scaffold.bal
heshanpadmasiri: If I understand this correctly left side (currently (` w | supported | t:DECIMAL) == t:TOP`)) works by adding every thing that is not w or support (because t:TOP is `11111111111111111111111`)). So in the current case this means if the only missing part in the union of w and support is decimal this is true. If this is the case fallowing are the offsets needed here for any data
```
 UT_OBJECT_RO  = 0x06;
 UT_FUNCTION   = 0x0C;
 UT_TYPEDESC   = 0x0D;
 UT_HANDLE     = 0x0E;
 0x0F; // I couldn't find any thing corresponding to this offset
 UT_FUTURE     = 0x10;
 UT_STREAM     = 0x11;
 UT_OBJECT_RW  = 0x16;
```

jclark: The first `==` is handling the case where the user specified `any`. The `|` supported is dealing with narrowing from the supported types. The `|t:DECIMAL` is dealing with the fact the `<int>any` will have type `any - (float|decimal)`.

jclark: I just added t:NON_BEHAVIORAL for us to use here. I think this should do it:

```
int maximized = w | supported | t:DECIMAL;
if maximized == t:TOP || maximized == t:NON_BEHAVIOURAL|t:ERROR ...
```

jclark: If it gets much more complicated, we should just get rid of this checking and return the TaggedRepr. (We should keep the test for NEVER at the beginning).

heshanpadmasiri: fixed

compiler/modules/front.syntax/astString.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/parseTypeDesc.bal
heshanpadmasiri: factored out `parseFieldDesc`

compiler/modules/nback/scaffold.bal
```diff
@@ -471,8 +471,10 @@ function semTypeRepr(t:SemType ty) returns Repr {
         panic err:impossible("allocate register with never type");
     }
     int supported = t:NIL|t:BOOLEAN|t:INT|t:FLOAT|t:STRING|t:LIST|t:MAPPING|t:ERROR;
-    // DECIMAL is here for ConvertToInt or ConvertToFloat 
-    if (w | supported | t:DECIMAL) == t:TOP || (w & supported) == w {
+    // DECIMAL is here for ConvertToInt or ConvertToFloat
+    // TABLE and XML are here for anydata
+    int maximized = w | supported | t:DECIMAL | t:TABLE | t:XML;
+    if maximized == t:TOP || maximized == (t:NON_BEHAVIORAL|t:ERROR) {
```
jclark: My suggestion was:

```
int maximized = w | supported | t:DECIMAL;
if maximized == t:TOP || maximized == t:NON_BEHAVIOURAL|t:ERROR ...
```

which is not the same.  You need to call out explicitly if something I suggest doesn't work.

The `...` meant this:

```
int maximized = w | supported | t:DECIMAL;
if maximized == t:TOP || maximized == t:NON_BEHAVIOURAL|t:ERROR || (w & supported) == w { ... }
```

jclark (changes requested):

compiler/modules/nback/scaffold.bal
heshanpadmasiri: fixed

PR #750, opened 2021-12-05
Implement chained field access lvalues

### LINKED ISSUE #738: Support chained field access lvalues
With #571, we should generalize #510 to support chained field access for lvalues (e.g. `x.y.z = v;`).

This is just for field access using `v.k`, which can be done just in the front-end, not for member access using `v[k]`, which requires support in the backend for the filling-read operation.

Follow on to this is #684.

manuranga:
Fixes #738

### DIFF
--- a/compiler/modules/front.syntax/ast.bal
+++ b/compiler/modules/front.syntax/ast.bal
@@ -334,7 +334,7 @@ public type MemberAccessExpr record {|
 public type MemberAccessLExpr record {|
     *PositionFields;
     Position opPos;
-    VarRefExpr container;
+    VarRefExpr|FieldAccessLExpr container;
     Expr index;
 |};
 
@@ -351,7 +351,7 @@ public type FieldAccessExpr record {|
 public type FieldAccessLExpr record {|
     *PositionFields;
     Position opPos;
-    VarRefExpr container;
+    VarRefExpr|FieldAccessLExpr container;
     string fieldName;
 |};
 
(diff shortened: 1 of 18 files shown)


compiler/modules/front.syntax/ast.bal
```diff
@@ -351,7 +351,7 @@ public type FieldAccessExpr record {|
 public type FieldAccessLExpr record {|
     *PositionFields;
     Position opPos;
-    VarRefExpr container;
+    VarRefExpr|FieldAccessLExpr container;
```
jclark, line 354: This does not restrict things in the right way. The thing we cannot suppport is member access as an *intermediate* step, because that is what uses a filling read. So `x.y.z[0]` is fine but `x[0].y.z` isn't.

jclark (changes requested):

docs/subset11.md
```diff
@@ -173,9 +173,9 @@ lvexpr =
    | field-access-lvexpr
    | member-access-lvexpr 
 
-member-access-lvexpr = variable-reference-lvexpr "[" expression "]"
+member-access-lvexpr = ( variable-reference-lvexpr | field-access-lvexpr ) "[" expression "]"
```
jclark:
```
lvexpr =
   chainable-lvexpr
   | member-access-lvexpr

chainable-lvexpr = variable-reference-lvexpr | field-access-lvexpr
```

PR #753, opened 2021-12-07
Fix endPos confusion

### LINKED ISSUE #751: endPos confusion
currentEndPos is doing

```
   function currentEndPos() returns Position {
        return createPosition(self.lineIndex, self.codePointIndex-1);
    }
```

The `-1` is not right: it makes parse errors have one two few carets.

endPos - startPos should be equal to the length of the token.

Positions should be thought of as **between** characters: the startPos is the position preceding the first character of the token; the endPos is the position following the last character of the token.

heshanpadmasiri:
## Purpose
Fix endPos confusion
Resolves #751

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/front.syntax/parseExpr.bal
+++ b/compiler/modules/front.syntax/parseExpr.bal
@@ -145,7 +145,8 @@ function finishTypeTestExpr(Tokenizer tok, Expr expr, boolean negated, Position 
 function parseRangeExpr(Tokenizer tok) returns RangeExpr|err:Syntax {
     Position startPos = tok.currentStartPos();
     Expr lower = check parseAdditiveExpr(tok);
-    Position opPos = check tok.expectEnd("..<");
+    Position opPos = tok.currentStartPos();
+    check tok.expect("..<");
     Expr upper = check parseAdditiveExpr(tok);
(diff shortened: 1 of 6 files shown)


compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -145,7 +145,8 @@ function finishTypeTestExpr(Tokenizer tok, Expr expr, boolean negated, Position
 function parseRangeExpr(Tokenizer tok) returns RangeExpr|err:Syntax {
     Position startPos = tok.currentStartPos();
     Expr lower = check parseAdditiveExpr(tok);
-    Position opPos = check tok.expectEnd("..<");
+    Position opPos = tok.currentStartPos();
```
jclark: I suggest you add an `expectStart` method to deal with these.

jclark (changes requested):

compiler/modules/front.syntax/parseExpr.bal
heshanpadmasiri: fixed

PR #755, opened 2021-12-07
Compiler support for decimal literals

KavinduZoysa:
## Purpose
Related to #289


compiler/modules/nback/scaffold.bal
```diff
@@ -339,15 +350,15 @@ function addRuntimeFunctionDecl(llvm:Module mod, RuntimeFunction rf) returns llv
     return f;
 }
 
-function addStringDefn(llvm:Context context, llvm:Module mod, int defnIndex, string str) returns llvm:ConstPointerValue {
+function addStringDefn(llvm:Context context, llvm:Module mod, int defnIndex, string str, boolean isDecStr) returns llvm:ConstPointerValue {
```
jclark: I don't want to make this handle decimal constants also. Deal with decimal constants separately.

jclark (changes requested):

compiler/modules/nback/scaffold.bal
KavinduZoysa: Fixed

compiler/modules/nback/scaffold.bal
```diff
@@ -208,6 +209,16 @@ class Scaffold {
         return newDefn;
     }
 
+    function getDecimalString(string str) returns StringDefn {
+        StringDefn? curDefn = self.mod.stringDefns[str];
```
jclark: Don't use stringDefns. The decimal consts should be nul-terminated without any length.

jclark (changes requested):

compiler/modules/nback/scaffold.bal
KavinduZoysa: @jclark, I am a bit unclear about this comment. Did you mean we need to model const decimal string as shown below?

`@.str1 = internal unnamed_addr constant [6 x i8] c"2E+10\00", align 8`

jclark: It's got nothing to do with `.str`. Don't reuse any code of the related to dealing with Ballerina strings. A decimal constant should produce a C style nul-terminated string, so decimal 42 would produce a string with 3 bytes ASCII '4', ASCII '2' and 0.

compiler/modules/front/const.bal
```diff
@@ -517,22 +520,46 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
     }
 }
 
-function foldFloatLiteralExpr(FoldContext cx, t:SemType? expectedType, s:FpLiteralExpr expr) returns s:ConstValueExpr|FoldError {
-    // This will need to change when we support decimal
-    float|error result = floatFromDecimalLiteral(expr.untypedLiteral);
-    if result is float {
+function foldFpLiteralExpr(FoldContext cx, t:SemType? expectedType, s:FpLiteralExpr expr) returns s:ConstValueExpr|FoldError {
+    float|decimal|error result;
```
manuranga: Please test error cases such as `float f = 2.9d;`.

compiler/modules/front/const.bal
```diff
@@ -517,22 +520,46 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
+            result = floatFromFpLiteral(untypedLiteral);
+            ty = "float";            
+        }
+    }
+    if result is float|decimal {
         return { startPos: expr.startPos, endPos: expr.endPos, value: result };
     }
     else {
-        return cx.semanticErr("invalid float literal", cause=result, pos=expr.startPos);
+        return cx.semanticErr("invalid " + ty + " literal", cause=result, pos=expr.startPos);
```
manuranga: Do we ever each this line, if so for what input? if not we should simplify the code by removing error message.

compiler/modules/front/const.bal
```diff
@@ -556,10 +583,14 @@ function expectsFloat(t:SemType? semType) returns boolean {
     }
 }
 
-function floatFromDecimalLiteral(string digits) returns float|error {
+function floatFromFpLiteral(string digits) returns float|error {
```
manuranga: This renaming is wrong, original name used `Decimal` to mean non-hex, not to mean floating point.

compiler/modules/front/const.bal
```diff
@@ -517,22 +520,46 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
 }
 
 function foldIntLiteralExpr(FoldContext cx, t:SemType? expectedType, s:IntLiteralExpr expr) returns s:ConstValueExpr|FoldError {
     float|int|error result;
     string ty;
     if expr.base == 10 && expectsFloat(expectedType) {
-        result = floatFromDecimalLiteral(expr.digits);
+        result = floatFromFpLiteral(expr.digits);
         ty = "float"; 
     }
```
manuranga, line 573: We'll need `expectsDecimal`

compiler/modules/front/const.bal
manuranga: Now we have `decimal`type, name is confusing, we should rename, or get rid of this. But we have used the term `Decimal` in `scan.bal` to mean 10-ary as well. @jclark what do you think about replacing `Decimal` with `Denary` in the code base when it means 10-ary.

runtime/print.c
```diff
@@ -142,6 +143,13 @@ static void printTagged(FILE *fp, TaggedPtr p, int style, struct PrintStack *sta
                 }
             }
             break;
+        case TAG_DECIMAL:
+            {
+                char decStr[DECQUAD_String];
+                decQuadToString((void *)taggedToPtr(p), decStr);
```
manuranga, line 149: Shouldn't this be a case to `decQuad *` ?

compiler/modules/front.syntax/tests/testParser.bal
```diff
@@ -55,7 +55,7 @@ function testTokenizer(string k, string[] lines) returns error? {
             int tStart = lc[1];
             string tStr = tokenToString(t);
             string srcAtPos = src.substring(tStart, tStart + tStr.length());
-            if t is [HEX_INT_LITERAL, string] || t is [DECIMAL_FP_NUMBER, string, "f"?] {
+            if t is [HEX_INT_LITERAL, string] || t is [DECIMAL_FP_NUMBER, string, "f"|"d"?] {
                 // need to normalize `0x` vs `0X` and `f`
```
manuranga:
```suggestion
                // Canonicalize `0X` `F` `D` in source as `0x` `f` `d`.
```

compiler/modules/front/const.bal
```diff
@@ -556,10 +583,14 @@ function expectsFloat(t:SemType? semType) returns boolean {
...
 }
 
-function floatFromDecimalLiteral(string digits) returns float|error {
+function floatFromFpLiteral(string digits) returns float|error {
     return float:fromString(digits);
```
manuranga: Please verify that error case will never occur and then change the code as follows
```ballerina
function floatFromDecimalLiteral(string digits) returns float {
  float|error f = float:fromString(digits);
  if f is error {
      panic err:impossible();
  }
  return f;
}
```
And lets add a comment before method saying the prerequisites.

compiler/modules/nback/scaffold.bal
KavinduZoysa: Changed

compiler/modules/front/const.bal
KavinduZoysa: Removed

compiler/modules/front/const.bal
KavinduZoysa: Added

runtime/print.c
KavinduZoysa: Changed

compiler/modules/front.syntax/tests/testParser.bal
KavinduZoysa: Changed

compiler/modules/front/const.bal
KavinduZoysa: Changed

compiler/modules/front/const.bal
```diff
@@ -517,47 +520,88 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
 }
 
-function foldFloatLiteralExpr(FoldContext cx, t:SemType? expectedType, s:FpLiteralExpr expr) returns s:ConstValueExpr|FoldError {
-    // This will need to change when we support decimal
-    float|error result = floatFromDecimalLiteral(expr.untypedLiteral);
-    if result is float {
-        return { startPos: expr.startPos, endPos: expr.endPos, value: result };
+function foldFpLiteralExpr(FoldContext cx, t:SemType? expectedType, s:FpLiteralExpr expr) returns s:ConstValueExpr|FoldError { 
+    float|decimal result;
+    t:Context tx = cx.typeContext();
```
manuranga:
```suggestion
    t:Context tc = cx.typeContext();
```

compiler/modules/front/const.bal
```diff
@@ -517,47 +520,88 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
-        return t:isSubtypeSimple(t:intersect(semType, t:union(t:FLOAT, t:INT)), t:FLOAT);
+    return f;
+}
+
+// Even if the decimal floating point literal is parsed correctly,
+// infinities should through an error.
+function decimalFromFpLiteral(FoldContext cx, string decimalStr, Position pos) returns decimal|FoldError {
+    decimal|error d = decimal:fromString(decimalStr);
+    if d is error {
+        return cx.semanticErr("invalid decimal literal", pos);
```
manuranga: For what input will this happen?

compiler/modules/front/const.bal
```diff
@@ -517,47 +520,88 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
 }
 
-function floatFromDecimalLiteral(string digits) returns float|error {
-    return float:fromString(digits);
+// Even if the integer literal is parsed correctly,
+// infinities should through an error.
+function intFromLiteral(FoldContext cx, s:IntLiteralExpr expr) returns int|FoldError {
+    int|error i = s:intFromIntLiteral(expr.base, expr.digits);
+    if i is error {
+        return cx.semanticErr("invalid int literal", expr.startPos);
```
manuranga: For what input will this happen?

compiler/modules/front/const.bal
```diff
@@ -517,47 +520,88 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
+    t:Context tx = cx.typeContext();
+    s:FpTypeSuffix? typeSuffix = expr.typeSuffix;
+    string untypedLiteral = expr.untypedLiteral;
+    Position startPos = expr.startPos;
+    if typeSuffix != () {
+        result = typeSuffix is s:FLOAT_TYPE_SUFFIX ? floatFromFpLiteral(untypedLiteral) : check decimalFromFpLiteral(cx, untypedLiteral, startPos);
     }
     else {
-        return cx.semanticErr("invalid float literal", cause=result, pos=expr.startPos);
+        if expectedType == () || expectsFloat(tx, expectedType) {
```
manuranga: else if

compiler/modules/front/const.bal
```diff
@@ -517,47 +520,88 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
     }
+    return { startPos: startPos, endPos: expr.endPos, value: result };
+}
+
+function expectsInt(t:Context cx, t:SemType semType) returns boolean {
+    return !t:isEmpty(cx, t:intersect(semType, t:INT));
+}
+
+function expectsFloat(t:Context cx, t:SemType semType) returns boolean {
+    return !t:isEmpty(cx, t:intersect(semType, t:FLOAT));
```
manuranga: original code was `t:isSubtypeSimple(t:intersect(semType, t:union(t:FLOAT, t:INT)), t:FLOAT)` why did you change it?

compiler/modules/front/const.bal
KavinduZoysa: Changed

compiler/modules/front/const.bal
KavinduZoysa: For values greater than decimal max or less than min(ex: 1e6145, -1e6145), these values are parsed correctly, but fromString should give an error(https://github.com/ballerina-platform/ballerina-lang/issues/33805#issuecomment-988473730).

compiler/modules/front/const.bal
KavinduZoysa: For values greater than int max or less than min (ex: 9223372036854775808).

compiler/modules/front/const.bal
KavinduZoysa: Changed

compiler/modules/front/const.bal
KavinduZoysa: Yes, it was implemented initially based on the 4th bullet point in https://ballerina.io/spec/lang/2021R1/#section_6.5. Now it should be changed according to https://github.com/ballerina-platform/ballerina-spec/issues/943#issuecomment-939932254.
I have tried to refactor both methods `foldFpLiteralExpr` and `foldIntLiteralExpr` according to this.

compiler/modules/front/const.bal
manuranga: Does this mean large float values don't give an error and large decimals do? that is odd.

manuranga: Ahh this is because we don't have decimals infinity.

compiler/modules/front/const.bal
```diff
@@ -517,47 +520,94 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
+    Position startPos = expr.startPos;
+    if typeSuffix != () {
+        result = typeSuffix is s:FLOAT_TYPE_SUFFIX ? floatFromFpLiteral(untypedLiteral) : check decimalFromFpLiteral(cx, untypedLiteral, startPos);
     }
     else {
-        return cx.semanticErr("invalid float literal", cause=result, pos=expr.startPos);
+        if expectedType == () {
+            result = floatFromFpLiteral(untypedLiteral);
+        }
+        else if expectsFloat(tc, expectedType) {
```
manuranga: merge all three

compiler/modules/front/const.bal
```diff
@@ -517,47 +520,94 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
+    if f is error {
+        panic err:impossible();
     }
-    else {
-        return t:isSubtypeSimple(t:intersect(semType, t:union(t:FLOAT, t:INT)), t:FLOAT);
+    return f;
+}
+
+// Even if the decimal floating point literal is parsed correctly,
+// infinities should through an error.
```
manuranga: overflow should throw an error.

compiler/modules/front/const.bal
KavinduZoysa: Changed

compiler/modules/front/const.bal
KavinduZoysa: Changed

compiler/modules/nback/scaffold.bal
```diff
@@ -208,6 +211,16 @@ class Scaffold {
         return newDefn;
     }
 
+    function getDecimalString(string str) returns DecimalStringDefn {
```
manuranga: Argument should be a `decimal`.

compiler/modules/front/const.bal
```diff
@@ -517,47 +520,86 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
+    if f is error {
+        panic err:impossible();
     }
-    else {
-        return t:isSubtypeSimple(t:intersect(semType, t:union(t:FLOAT, t:INT)), t:FLOAT);
+    return f;
+}
+
+// Even if the decimal floating point literal is parsed correctly,
+// overflow should through an error.
```
manuranga: Spelling mistake in throw.

compiler/modules/nback/scaffold.bal
```diff
@@ -208,6 +211,17 @@ class Scaffold {
         return newDefn;
     }
 
+    function getDecimalString(decimal decimalValue) returns DecimalStringDefn {
```
manuranga: minor: prefer `val` here.

compiler/modules/front/const.bal
```diff
@@ -517,47 +520,86 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
-    float|error result = floatFromDecimalLiteral(expr.untypedLiteral);
-    if result is float {
-        return { startPos: expr.startPos, endPos: expr.endPos, value: result };
+function foldFpLiteralExpr(FoldContext cx, t:SemType? expectedType, s:FpLiteralExpr expr) returns s:ConstValueExpr|FoldError { 
+    float|decimal result;
+    t:Context tc = cx.typeContext();
+    s:FpTypeSuffix? typeSuffix = expr.typeSuffix;
+    string untypedLiteral = expr.untypedLiteral;
+    Position startPos = expr.startPos;
+    if typeSuffix != () {
```
manuranga: - let's use destructure
- move `result` last since it's a no-init decl
```suggestion
    var { typeSuffix, untypedLiteral, startPos } = expr.typeSuffix;
    t:Context tc = cx.typeContext();
    float|decimal result;
    if typeSuffix != () {
```

compiler/modules/front/const.bal
```diff
@@ -517,47 +520,84 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
+    }
+    else if expectedType == () || expectsFloat(tc, expectedType) {
+        result = floatFromFpLiteral(untypedLiteral);
+    }
+    else if expectsDecimal(tc, expectedType) {
+        result = check decimalFromFpLiteral(cx, untypedLiteral, startPos);
     }
     else {
-        return cx.semanticErr("invalid float literal", cause=result, pos=expr.startPos);
+        result = floatFromFpLiteral(untypedLiteral);
```
manuranga: Do we ever reach here for semantically valid input? if not let's add a comment saying semantically invalid.
(Another option is to throw an error here, but we'll have to look into if that's Ok)

compiler/modules/nback/scaffold.bal
KavinduZoysa: Changed

compiler/modules/front/const.bal
KavinduZoysa: Changed

compiler/modules/front/const.bal
KavinduZoysa: The cases like this `int i = <int>1.2;` reaches this point.
The value `1.2` is a fpLiteral and the type of this value can be either float or decimal with preference float, decimal. 
The expected type is int.
Now we need to check intesection between float and int and then decimal and int. Since both cases fail we need to consider the most prefered basic type.

manuranga: Makes sense.

runtime/print.c
```diff
@@ -142,6 +143,13 @@ static void printTagged(FILE *fp, TaggedPtr p, int style, struct PrintStack *sta
                 }
             }
             break;
+        case TAG_DECIMAL:
+            {
+                char decStr[DECQUAD_String];
+                decQuadToString((decQuad *)taggedToPtr(p), decStr);
```
jclark: Better to use `(void *)` here, since the thing that needs the cast is the change of address space.

jclark:
Need some error tests e.g.

```
int x = 1.2d;
```

Also need a test for a decimal that overflows or underflows.

compiler/modules/nback/symbol.bal
```diff
@@ -65,6 +65,10 @@ function stringDefnSymbol(int n) returns string {
     return ".str" + n.toString();
 }
 
+function decimalStringDefnSymbol(int n) returns string {
+    return ".decstr" + n.toString();
```
jclark: Use just ".dec".

compiler/modules/nback/symbol.bal
```diff
@@ -65,6 +65,10 @@ function stringDefnSymbol(int n) returns string {
     return ".str" + n.toString();
 }
 
+function decimalStringDefnSymbol(int n) returns string {
```
jclark: Use just `decimalDefnSymbol`

compiler/modules/nback/scaffold.bal
```diff
@@ -74,6 +75,7 @@ const STRING_VARIANT_LARGE = 1;
 type StringVariant STRING_VARIANT_MEDIUM|STRING_VARIANT_LARGE; // STRING_VARIANT_SMALL|;
 
 type StringDefn llvm:ConstPointerValue;
+type DecimalStringDefn llvm:ConstPointerValue;
```
jclark: DecimalDefn

compiler/modules/nback/scaffold.bal
```diff
@@ -84,6 +86,7 @@ type Module record {|
     ImportedFunctionTable importedFunctions = table [];
     llvm:PointerValue stackGuard;
     map<StringDefn> stringDefns = {};
+    map<DecimalStringDefn> decimalStringDefns = {};
```
jclark: decimalDefns

compiler/modules/nback/scaffold.bal
```diff
@@ -208,6 +211,17 @@ class Scaffold {
         return newDefn;
     }
 
+    function getDecimalString(decimal val) returns DecimalStringDefn {
```
jclark: getDecimal

compiler/modules/nback/scaffold.bal
```diff
@@ -21,6 +21,7 @@ const LLVM_VOID = "void";
 final llvm:PointerType LLVM_TAGGED_PTR = heapPointerType("i8");
 final llvm:PointerType LLVM_NIL_TYPE = LLVM_TAGGED_PTR;
 final llvm:PointerType LLVM_TAGGED_PTR_WITHOUT_ADDR_SPACE = llvm:pointerType("i8");
+final llvm:PointerType LLVM_PTR_WITHOUT_ADDR_SPACE = llvm:pointerType("i8");
```
jclark: LLVM_DECIMAL_CONST

compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -275,7 +275,7 @@ function startPrimaryExpr(Tokenizer tok) returns Expr|err:Syntax {
         check tok.advance();
         return expr;
     }
-    else if t is [DECIMAL_FP_NUMBER, string, FLOAT_TYPE_SUFFIX|()] {
+    else if t is [DECIMAL_FP_NUMBER, string, FLOAT_TYPE_SUFFIX|DECIMAL_TYPE_SUFFIX|()] {
```
jclark: Use `FpTypeSuffix?`

compiler/modules/types/core.bal
```diff
@@ -952,7 +952,7 @@ public function split(SemType t) returns SplitSemType  {
 }
 
 public type Value readonly & record {|
-    string|int|float|boolean|() value;
+    string|int|float|decimal|boolean|() value;
```
jclark: We now use this union in lots of places, both in this file and elsewhere. We should have a definition for it. I think t:SingleValue is best, because it represents a type that is a SingleValue.

```
public type SingleValue ()|boolean|int|decimal|float|string;
```


But then `Value` isn't good. Solve that like this (note `?` at the end):

```
public type OptSingleValue readonly & record {|
   SingleValue value;
|}?;
```

Change ResolvedConst to use SingleValue.

compiler/modules/bir/bir.bal
```diff
@@ -210,11 +210,12 @@ public type Insn
     |BranchInsn|CondBranchInsn|CatchInsn|PanicInsn|ErrorConstructInsn;
 
 public type Operand ConstOperand|Register;
-public type SimpleConstOperand ()|boolean|int|float;
+public type SimpleConstOperand ()|boolean|int|float|decimal;
 public type ConstOperand SimpleConstOperand|string;
```
jclark: Want to use new t:SingleValue define here.

Can we get rid of SingleConstOperand?

compiler/modules/front.syntax/ast.bal
```diff
@@ -395,7 +395,7 @@ public type ConstShapeExpr ConstValueExpr|FloatZeroExpr;
 
 public type ConstValueExpr record {|
     *PositionFields;
-    ()|boolean|int|float|string value;
+    ()|boolean|int|float|decimal|string value;
```
jclark: t:SingleValue

compiler/modules/front/const.bal
```diff
@@ -14,7 +14,7 @@ import wso2/nballerina.comm.diagnostic as d;
 import wso2/nballerina.front.syntax as s;
 import wso2/nballerina.types as t;
 
-type SimpleConst string|int|float|boolean|();
+type SimpleConst string|int|float|decimal|boolean|();
```
jclark: t:SingleValue

compiler/modules/front.syntax/tests/testParser.bal
```diff
@@ -55,8 +55,8 @@ function testTokenizer(string k, string[] lines) returns error? {
             int tStart = lc[1];
             string tStr = tokenToString(t);
             string srcAtPos = src.substring(tStart, tStart + tStr.length());
-            if t is [HEX_INT_LITERAL, string] || t is [DECIMAL_FP_NUMBER, string, "f"?] {
-                // need to normalize `0x` vs `0X` and `f`
+            if t is [HEX_INT_LITERAL, string] || t is [DECIMAL_FP_NUMBER, string, "f"|"d"?] {
```
jclark: `FpTypeSuffix?`

compiler/modules/front/const.bal
jclark: t:SingleValue

compiler/modules/front/const.bal
```diff
@@ -517,47 +520,84 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
     }
     else {
-        return cx.semanticErr("invalid " + ty + " literal", cause=result, pos=expr.startPos);
+        result = check intFromLiteral(cx, expr);
     }
+    return { startPos: startPos, endPos: expr.endPos, value: result };
+}
+
+function expectsInt(t:Context cx, t:SemType semType) returns boolean {
+    return !t:isEmpty(cx, t:intersect(semType, t:INT));
```
jclark: Better to use `t:isNever` here.  Also we have a lot of cases where we are doing `t:isNever(t:intersect(semType, bitSet))`. So we should have a function in types that does it directly (by looking at some|all in ComplexSemType case) and get rid of these three functions. What to call it? I suggest `t:includesSome`. So instead of `expectsFloat(tc, expectedType)` we have `t:includesSome(expectedType, t:FLOAT)`

compiler/modules/front/const.bal
```diff
@@ -517,47 +520,84 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
+}
+
+function expectsDecimal(t:Context cx, t:SemType semType) returns boolean {
+    return !t:isEmpty(cx, t:intersect(semType, t:DECIMAL));
+}
+
+// Since the binary floating point literal is parsed correctly,
+// it is impossible to throw an error.
+function floatFromFpLiteral(string digits) returns float {
+    float|error f = float:fromString(digits);
```
jclark: Just use `return checkpanic float:fromString(digits);`

compiler/modules/front/const.bal
```diff
@@ -517,47 +520,84 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
 }
 
-function floatFromDecimalLiteral(string digits) returns float|error {
-    return float:fromString(digits);
+// Even if the integer literal is parsed correctly,
+// overflow should throw an error.
+function intFromLiteral(FoldContext cx, s:IntLiteralExpr expr) returns int|FoldError {
+    int|error i = s:intFromIntLiteral(expr.base, expr.digits);
+    if i is error {
+        return cx.semanticErr("invalid int literal", expr.startPos);
```
jclark: Previous code was preserving the error from s:intFromIntLiteral in the `cause`.

compiler/modules/front/const.bal
```diff
@@ -517,47 +520,84 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
-        return t:isSubtypeSimple(t:intersect(semType, t:union(t:FLOAT, t:INT)), t:FLOAT);
+    return f;
+}
+
+// Even if the decimal floating point literal is parsed correctly,
+// overflow should throw an error.
+function decimalFromFpLiteral(FoldContext cx, string decimalStr, Position pos) returns decimal|FoldError {
+    decimal|error d = decimal:fromString(decimalStr);
+    if d is error {
+        return cx.semanticErr("invalid decimal literal", pos);
```
jclark: Use spec consistent terminology: "invalid decimal floating point number".

jclark: Also preserve the error from `decimal:fromString` in cause.

jclark (changes requested):

runtime/print.c
KavinduZoysa: Fixed

compiler/modules/nback/symbol.bal
KavinduZoysa: Fixed

compiler/modules/nback/symbol.bal
KavinduZoysa: Changed

compiler/modules/nback/scaffold.bal
KavinduZoysa: Changed

compiler/modules/nback/scaffold.bal
KavinduZoysa: Changed

compiler/modules/nback/scaffold.bal
KavinduZoysa: Changed

compiler/modules/nback/scaffold.bal
KavinduZoysa: Changed

compiler/modules/front.syntax/parseExpr.bal
KavinduZoysa: Fixed

compiler/modules/front.syntax/tests/testParser.bal
KavinduZoysa: Fixed

compiler/modules/front/const.bal
KavinduZoysa: Fixed

compiler/modules/front/const.bal
KavinduZoysa: Fixed

compiler/modules/front/const.bal
KavinduZoysa: Fixed

compiler/modules/front/const.bal
KavinduZoysa: Changed

KavinduZoysa:
> Need some error tests e.g.
> 
> ```
> int x = 1.2d;
> ```
> 
Added.
> Also need a test for a decimal that overflows or underflows.

Added `fe` tests for this because `decimal:fromString()` does not capture overflows and underflows in jballerina.
https://github.com/ballerina-platform/ballerina-lang/issues/33805

compiler/modules/types/core.bal
KavinduZoysa: @jclark, I did this change locally and it seems a considerable amount of places are changed. So shall I send a separate PR for this part?

KavinduZoysa: This change is done seperately in #769

compiler/modules/bir/bir.bal
KavinduZoysa: Fixed in #769

compiler/modules/front.syntax/ast.bal
KavinduZoysa: Fixed in #769

compiler/modules/front/const.bal
KavinduZoysa: Fixed in #769

KavinduZoysa: Fixed in #769

KavinduZoysa:
@jclark, I updated the PR with the changes of #769.

compiler/modules/types/core.bal
```diff
@@ -1007,7 +1011,7 @@ public function split(SemType t) returns SplitSemType  {
     }
 }
 
-public type SingleValue ()|boolean|int|float|string;
+public type SingleValue ()|boolean|int|float|decimal|string;
```
jclark, line 1014: There are 4 functions in this file that need to use this type for parameters.

jclark (changes requested):

compiler/modules/types/core.bal
KavinduZoysa: Changed.

compiler/modules/front.syntax/ast.bal
```diff
@@ -542,7 +542,7 @@ public type SingletonTypeDesc record {|
     (string|float|int|boolean|decimal) value;
 |};
 
-public type SubsetBuiltinTypeName "any"|"boolean"|"int"|"float"|"string"|"error";
+public type SubsetBuiltinTypeName "any"|"boolean"|"int"|"float"|"decimal"|"string"|"error";
```
jclark: You need to remove `decimal` from BuiltinTypeName

compiler/modules/front/const.bal
```diff
@@ -535,47 +538,66 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
-    float|error result = floatFromDecimalLiteral(expr.untypedLiteral);
-    if result is float {
-        return { startPos: expr.startPos, endPos: expr.endPos, value: result };
+function foldFpLiteralExpr(FoldContext cx, t:SemType? expectedType, s:FpLiteralExpr expr) returns s:ConstValueExpr|FoldError { 
+    var { typeSuffix, untypedLiteral, startPos } = expr;
+    float|decimal result;
+    if typeSuffix != () {
+        result = typeSuffix is s:FLOAT_TYPE_SUFFIX ? floatFromFpLiteral(untypedLiteral) : check decimalFromFpLiteral(cx, untypedLiteral, startPos);
+    }
+    else if expectedType == () || t:includesSome(expectedType, t:FLOAT) {
```
jclark: You can fix this condition to combine this branch and the last one,  Just add `|| !t:includesSome(expectedType, t:DECIMAL)`.

jclark (changes requested):

compiler/modules/front.syntax/ast.bal
jclark: You will also need to fix resolveBuiltinTypeDesc

jclark (changes requested):

compiler/modules/front.syntax/ast.bal
KavinduZoysa: Removed

compiler/modules/front/const.bal
KavinduZoysa: Changed

compiler/modules/front.syntax/ast.bal
KavinduZoysa: Changed

jclark (changes requested):
We should have tests using `decimal` with variables/parameters/return values of type `any`.

compiler/modules/front/const.bal
```diff
@@ -544,47 +547,63 @@ function foldVarRefExpr(FoldContext cx, t:SemType? expectedType, s:VarRefExpr ex
...
 }
 
-function expectsFloat(t:SemType? semType) returns boolean {
-    if semType == () {
-        return false;
-    }
-    else {
-        return t:isSubtypeSimple(t:intersect(semType, t:union(t:FLOAT, t:INT)), t:FLOAT);
+// Since the binary floating point literal is parsed correctly,
+// it is impossible to throw an error.
```
jclark: We don't "throw" errors in Ballerina. We either panic or return errors. (Applies to next two functions also.)

jclark (changes requested):

compiler/modules/front/const.bal
KavinduZoysa: Changed

KavinduZoysa:
> We should have tests using `decimal` with variables/parameters/return values of type `any`.

Added more tests

jclark:
Also make sure we can store decimal values inside lists/mappings with a member type of any.

KavinduZoysa:
> Also make sure we can store decimal values inside lists/mappings with a member type of any.

Added more tests

compiler/testSuite/11-decimal/any1-v.bal
```diff
@@ -0,0 +1,40 @@
...
+
+function bar(any a) returns any {
+    return a;
+}
+
+public function main() {
+    any a1 = 1d;
+    io:println(a1); // @output 1
+    any a2 = 1e6144d;
+    io:println(a2); // @output 1.000000000000000000000000000000000E+6144
```
jclark, line 25: How is that right?

jclark (changes requested):

compiler/testSuite/11-decimal/any1-v.bal
KavinduZoysa: @jclark, I went through decNumber source code to find the reason behind that. Please go through the below table.

|decimal string|output of decQuadFromString|exponent from decQuadGetExponent|
|---|----|---|
|1E6110|1E+6110|6110|
|1E6111|1E+6111|6111|
|1E6112|1.0E+6112|6111|
|1E6113|1.00E+6113|6111|
|1E6114|1.000E+6114|6111|

- The maximum value of the adjusted exponent(E_max) is `6144`
- When the exponent is adjusted, the coefficient is represented in scientific notation
- But when storing within 128 bits, the coefficient is a positive integer or zero
- Therefore when storing the exponent, the maximum exponent should be considered as `6111` (=6144 - (34 - 1))
- For example, consider 1E6113,
     - exponent is greater than 6111
     - in order to reduce the exponent to 6111, the coefficient should be changed as 100
     - now coefficient = 100, exponent = 6111
     - Therefore when printing we can see 1.00E+6113

Please check the source code https://github.com/dnotq/decNumber/blob/8852ef5b4339d611f493e94a791f3543b6ee81cd/decNumber-icu-368/decCommon.c#L417
In this line `DECEMAX-(DECPMAX-1)` means `6144-(34-1)`

jclark (changes requested):
> @jclark, I went through decNumber source code to find the reason behind that.

Please put a concise summary of that in the test case.

KavinduZoysa:
> > @jclark, I went through decNumber source code to find the reason behind that.
> 
> Please put a concise summary of that in the test case.

Added

compiler/testSuite/11-decimal/any1-v.bal
manuranga: This is lot of technical details, but what is the bottom line? Is it a bug? is it because library doesn't guarantee (close to) shortest string? Is there a different function we can call? what happens in GCC?

KavinduZoysa: This is not a bug. Since the output of `toString` function is represented in scientific notation, we can see these trailing zeros. I have tried decQuadReduce method and it does not remove these trailing zeros. I will try GCC function and update.

manuranga: From what I understand you can't add trailing zeros to decimal because `1.0d !== 1.00d`.

jclark: This is a bit strange, but I think it is correct behavior because the exponent is limited to 6111.

PR #756, opened 2021-12-07
Remove bir:OrderType

### LINKED ISSUE #752: Get rid of bir:OrderType
This is just creating complexity without adding value.

manuranga:
Fixes #752

### DIFF
--- a/compiler/modules/types/core.bal
+++ b/compiler/modules/types/core.bal
@@ -171,6 +171,13 @@ public type BddMemo record {|
 
 type BddMemoTable table<BddMemo> key(bdd);
 
+type ComparableMemo record {|
+    readonly SemType semType1;
+    readonly SemType semType2;
+    // SUBSET We assume recursive types aren't comparable, reconsider after adding tuples.
+    boolean comparable = false;
+|};
+
 // Operations on types require a Context.
 // There can be multiple contexts for the same Env.
 // Whereas an Env is isolated, a Context is not isolated.
@@ -181,6 +188,8 @@ public class Context {
     BddMemoTable listMemo = table [];
     BddMemoTable mappingMemo = table [];
     BddMemoTable functionMemo = table [];
+    final table<ComparableMemo> key(semType1, semType2) comparableMemo = table [];
+
     function init(Env env) {
         self.env = env;
     }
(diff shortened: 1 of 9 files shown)


compiler/modules/types/core.bal
```diff
@@ -171,6 +171,12 @@ public type BddMemo record {|
 
 type BddMemoTable table<BddMemo> key(bdd);
 
+type ComparableMemo record {|
+    readonly [SemType, SemType] semTypes;
+    // XXX We assume recursive types aren't comparable, reconsider after adding tuples.
```
jclark: Use `// SUBSET` comment here

compiler/modules/types/core.bal
```diff
@@ -181,6 +187,7 @@ public class Context {
     BddMemoTable listMemo = table [];
     BddMemoTable mappingMemo = table [];
     BddMemoTable functionMemo = table [];
+    final table<ComparableMemo> key(semTypes) comparableMemo = table[];
```
jclark: Blank line after this and space before `[]`.

compiler/modules/types/core.bal
```diff
@@ -171,6 +171,12 @@ public type BddMemo record {|
 
 type BddMemoTable table<BddMemo> key(bdd);
 
+type ComparableMemo record {|
+    readonly [SemType, SemType] semTypes;
```
jclark: I think it's simpler to use multiple keys here:

```
  readonly SemType semType1;
  readonly SemType semType2;
```

then `key(semType1, semType2)`.

compiler/modules/types/core.bal
```diff
@@ -757,6 +764,30 @@ public function uniformTypeCode(UniformTypeBitSet bitSet) returns UniformTypeCod
...
+}
+
+function comparableInternal(Context cx, SemType t1, SemType t2) returns boolean {
+    SemType semType = diff(union(t1, t2), NIL);
+    if isSubtypeSimple(semType, SIMPLE_OR_STRING) {
+        int nOrderings = lib:bitCount(widenToUniformTypes(semType));
+        return nOrderings <= 1;
+    }
+    if isSubtypeSimple(semType, LIST) {
+        return comparable(cx, listMemberType(cx, t1, ()), listMemberType(cx, t2, ()));
```
jclark: Does this deal with the case where t1 or t1 has type `()`? If so, we should have a test for that.

We also need a SUBSET comment for tuples.

jclark (changes requested):
A nice lot of code being deleted here!

compiler/modules/types/core.bal
manuranga: There are already tests covering this in [order2-v.bal](https://github.com/ballerina-platform/nballerina/blob/6da311d71760793685621edea27e5ccb3cf4a98d/compiler/testSuite/11-nested/order2-v.bal#L9).

`comparable(NIL, INT_ARR) = true` because `listMemberType(NIL) = NEVER` and `comparable(NEVER, INT) = true`.

Maybe it's wrong for `listMemberType` to return `NEVER` in this case.

compiler/modules/types/core.bal
```diff
@@ -757,6 +766,31 @@ public function uniformTypeCode(UniformTypeBitSet bitSet) returns UniformTypeCod
...
+
+function comparableInternal(Context cx, SemType t1, SemType t2) returns boolean {
+    SemType semType = diff(union(t1, t2), NIL);
+    if isSubtypeSimple(semType, SIMPLE_OR_STRING) {
+        int nOrderings = lib:bitCount(widenToUniformTypes(semType));
+        return nOrderings <= 1;
+    }
+    if isSubtypeSimple(semType, LIST) {
+        // SUBSET need to iterate members when tuples are supported
+        return comparable(cx, listMemberType(cx, t1, ()), listMemberType(cx, t2, ()));
```
jclark: We ought to memoize just this case (when we have two list types). There is no point memoizing before this because we can do it very quickly on the fly and there is no possibility of recursion i.e. the two functions and comparable and comparableList.

jclark (changes requested):

PR #764, opened 2021-12-12
Compiler support for decimal arithmetic operations

### LINKED ISSUE #759: Implement binary arithmetic operations for `decimal` type
Implement  `+`, `-`, `*`, `/`,`%` for decimal type

Part of #289

KavinduZoysa:
## Purpose
Fixes #759


KavinduZoysa:
~~This PR depends on #755~~
~~The last two commits are relevant to this PR.~~

compiler/modules/bir/bir.bal
```diff
@@ -275,6 +278,14 @@ public type FloatArithmeticBinaryInsn readonly & record {|
     FloatOperand[2] operands;
 |};
 
+public type DecimalArithmeticBinaryInsn readonly & record {|
```
jclark, line 284: Add a comment saying this is a PPI.

compiler/modules/nback/number.bal
```diff
@@ -86,6 +86,37 @@ function buildFloatArithmeticBinary(llvm:Builder builder, Scaffold scaffold, bir
...
+    llvm:Value panicCode = builder.extractValue(resultWithErr, 1);
+    builder.condBr(builder.iCmp("ugt", panicCode, llvm:constInt("i64", 0)), errBlock, continueBlock);
+    builder.positionAtEnd(errBlock);
+    builder.store(buildErrorForPanic(builder, scaffold, panicCode, insn.pos), scaffold.panicAddress());
+    builder.br(scaffold.getOnPanic());
+    builder.positionAtEnd(continueBlock);
+    buildStoreDecimal(builder, scaffold, builder.extractValue(resultWithErr, 0), insn.result);                                  
+}
+
+final readonly & map<string> decimalArithmeticFuncName = {
```
jclark: Put this below (under floatArithmeticOps) and leave out the `decimal_` prefix. Then use this to build at module-init time a map from the operator to the RuntimeFunction.

jclark (changes requested):

compiler/modules/nback/number.bal
```diff
@@ -12,6 +12,34 @@ final RuntimeFunction floatToIntFunction = {
     attrs: ["nounwind", "readnone", "speculatable", "willreturn"]
 };
 
+final map<RuntimeFunction> decimalArithmeticFuncs = {};
```
jclark: This is not the right way to do things.

- The type needs to be `readonly`
- Don't use `init`
- Initialize the variable with a call to a function that returns the map

jclark (changes requested):

compiler/modules/bir/bir.bal
KavinduZoysa: Added

compiler/modules/nback/number.bal

jclark:
Need tests for decimal `const` both with and without explicit decimal type.

KavinduZoysa:
> Need tests for decimal `const` both with and without explicit decimal type.

Added more tests

runtime/panic.c
```diff
@@ -33,6 +33,7 @@ DECL_PANIC_MESSAGE(PANIC_LIST_STORE, "bad list store");
 DECL_PANIC_MESSAGE(PANIC_MAPPING_STORE, "bad mapping store");
 DECL_PANIC_MESSAGE(PANIC_NO_FILLER, "list has no filler");
 DECL_PANIC_MESSAGE(PANIC_INTERNAL_ERROR, "internal error");
+DECL_PANIC_MESSAGE(PANIC_INVALID_DECIMAL, "invalid decimal");
```
jclark, line 36: Don't think this should be here.

compiler/modules/front/const.bal
```diff
@@ -277,6 +277,15 @@ function foldBinaryArithmeticExpr(FoldContext cx, t:SemType? expectedType, s:Bin
             }
             return foldedBinaryConstExpr(f, t:FLOAT, leftExpr, rightExpr);
         }
+        else if left is decimal && right is decimal {
+            decimal|error result = trap decimalArithmeticEval(expr.arithmeticOp, left, right);
```
jclark, line 286: This is not correct because of #783.  It will need code similar to FloatZero above. For now, just put in a comment here.

compiler/modules/nback/number.bal
```diff
@@ -12,6 +12,20 @@ final RuntimeFunction floatToIntFunction = {
     attrs: ["nounwind", "readnone", "speculatable", "willreturn"]
 };
 
+final readonly & map<RuntimeFunction> decimalArithmeticFuncs = decimalArithmeticFuncsInit();
+
+function decimalArithmeticFuncsInit() returns readonly & map<RuntimeFunction> {
+    map<RuntimeFunction> m = {};
+    llvm:FunctionType ty = {
```
jclark: Declare as `readonly &`.

compiler/modules/nback/number.bal
```diff
@@ -12,6 +12,20 @@ final RuntimeFunction floatToIntFunction = {
     attrs: ["nounwind", "readnone", "speculatable", "willreturn"]
 };
 
+final readonly & map<RuntimeFunction> decimalArithmeticFuncs = decimalArithmeticFuncsInit();
+
+function decimalArithmeticFuncsInit() returns readonly & map<RuntimeFunction> {
```
jclark: Init is not the right name for something that returns a value. Look at what I have done elsewhere in similar cases (probably something beginning with `create`).

jclark (changes requested):

runtime/panic.c
KavinduZoysa: @jclark, this is a bug because I have to do this change under #704.

This is the comment that was discussed about this panic code
https://github.com/ballerina-platform/nballerina/pull/704#issuecomment-974738928

KavinduZoysa: Shall I change `invalid decimal` into `invalid operation`?

compiler/modules/front/const.bal
KavinduZoysa: Added a comment

compiler/modules/nback/number.bal

compiler/modules/nback/number.bal

compiler/modules/nback/number.bal
```diff
@@ -95,6 +109,21 @@ function buildFloatArithmeticBinary(llvm:Builder builder, Scaffold scaffold, bir
...
 }
 
+function buildDecimalArithmeticBinary(llvm:Builder builder, Scaffold scaffold, bir:DecimalArithmeticBinaryInsn insn) {
+    llvm:Value lhs = buildDecimal(builder, scaffold, insn.operands[0]);
+    llvm:Value rhs = buildDecimal(builder, scaffold, insn.operands[1]);
+    llvm:Value resultWithErr = <llvm:Value>builder.call(scaffold.getRuntimeFunctionDecl(decimalArithmeticFuncs.get(insn.op)), [lhs, rhs]);
+    llvm:BasicBlock continueBlock = scaffold.addBasicBlock();
+    llvm:BasicBlock errBlock = scaffold.addBasicBlock();
+    llvm:Value panicCode = builder.extractValue(resultWithErr, 1);
+    builder.condBr(builder.iCmp("ugt", panicCode, llvm:constInt("i64", 0)), errBlock, continueBlock);
```
manuranga, line 119: Can't we branch on the panic code it self?
```suggestion
    builder.condBr(panicCode, errBlock, continueBlock);
```

jclark: panicCode isn't an i1. Doesn't condBr need an i1?

manuranga:
@KavinduZoysa do we have a test for `0d/0d`, I don't see one in runtime either, if so shall we add one.

jclark:
We should definitely add 0d/0d.

PR #769, opened 2021-12-14
Refactor record `Value`

KavinduZoysa:
## Purpose
$subject

### DIFF
--- a/compiler/modules/bir/bir.bal
+++ b/compiler/modules/bir/bir.bal
@@ -210,8 +210,7 @@ public type Insn
     |BranchInsn|CondBranchInsn|CatchInsn|PanicInsn|ErrorConstructInsn;
 
 public type Operand ConstOperand|Register;
-public type SimpleConstOperand ()|boolean|int|float;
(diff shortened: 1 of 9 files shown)


compiler/modules/bir/bir.bal
```diff
@@ -210,8 +210,7 @@ public type Insn
     |BranchInsn|CondBranchInsn|CatchInsn|PanicInsn|ErrorConstructInsn;
 
 public type Operand ConstOperand|Register;
-public type SimpleConstOperand ()|boolean|int|float;
-public type ConstOperand SimpleConstOperand|string;
+public type ConstOperand t:SingleValue;
```
manuranga, line 213: Why not get rid of this altogether ?

jclark:
> Why not get rid of this altogether ? 

I prefer to keep it because they may diverge. Ballerina allows structures to be const, but we don't support this yet.

jclark (approved):
I am merging this before #755 because there are several functions in core.bal that you can only fix to use SingleValue once SingleValue includes decimal.

PR #775, opened 2021-12-18
Allow for `decimal` type descriptor

### LINKED ISSUE #766: Allow `decimal` as a type descriptor
Support decimal in var-decl.
Eg: `decimal x = 10d;`

Part of #289

KavinduZoysa:
## Purpose
Fixes #766

### DIFF
--- a/compiler/modules/nback/scaffold.bal
+++ b/compiler/modules/nback/scaffold.bal
@@ -501,9 +501,9 @@ function semTypeRepr(t:SemType ty) returns Repr {
     if w == t:NEVER {
         panic err:impossible("allocate register with never type");
     }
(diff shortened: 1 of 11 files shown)


compiler/testSuite/11-decimal/optinal1-v.bal
```diff
@@ -0,0 +1,9 @@
+import ballerina/io;
```
jclark: Fix spelling of filename of this test.

jclark (changes requested):
Test cases need to be in 12-decimal.

compiler/testSuite/11-decimal/optinal1-v.bal
KavinduZoysa: Changed

KavinduZoysa:
> Test cases need to be in 12-decimal.

Changed

PR #776, opened 2021-12-18
Compiler support for decimal negate

### LINKED ISSUE #779: Implement unary minus for `decimal`
This requires adding a DecimalNegate insn.

We also need to add runtime support for this.

KavinduZoysa:
## Purpose
Fixes #779


compiler/modules/nback/number.bal
```diff
@@ -12,6 +12,15 @@ final RuntimeFunction floatToIntFunction = {
     attrs: ["nounwind", "readnone", "speculatable", "willreturn"]
 };
 
+final RuntimeFunction decimalNegateFunction = {
+    name: "decimal_neg",
+    ty: {
+        returnType: LLVM_TAGGED_PTR,
+        paramTypes: [LLVM_TAGGED_PTR]
+    },
+    attrs: ["readnone"]
```
jclark: readonly not readnone.  decimal_const is wrong too.

jclark (changes requested):

compiler/modules/nback/number.bal

compiler/testSuite/11-decimal/neg1-v.bal
```diff
@@ -0,0 +1,17 @@
+import ballerina/io;
```
jclark: We are in subset 12 now, so 12-decimal

compiler/modules/nback/number.bal
```diff
@@ -12,6 +12,15 @@ final RuntimeFunction floatToIntFunction = {
     attrs: ["nounwind", "readnone", "speculatable", "willreturn"]
 };
 
+final RuntimeFunction decimalNegateFunction = {
```
jclark: This needs to be called decimalNegFunction (name is always same as runtime function name in camel case.)

compiler/testSuite/11-decimal/neg1-v.bal
```diff
@@ -0,0 +1,17 @@
...
+public function main() {
```
jclark: You need to have cases where the operand is const.

jclark (changes requested):

compiler/testSuite/11-decimal/neg1-v.bal
KavinduZoysa: Changed

compiler/modules/nback/number.bal

compiler/testSuite/11-decimal/neg1-v.bal
KavinduZoysa: Added test cases

Issue #272, opened 2021-07-30
Types should be narrowed as for `else` after `if` that cannot complete normally


jclark:
Follow on from #129 and #156

Specified https://github.com/ballerina-platform/ballerina-spec/issues/73

manuranga:
Simple case can be handled by adding a narrowing just after `if` (or `else`).
```ballerina
import ballerina/io;

public function main() {
    int|boolean x = 2;
    if x is boolean {
        boolean _ = x;
        return;
    }
    int _ = x;
}
```
A more involved case:

```ballerina
import ballerina/io;

public function main() {
    int|boolean|float x = 0.1;
    if x is int|boolean {
        if x is int {
            return;
        }
        boolean _ = x;
    }
    boolean|float _ = x;
}
```
At the end of the outer `if`, we need `x` to be narrowed to `narrowing.ifFalse ∪ narrowing-at-end-of-body`.

Another case:
```ballerina
public function func(boolean y) {
    int|boolean|float x = 0.1;
    if y == true {
        if x is int|boolean {
            return;
        }
    }
    else {
        if x is boolean|float {
            return;
        }
    }
    float|int _ = x;
}
```
Merging of narrowings has to happen for all `if`s. Even the ones without a narrowing in the condition.

manuranga:
My current plan is to modify `codeGenStmts` to return bindings, and at the end of if-else combine bindings from both paths into a new binding.

manuranga:
Was able to handle above 3rd case with above approach. That is, get the bindings from if-body and else-body via modified `codeGenStmts`, diff it with current binding, union two sets of bindings and add `CondNarrowInsn`s after else body. I am not giving the correct `basis` yet.

Two more interesting cases that got handled : [multi-var-narrow](https://gist.github.com/manuranga/fafc963fd152b3b862973d2bd8401c08), [tighten-inner.bal ](https://gist.github.com/manuranga/bec1902d3c12dd65c8f0e15569784f03)

PR #788, opened 2021-12-20
Narrow statements

manuranga:
Part of #272

### DIFF
--- a/compiler/modules/front/codeGen.bal
+++ b/compiler/modules/front/codeGen.bal
@@ -28,9 +28,16 @@ type Binding record {|
     Binding? unnarrowed = ();
 |};
 
+type StmtNarrowing record {|
+    Binding binding;
+    t:SemType ifCompletesNormally;
+    bir:Result basis;
+|};
+
 type StmtEffect record {|
     bir:BasicBlock? block;
     Binding? bindings = ();
+    StmtNarrowing[] narrowings = [];
     int[] assignments = [];
 |};
 
@@ -40,7 +47,7 @@ type ExprEffect record {|
     // This is non-nil when the expression is a boolean expression that case causes the type
     // of a variable to be narrowed when the expression is used as a condition.
     // We do not have `&&` and `||` yet, so only one variable is narrowed.
-    Narrowing? narrowing = ();
+    ExprNarrowing? narrowing = ();
     // This is non-nil when the expression is a variable reference.
     Binding? binding = ();
 |};
@@ -67,7 +74,7 @@ type RegExprEffect record {|
 
 // This describes the narrowing of the type of a variable
 // caused by the use of a boolean expression in a condition.
-type Narrowing record {|
+type ExprNarrowing record {|
     Binding binding;
     t:SemType ifTrue;
     t:SemType ifFalse;
@@ -285,7 +292,7 @@ function codeGenFunction(ModuleSymbols mod, s:FunctionDefn defn, bir:FunctionSig
         bir:Register reg = cx.createVarRegister(signature.paramTypes[i], param.name, param.namePos);
         bindings = { name: <string>param.name, reg, prev: bindings, isFinal: true };
     }
-    var { block: endBlock } = check codeGenStmts(cx, startBlock, { bindings }, defn.body);
+    var { block: endBlock } = check codeGenStmtBlock(cx, startBlock, { bindings }, defn.body);
     if endBlock != () {
         bir:RetInsn ret = { operand: (), pos: defn.body.closeBracePos };
         endBlock.insns.push(ret);
@@ -318,10 +325,18 @@ function codeGenOnPanic(CodeGenContext cx, Position pos) {
     }
 }
 
-function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialEnv, s:StmtBlock block) returns CodeGenError|StmtEffect {
+// If block of stmts doesn't complete normally, will return empty narrowings and assignments.
+function codeGenStmtBlock(CodeGenContext cx, bir:BasicBlock bb, Environment initialEnv, s:StmtBlock block, StmtNarrowing? initialNarrowing = ()) returns CodeGenError|StmtEffect {
     bir:BasicBlock? curBlock = bb;
     Environment env = environmentCopy(initialEnv);
     final int startRegister = cx.nextRegisterNumber();
+    // SUBSET with && || initialNarrowing will need to become a list
+    StmtNarrowing[] narrowings = initialNarrowing != () ? [initialNarrowing] : [];
+    // initialNarrowing is processed under the cloned env so the effects of it is confined to the block
+    updateAssignments(env, { block: bb, narrowings });
+    addNarrowings(cx, bb, env, narrowings, block.startPos);
+    int lastStmtIndex = block.stmts.length() - 1;
+    int stmtIndex = 0;
     foreach var stmt in block.stmts {
         StmtEffect effect;
         if curBlock == () {
@@ -367,15 +382,32 @@ function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialE
         }
         if curBlock == () {
             env.assignments.setLength(0);
+            narrowings.setLength(0);
         }
         else {
-            env.assignments.push(...effect.assignments);
+            updateAssignments(env, effect);
+            // Compound statements will gen narrowings post-block, no need to narrow after last stmt
+            if stmtIndex != lastStmtIndex {
+                addNarrowings(cx, curBlock, env, effect.narrowings, stmt.endPos);
+            }
+            addIntersectStmtNarrowings(narrowings, effect.narrowings);
         }
-    }                
+        stmtIndex += 1;
+    }
     check unusedLocalVariables(cx, env, initialEnv.bindings);
     int[] assignments = [];
     addAssignments(assignments, env.assignments, startRegister);
-    return { block: curBlock, assignments };
+    return { block: curBlock, assignments, narrowings };
+}
+
+function updateAssignments(Environment env, StmtEffect effect) {
+    foreach var n in effect.narrowings {
+        int? invalidated = env.assignments.indexOf(unnarrowBinding(n.binding).reg.number);
+        if invalidated != () {
+            _ = env.assignments.remove(invalidated);
+        }
+    }
+    env.assignments.push(...effect.assignments);
 }
 
 function unusedLocalVariables(CodeGenContext cx, Environment env, Binding? bindingLimit) returns CodeGenError? {
@@ -417,7 +449,7 @@ function codeGenForeachStmt(CodeGenContext cx, bir:BasicBlock startBlock, Enviro
     loopHead.insns.push(branch);
     cx.pushLoopContext(exit, ());
     Binding loopBindings = { name: varName, reg: loopVar, prev: env.bindings, isFinal: true };
-    var { block: loopEnd, assignments } = check codeGenStmts(cx, loopBody, { bindings: loopBindings }, stmt.body);
+    var { block: loopEnd, assignments } = check codeGenStmtBlock(cx, loopBody, { bindings: loopBindings }, stmt.body);
 
     bir:BasicBlock? loopStep = cx.loopContinueBlock();
     if loopEnd != () {
@@ -472,7 +504,7 @@ function codeGenWhileStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environm
     }
     afterCondition.insns.push(branch);
     cx.pushLoopContext(exit, loopHead);
-    var { block: loopEnd, assignments } = check codeGenStmts(cx, loopBody, env, stmt.body);
+    var { block: loopEnd, assignments } = check codeGenStmtBlock(cx, loopBody, env, stmt.body);
     if loopEnd != () {
         loopEnd.insns.push(branchToLoopHead);
         check validLoopAssignments(cx, assignments);
@@ -624,6 +656,7 @@ function codeGenMatchStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environm
     foreach int clauseIndex in 0 ..< stmt.clauses.length() {
         s:MatchClause clause = stmt.clauses[clauseIndex];
         bir:BasicBlock stmtBlock = clauseBlocks[clauseIndex];
+        StmtNarrowing? narrowings = ();
         Environment clauseEnv = env;
         // Do type narrowing
         if binding != () {
@@ -650,10 +683,10 @@ function codeGenMatchStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environm
             if basis != () {
                 // Will need readOnlyIntersect when we have proper match patterns
                 t:SemType narrowedType = t:intersect(matchedType, clauseLooksLike[clauseIndex]);
-                clauseEnv = codeGenNarrowing(cx, stmtBlock, env, binding, narrowedType, basis, clause.opPos);
+                narrowings = { basis, ifCompletesNormally: narrowedType, binding };
             }
         } 
-        var { block: stmtBlockEnd, assignments: blockAssignments } = check codeGenStmts(cx, stmtBlock, clauseEnv, clause.block);
+        var { block: stmtBlockEnd, assignments: blockAssignments } = check codeGenStmtBlock(cx, stmtBlock, clauseEnv, clause.block, narrowings);
         if stmtBlockEnd == () {
             continue;
         }
@@ -702,7 +735,7 @@ function maybeCreateBasicBlock(CodeGenContext cx, bir:BasicBlock? block) returns
 
 function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environment env, s:IfElseStmt stmt) returns CodeGenError|StmtEffect {
     var { condition, ifTrue, ifFalse } = stmt;
-    var { result: operand, block: branchBlock, narrowing } = check codeGenConditionalExpr(cx, startBlock, env, condition);
+    var { result: operand, block: branchBlock, narrowing: condNarrowing } = check codeGenConditionalExpr(cx, startBlock, env, condition);
     if operand is boolean {
         s:StmtBlock? taken;
         s:StmtBlock? notTaken;
@@ -727,7 +760,7 @@ function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
             return cx.semanticErr("unreachable code", s:range(errStmt));
         }
         if taken is s:StmtBlock {
-            return codeGenStmts(cx, branchBlock, env, taken);
+            return codeGenStmtBlock(cx, branchBlock, env, taken);
         }
         else {
             // if false whithout else block
@@ -736,8 +769,8 @@ function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
     }
     else {
         bir:BasicBlock ifBlock = cx.createBasicBlock();
-        Environment ifEnv = narrowing == () ? env : codeGenIfElseNarrowing(cx, ifBlock, env, narrowing, true, stmt.condition.startPos);
-        var { block: ifContBlock, assignments } = check codeGenStmts(cx, ifBlock, ifEnv, ifTrue);
+        StmtNarrowing? condIfNarrowing = stmtNarrowingFromExprNarrowing(condNarrowing, true);
+        var { block: ifContBlock, assignments, narrowings: ifNarrowings } = check codeGenStmtBlock(cx, ifBlock, env, ifTrue, condIfNarrowing);
         bir:BasicBlock contBlock;
         if ifFalse == () {
             // just an if branch
@@ -748,13 +781,15 @@ function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
                 bir:BranchInsn branch = { dest: contBlock.label, pos: stmt.condition.startPos };
                 ifContBlock.insns.push(branch);
             }
-            return { block: contBlock, assignments };    
+            StmtNarrowing? elseNarrowings = stmtNarrowingFromExprNarrowing(condNarrowing, false);
+            StmtNarrowing[] narrowings = combineIfElseNarrowings(ifNarrowings, ifContBlock != (), elseNarrowings != () ? [elseNarrowings] : [], true);
+            return { block: contBlock, assignments, narrowings };
         }
         else {
             // an if and an else
             bir:BasicBlock elseBlock = cx.createBasicBlock();
-            Environment elseEnv = narrowing == () ? env : codeGenIfElseNarrowing(cx, elseBlock, env, narrowing, false, stmt.condition.startPos);
-            var { block: elseContBlock, assignments: elseAssignments } = check codeGenStmts(cx, elseBlock, elseEnv, ifFalse);
+            StmtNarrowing? condElseNarrowing = stmtNarrowingFromExprNarrowing(condNarrowing, false);
+            var { block: elseContBlock, assignments: elseAssignments, narrowings: elseNarrowings } = check codeGenStmtBlock(cx, elseBlock, env, ifFalse, condElseNarrowing);
             bir:CondBranchInsn condBranch = { operand, ifTrue: ifBlock.label, ifFalse: elseBlock.label, pos: stmt.condition.startPos };
             branchBlock.insns.push(condBranch);
             if ifContBlock == () && elseContBlock == () {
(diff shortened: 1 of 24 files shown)


compiler/modules/front/codeGen.bal
```diff
@@ -26,9 +27,18 @@ type Binding record {|
     Binding? unnarrowed = ();
 |};
 
+// pr-todo: can this be a sub type?
+type StmtNarrowing record {|
+    Binding binding;
+    t:SemType completeNormally;
```
jclark: Call this `ifCompletesNormally` for symmetry with Narrowing.

manuranga:
I think I have to change the logic to  
`ifStmtNarrowing = (ifExpNarrowing ⊕ ifBodyNarrwoing) U (elseExpNarrowing ⊕ elseBodyNarrowing)`
(here ⊕ means override, ie: narrow further)

compiler/modules/front/codeGen.bal
```diff
@@ -63,6 +73,7 @@ type RegExprEffect record {|
     bir:Register result;
 |};
 
+// pr-todo: rename to ExprNarrowing
```
jclark: Agree this should be renamed

compiler/modules/front/codeGen.bal
```diff
@@ -363,11 +374,23 @@ function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialE
         else {
             env.assignments.push(...effect.assignments);
         }
+
+        StmtNarrowing[] narrowings = effect.narrowings;
```
jclark: We need to be clear about the invariant the code is maintaining. Is it that a statement never returns narrowings for variables that were declared within it?

compiler/modules/front/codeGen.bal
```diff
@@ -716,11 +739,55 @@ function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
...
+            // pr-todo: this should be pushUnion
+            ns.push(elseNarrowing);
+        }
+    }
+    return ns;
+}
+
+
+// if already exists in the list, union, else append
+function pushUnion(StmtNarrowing[] list, StmtNarrowing annex) {
```
jclark: I would call this something like `addStmtNarrowing`.

compiler/modules/front/codeGen.bal
```diff
@@ -363,11 +374,23 @@ function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialE
...
+            //pr-todo: narrowings at the end of a function is useless, can we prevent?
+            foreach var n in narrowings {
+                { bindings: bindings } = codeGenNarrowing(cx, curBlock, env, n.binding, n.completeNormally, n.basis, stmt.startPos);
+                env.bindings = bindings;
+            }
+        }
     }                
     check unusedLocalVariables(cx, env, initialEnv.bindings);
     int[] assignments = [];
     addAssignments(assignments, env.assignments, startRegister);
```
jclark, line 399: I am not clear what your strategy is for dealing with assignments...

jclark (changes requested):

compiler/modules/front/codeGen.bal
manuranga: Yes, each statement should return narrowing that is to be visible to the next statement. That means compounds statements shouldn't return narrowing internal to them.

compiler/modules/front/codeGen.bal
manuranga: I have a counterpart to this now, so picked a different name.

compiler/modules/front/codeGen.bal
manuranga: Assignments worked even without me doing anything. Need to test a bit more to see if that works for all cases.

compiler/modules/front/codeGen.bal
```diff
@@ -605,10 +628,10 @@ function codeGenMatchStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environm
             if basis != () {
                 // Will need readOnlyIntersect when we have proper match patterns
                 t:SemType narrowedType = t:intersect(matchedType, clauseLooksLike[clauseIndex]);
-                clauseEnv = codeGenNarrowing(cx, stmtBlock, env, binding, narrowedType, basis, clause.opPos);
+                env.bindings = codeGenNarrowing(cx, stmtBlock, env, binding, narrowedType, basis, clause.opPos);
```
jclark: Here you are mutating env, where the previous code was creating a new Env.  I doubt that is correct.

compiler/modules/front/codeGen.bal
```diff
@@ -716,22 +744,67 @@ function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
...
+            }
+            return { block: contBlock, assignments, narrowings };
         }
     }
 }
 
-function codeGenIfElseNarrowing(CodeGenContext cx, bir:BasicBlock bb, Environment env, Narrowing narrowing, boolean condition, Position pos) returns Environment {
+// If already exists in the list, union, else append
+function broadenStmtNarrowings(StmtNarrowing[] list, StmtNarrowing annex) {
+    foreach int i in 0..<list.length() {
```
jclark: Space after `<`.

compiler/modules/front/codeGen.bal
```diff
@@ -716,22 +744,67 @@ function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
...
+            };
+            return;
+        }
+    }
+    list.push(annex);
+}
+
+// If already exists in the list, replace, else append
+function tightenStmtNarrowings(StmtNarrowing[] list, StmtNarrowing annex) {
+    foreach int i in 0..<list.length() {
```
jclark: Space after `<`

compiler/modules/front/codeGen.bal
```diff
@@ -716,22 +744,67 @@ function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
...
+                broadenStmtNarrowings(narrowings, n);
+            }
+            return { block: contBlock, assignments, narrowings };
         }
     }
 }
 
-function codeGenIfElseNarrowing(CodeGenContext cx, bir:BasicBlock bb, Environment env, Narrowing narrowing, boolean condition, Position pos) returns Environment {
+// If already exists in the list, union, else append
+function broadenStmtNarrowings(StmtNarrowing[] list, StmtNarrowing annex) {
```
jclark: `annex` is not a good word to use here.

compiler/modules/front/codeGen.bal
```diff
@@ -716,22 +744,67 @@ function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
...
+                binding: annex.binding,
+                ifCompletesNormally: t:union(annex.ifCompletesNormally, existing.ifCompletesNormally)
+            };
+            return;
+        }
+    }
+    list.push(annex);
+}
+
+// If already exists in the list, replace, else append
```
jclark: I would expect this to be intersecting. Why isn't it doing that?

compiler/modules/front/codeGen.bal
```diff
@@ -716,22 +744,67 @@ function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
...
+}
+
+function codeGenIfElseNarrowing(CodeGenContext cx, bir:BasicBlock bb, Environment env, ExprNarrowing exprNarrowing, boolean condition, Position pos) returns Environment {
+    StmtNarrowing narrowing = exprNarrowingStmtNarrowing(exprNarrowing, condition);
+    Binding bindings = codeGenNarrowing(cx, bb, env, narrowing.binding, narrowing.ifCompletesNormally, narrowing.basis, pos);
+    // pr-todo: find what happens to env.narrowing
+    return { bindings, assignments: env.assignments, narrowings: [narrowing] };
+}
+
+function exprNarrowingStmtNarrowing(ExprNarrowing narrowing, boolean condition) returns StmtNarrowing {
```
jclark: This does need a preposition. The Ballerina-ish name would be stmtNarrowingFromExprNarrowing.

compiler/modules/front/codeGen.bal
```diff
@@ -694,13 +717,18 @@ function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
                 bir:BranchInsn branch = { dest: contBlock.label, pos: stmt.condition.startPos };
                 ifContBlock.insns.push(branch);
             }
-            return { block: contBlock, assignments };    
+            StmtNarrowing[] narrowings = ifNarrowings;
+            if exprNarrowing != () {
+                StmtNarrowing elseNarrowing = exprNarrowingStmtNarrowing(exprNarrowing, false);
+                broadenStmtNarrowings(narrowings, elseNarrowing);
```
jclark: Why is this broadening?

compiler/modules/front/codeGen.bal
```diff
@@ -716,22 +744,67 @@ function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
...
+                ifCompletesNormally: t:union(annex.ifCompletesNormally, existing.ifCompletesNormally)
+            };
+            return;
+        }
+    }
+    list.push(annex);
+}
+
+// If already exists in the list, replace, else append
+function tightenStmtNarrowings(StmtNarrowing[] list, StmtNarrowing annex) {
```
jclark: We don't use broaden/tighten as a pair anywhere else. I would expect that the semantics were union and intersection and we could name accordingly.

compiler/modules/front/codeGen.bal
```diff
@@ -648,7 +671,7 @@ function maybeCreateBasicBlock(CodeGenContext cx, bir:BasicBlock? block) returns
 
 function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environment env, s:IfElseStmt stmt) returns CodeGenError|StmtEffect {
     var { condition, ifTrue, ifFalse } = stmt;
-    var { result: operand, block: branchBlock, narrowing } = check codeGenConditionalExpr(cx, startBlock, env, condition);
+    var { result: operand, block: branchBlock, narrowing: exprNarrowing } = check codeGenConditionalExpr(cx, startBlock, env, condition);
```
jclark: I suggest condNarrowing rather than exprNarrowing (since you are using if/elseNarrowings not stmtNarrowings below).

compiler/modules/front/codeGen.bal
```diff
@@ -359,15 +375,23 @@ function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialE
         }
         if curBlock == () {
             env.assignments.setLength(0);
+            env.narrowings.setLength(0);
         }
         else {
             env.assignments.push(...effect.assignments);
+            foreach var n in effect.narrowings {
+                tightenStmtNarrowings(env.narrowings, n);
+                //pr-todo: narrowings at the end of a function is useless, can we prevent?
```
jclark: Don't understand the comment

compiler/modules/front/codeGen.bal
```diff
@@ -716,22 +744,67 @@ function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
...
+// If already exists in the list, replace, else append
+function tightenStmtNarrowings(StmtNarrowing[] list, StmtNarrowing annex) {
+    foreach int i in 0..<list.length() {
+        StmtNarrowing existing = list[i];
+        if existing.binding.name == annex.binding.name {
+            list[i] = {
+                basis: { and: [annex.basis, existing.basis] },
+                binding: annex.binding,
+                // Assuming annex.ifCompletesNormally is a subset of existing.ifCompletesNormally
+                ifCompletesNormally: annex.ifCompletesNormally
```
jclark: It is not consistent to use `and` for the basis and not use `intersect` when computing the type.

jclark (changes requested):

jclark:
I thought about the assignments issue some more following our discussion yesterday. I think it should work to change what we store for each assignment to be both the register number of the variable assigned to and the number of allocated registers at the point of assignment.

Can we get this problem with assignments without #272?

manuranga:
After testing more, invalidation of assignments approach continues to work, but I need to do it not only in `codeGenStmts` but also in `codeGenIfElseStmt`, since assignments should be invalidated due to the if-else expression. I am trying to refactor it such that `codeGenNarrowing` is only called form `codeGenStmts`. This way `assignments` and `narrowings` are only mutated in `codeGenStmts`.

compiler/modules/front/codeGen.bal
```diff
@@ -312,11 +328,15 @@ function codeGenOnPanic(CodeGenContext cx, Position pos) {
...
-function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialEnv, s:StmtBlock block) returns CodeGenError|StmtEffect {
+// If block of stmts doesn't complete normally, will return empty narrowings and assignments.
+function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialEnv, s:StmtBlock block, StmtNarrowing[] initialNarrowings = []) returns CodeGenError|StmtEffect {
     bir:BasicBlock? curBlock = bb;
     Environment env = environmentCopy(initialEnv);
     final int startRegister = cx.nextRegisterNumber();
-    foreach var stmt in block.stmts {
+    applyEffect(env, { block: bb, narrowings: initialNarrowings });
+    codeGenNarrowings(cx, bb, env, initialNarrowings, block.startPos);
+    int lastStmt = block.stmts.length() - 1;
```
jclark: lastStmtIndex

compiler/modules/front/codeGen.bal
```diff
@@ -312,11 +328,15 @@ function codeGenOnPanic(CodeGenContext cx, Position pos) {
...
+    foreach var [i, stmt] in block.stmts.enumerate() {
```
jclark: Let's call it stmtIndex.

enumerate here is pretty but gratuitously inefficient for something that is used for every statement.

compiler/modules/front/codeGen.bal
```diff
@@ -758,24 +804,97 @@ function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
...
+        }
+        [false, true] => {
+            return elseNarrowings;
+        }
+        [false, false] => {
+            return [];
+        }
+    }
+    // JBUG
+    panic err:impossible();
```
jclark: This will eventually cause our code not to compile so I would remove the `[false, false]` case above and return `[]` here.

jclark (commented):
This is looking good.

manuranga:
@jclark I believe this is ready to be merged. I addressed both of the concerns I had this morning.
Conclusions are:
1) Assignment invalidation *do* need to consider the unnarrowed registry
2) Can remove narrowings from the `Environment`. `basis` information only needs to flow inside-out. This is because each `CondNarrowInsn` only narrows incrementally, so it only needs to provide the basis for narrowing since the last binding.

compiler/modules/front/codeGen.bal
```diff
@@ -28,9 +28,16 @@ type Binding record {|
...
+type StmtNarrowing record {|
+    Binding binding;
+    t:SemType ifCompletesNormally;
+    bir:Result basis;
+|};
+
 type StmtEffect record {|
     bir:BasicBlock? block;
     Binding? bindings = ();
+    StmtNarrowing[] narrowings = [];
```
jclark, line 40: I think it should work to store the basis in the binding, instead of doing this.

compiler/modules/front/codeGen.bal
```diff
@@ -318,12 +325,19 @@ function codeGenOnPanic(CodeGenContext cx, Position pos) {
...
 
-function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialEnv, s:StmtBlock block) returns CodeGenError|StmtEffect {
+// If block of stmts doesn't complete normally, will return empty narrowings and assignments. initialNarrowings array will be modified.
+function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialEnv, s:StmtBlock block, StmtNarrowing[] initialNarrowings = []) returns CodeGenError|StmtEffect {
     bir:BasicBlock? curBlock = bb;
     Environment env = environmentCopy(initialEnv);
     final int startRegister = cx.nextRegisterNumber();
-    foreach var stmt in block.stmts {
+    StmtNarrowing[] narrowings = initialNarrowings;
+    applyEffect(env, { block: bb, narrowings });
```
jclark: applyStmtEffect

jclark: Use initialNarrowings here

compiler/modules/front/codeGen.bal
```diff
@@ -318,12 +325,19 @@ function codeGenOnPanic(CodeGenContext cx, Position pos) {
     }
 }
 
-function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialEnv, s:StmtBlock block) returns CodeGenError|StmtEffect {
+// If block of stmts doesn't complete normally, will return empty narrowings and assignments. initialNarrowings array will be modified.
+function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialEnv, s:StmtBlock block, StmtNarrowing[] initialNarrowings = []) returns CodeGenError|StmtEffect {
     bir:BasicBlock? curBlock = bb;
     Environment env = environmentCopy(initialEnv);
     final int startRegister = cx.nextRegisterNumber();
```
jclark, line 332: Add comment that this makes the register number of the narrowed variables >= startRegister

compiler/modules/front/codeGen.bal
```diff
@@ -318,12 +325,19 @@ function codeGenOnPanic(CodeGenContext cx, Position pos) {
...
-    foreach var stmt in block.stmts {
+    StmtNarrowing[] narrowings = initialNarrowings;
+    applyEffect(env, { block: bb, narrowings });
+    codeGenNarrowings(cx, bb, env, narrowings, block.startPos);
+    int lastStmtIndex = block.stmts.length() - 1;
+    int stmtIndex = 0;
+    while stmtIndex <= lastStmtIndex {
```
jclark: go back to foreach and incremenet stmtIndex

compiler/modules/front/codeGen.bal
```diff
@@ -769,24 +806,96 @@ function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
...
     }
 }
 
-function codeGenIfElseNarrowing(CodeGenContext cx, bir:BasicBlock bb, Environment env, Narrowing narrowing, boolean condition, Position pos) returns Environment {
-    boolean insnResult = condition == !narrowing.negated;
-    // JBUG #33303 without parentheses this gets a parse error
-    t:SemType narrowedType = insnResult ? (narrowing.ifTrue) : narrowing.ifFalse;
-    if narrowedType === t:NEVER {
-        panic err:impossible("narrowed to never type");
+function codeGenNarrowings(CodeGenContext cx, bir:BasicBlock bb, Environment env, StmtNarrowing[] narrowings, Position pos) {
```
jclark: This function can be improved or merged

compiler/modules/front/codeGen.bal
```diff
@@ -650,7 +681,11 @@ function codeGenMatchStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environm
             if basis != () {
                 // Will need readOnlyIntersect when we have proper match patterns
                 t:SemType narrowedType = t:intersect(matchedType, clauseLooksLike[clauseIndex]);
-                clauseEnv = codeGenNarrowing(cx, stmtBlock, env, binding, narrowedType, basis, clause.opPos);
+                // XXX: match shouldn't code gen narrowings, it should let codeGenStmts handle it
```
jclark: Let's do this

compiler/modules/front/codeGen.bal
```diff
@@ -318,12 +325,19 @@ function codeGenOnPanic(CodeGenContext cx, Position pos) {
     }
 }
 
-function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialEnv, s:StmtBlock block) returns CodeGenError|StmtEffect {
+// If block of stmts doesn't complete normally, will return empty narrowings and assignments. initialNarrowings array will be modified.
+function codeGenStmts(CodeGenContext cx, bir:BasicBlock bb, Environment initialEnv, s:StmtBlock block, StmtNarrowing[] initialNarrowings = []) returns CodeGenError|StmtEffect {
```
jclark: rename to codeGenStmtBlock

jclark (changes requested):

compiler/modules/front/codeGen.bal
manuranga: This is no longer the case since I am now using `unnarrowBinding`, But I still believe this is the right place to process `initialNarrowings` since new bindings should live inside block scope. I added a comment saying that.

jclark: Does it still modify `initialNarrowings`? If so, why?

compiler/modules/front/codeGen.bal
manuranga: Now that I moved `codeGenNarrowings` out of this, I think a better name would be `updateAssignments`

compiler/modules/front/codeGen.bal
manuranga: Yes, (but not here, in the loop) `initialNarrowings` grows into the narrowings of the block, I could clone it to keep things clean.

jclark: Seems a bit confusing to call it initialNarrowings and then mutate it into something that isn't the initialNarrowings.

manuranga: Ya that is not nice. Easy fix is to clone. Maybe a better solution is to accept `ExprNarrowing initialNarrowing` and create the array here. But then I'd have to accept a boolean as well. (Or I can get rid of the boolean, have a single `Narrowing` type that can be used for stmt as well, where `type ExprNarrowing  [Narrowing, Narrowing];`)

manuranga: I'll get around this for now by making it `StmtNarrowing? initialNarrowing` and cross this bridge when we are doing `&&` and `||`.

jclark:
Is this ready?

manuranga:
Yes, for some reason GitHub review request button is not working.


https://user-images.githubusercontent.com/1686124/148156717-ac8fd5b0-183d-4b38-8a94-b9e4cf2c6bbb.mov

compiler/modules/front/codeGen.bal
```diff
@@ -769,40 +804,106 @@ function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
...
+                break;
+            }
+        }
+        if !added {
+            dest.push(s);
+        }
+    }
+}
+
+function optStmtNarrowingsFromExprNarrowing(ExprNarrowing? narrowing, boolean condition) returns StmtNarrowing? {
```
jclark: Just `stmtNarrowingFromExprNarrowing` (the ExprNarrowing has `?`)

compiler/modules/front/codeGen.bal
```diff
@@ -769,40 +804,106 @@ function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
...
+        t:SemType narrowedType = insnResult ? (narrowing.ifTrue) : narrowing.ifFalse;
+        if narrowedType === t:NEVER {
+            panic err:impossible("narrowed to never type");
+        }
+        bir:Result basis = { insn: narrowing.testInsn, result: insnResult };
+        return { basis, binding: narrowing.binding, ifCompletesNormally: narrowedType };
+    }
+}
+
+function codeGenNarrowings(CodeGenContext cx, bir:BasicBlock bb, Environment env, StmtNarrowing[] narrowings, Position pos) {
```
jclark: I don't think this should be called `codeGenX`. What it is doing is adding the narrowings to the env. Maybe `addNarrowings`.

jclark (changes requested):
Just a couple of naming tweaks...

manuranga:
@jclark Ready to review.

I tried the GitHub CLI this time, still not working.
```
> gh pr ready 788
! Pull request #788 is already "ready for review"
```

compiler/modules/front/codeGen.bal
```diff
@@ -769,40 +804,106 @@ function codeGenIfElseStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
...
+                break;
+            }
+        }
+        if !added {
+            dest.push(s);
+        }
+    }
+}
+
+function stmtNarrowingsFromExprNarrowing(ExprNarrowing? narrowing, boolean condition) returns StmtNarrowing? {
```
jclark: stmtNarrowing not stmtNarrowings.

jclark (changes requested):

compiler/modules/front/codeGen.bal
manuranga: Fixed.

PR #793, opened 2021-12-20
Fix float/int conversion bug

### LINKED ISSUE #792: Compiler crash in float type cast
```import ballerina/io;
public function main() {
    int|float x = 1;
    float f = <float>x;
    io:println(f);
}
```
For this example compiler crashes with,
```
error: store type mismatch: i8 addrspace(1)*, double*
	at wso2.nballerina.comm.err.0:illegalArgument(err.bal:36)
	   wso2.nballerina.print:store(llvm.bal:1018)
	   wso2.nballerina.nback.0:buildConvertToFloat(number.bal:200)
	   wso2.nballerina.nback.0:buildBasicBlock(basic.bal:82)
	   wso2.nballerina.nback.0:buildFunctionBody(module.bal:113)
	   wso2.nballerina.nback.0:buildModule(module.bal:60)
	   wso2.nballerina.0.CompileContext:buildModule(compile.bal:41)
	   wso2.nballerina.0:processModule(compile.bal:91)
	   wso2.nballerina.0:compileBalFile(compile.bal:76)
	   wso2.nballerina.0:main(main.bal:62)

### PR DESCRIPTION
**Pull request by manuranga**
Fixes #792

### DIFF
--- a/compiler/modules/nback/number.bal
+++ b/compiler/modules/nback/number.bal
@@ -1,6 +1,5 @@
 
 import wso2/nballerina.bir;
-import wso2/nballerina.types as t;
 import wso2/nballerina.print.llvm;
 
 final RuntimeFunction floatToIntFunction = {
@@ -124,7 +123,6 @@ function buildConvertToInt(llvm:Builder builder, Scaffold scaffold, bir:ConvertT
     }
     // convert to int form tagged pointer
 
-    t:SemType semType = insn.operand.semType;
     llvm:PointerValue tagged = <llvm:PointerValue>val;
     llvm:BasicBlock joinBlock = scaffold.addBasicBlock();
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: compiler/modules/nback/number.bal [hunk 1]
```diff
@@ -138,8 +137,11 @@ function buildConvertToInt(llvm:Builder builder, Scaffold scaffold, bir:ConvertT
     buildConvertFloatToInt(builder, scaffold, buildUntagFloat(builder, scaffold, tagged), insn);
     builder.br(joinBlock);
 
-    builder.positionAtEnd(<llvm:BasicBlock>noFloatBlock);
-    if !t:isSubtypeSimple(semType, t:FLOAT) {
+    builder.positionAtEnd(noFloatBlock);
+    if t:isSubtypeSimple(insn.result.semType, t:INT) {
```
jclark: This doesn't look good to me.  We should be looking at the Repr to decide whether to convert the representation not the SemType.  There are a bunch of functions already that deal with converting between tagged/untagged representations.  If there aren't ones that work for this case, we should add them.

jclark (changes requested):

compiler/modules/nback/number.bal
jclark: But this is going to be too complicated to inline especially with decimal. Better to rewrite using a runtime function.

jclark (changes requested):

compiler/modules/nback/number.bal
manuranga: May I add a rt function in this PR?

PR #797, opened 2021-12-21
Implement relational operators for decimal

### LINKED ISSUE #780: Implement relational operators for `decimal`
Make `< > <= >=` work for `decimal`.

They also need to work for `decimal[]` and `decimal?` and other ordered types involving decimal.

ushirask:
Resolves #780 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/nback/compare.bal
```diff
@@ -47,6 +47,15 @@ final RuntimeFunction booleanCompareFunction = {
     attrs: ["readonly"]
 };
 
+final RuntimeFunction decimalCompareFunction = {
```
jclark, line 59: The _cmp and _compare functions aren't the same. _cmp returns <0, 0, >0 (like strcmp).  The _compare functions return a different set of values that allow for unordered.

The name of the RuntimeFunction variable should always be the same as the runtime function name in camel case.

jclark (changes requested):

compiler/modules/nback/compare.bal
```diff
@@ -137,6 +156,9 @@ function buildCompare(llvm:Builder builder, Scaffold scaffold, bir:CompareInsn i
                 if isOperandIntSubtypeArray(tc, lhs) && isOperandIntSubtypeArray(tc, rhs) {
                     buildCompareSpecializedIntList(builder, scaffold, insn, lhsValue, rhsValue);
                 }
+                else if isOperandDecimalSubtypeArray(tc, lhs) && isOperandDecimalSubtypeArray(tc, rhs) {
```
jclark: The decimal comparison should work like string: do a buildCompareDecimal like buildCompareString. Use decimal_cmp. We don't need a specialized version of compare for arrays of decimals.

jclark (changes requested):

compiler/modules/nback/compare.bal
ushirask: If we replicate what we do with string here, similar to string, in a case like 
```
decimal[] x = [1.5d];
decimal[] y = [3d];
io:println(x < y); // @output true
io:println(x > y); // @output false
```
we would be using `array_generic_compare` right? In such a case `array_generic_compare` doesn't seem to give the correct result with decimal values.

PR #798, opened 2021-12-21
Implement nil lifting for operators with numeric operands and a numeric result

### LINKED ISSUE #733: Implement nil lifting for operators with numeric operands and a numeric result
As described here: https://github.com/ballerina-platform/ballerina-spec/issues/191

heshanpadmasiri:
## Purpose
Implement nil lifting for operators with numeric operands and a numeric result
Resolves #733

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/front/codeGen.bal
```diff
@@ -1118,8 +1125,12 @@ function codeGenExpr(CodeGenContext cx, bir:BasicBlock bb, Environment env, s:Ex
         var { opPos: pos, bitwiseOp: op, left, right } => {
             var { result: l, block: block1} = check codeGenExprForInt(cx, bb, env, left);
             var { result: r, block: nextBlock } = check codeGenExprForInt(cx, block1, env, right);
+            bir:Operand lhs;
+            bir:Operand rhs;
+            bir:BasicBlock? nilBlock;
+            { lhs, rhs, nextBlock, nilBlock } = codeGenBinaryNilLift(cx, l, r, nextBlock, pos);
```
jclark: It's not really right to use codeGenExprForInt if the result might be `()` (you will get a misleading error message).  The codeGenNilLift should be taking a list of expressions not operands. Then you should check the type of the operand is int fter doing the nil lifting.

compiler/modules/front/codeGen.bal
```diff
@@ -1216,6 +1227,98 @@ function codeGenExpr(CodeGenContext cx, bir:BasicBlock bb, Environment env, s:Ex
     panic err:impossible("unrecognized expression type in code gen: " +  s:exprToString(expr));
 }
 
+type LiftResult record {|
```
jclark: The names here are not consistent. You are sometimes using `NilLift`, sometimes just `Lift` and sometimes just `Nil`.

compiler/modules/front/codeGen.bal
```diff
@@ -1216,6 +1227,98 @@ function codeGenExpr(CodeGenContext cx, bir:BasicBlock bb, Environment env, s:Ex
...
+    bir:Operand[] operands;
+    bir:BasicBlock nextBlock;
+    bir:BasicBlock? nilBlock;
```
jclark: nextBlock and nilBlock are not a good pair because `next` and `nil` do not have the same relationship to `Block`: `ifNilBlock` better.

jclark (changes requested):

compiler/modules/front/codeGen.bal

compiler/modules/front/codeGen.bal

compiler/modules/front/codeGen.bal

compiler/modules/front/codeGen.bal
```diff
@@ -1116,10 +1118,12 @@ function codeGenExpr(CodeGenContext cx, bir:BasicBlock bb, Environment env, s:Ex
             return codeGenCheckingExpr(cx, bb, env, checkingKeyword, operand, pos);
         }
         var { opPos: pos, bitwiseOp: op, left, right } => {
-            var { result: l, block: block1} = check codeGenExprForInt(cx, bb, env, left);
-            var { result: r, block: nextBlock } = check codeGenExprForInt(cx, block1, env, right);
+            var { lhs, rhs, nextBlock, ifNilBlock } = check codeGenBinaryNilLift(cx, env, left, right, bb, pos);
             // We evaluate the operands here, so we can reuse the function for compound assignment.
```
jclark: The comment is no longer applicable.

compiler/modules/front/codeGen.bal
```diff
@@ -1090,15 +1089,18 @@ function codeGenExpr(CodeGenContext cx, bir:BasicBlock bb, Environment env, s:Ex
...
         }
         // Bitwise complement
         { opPos: var pos, op: "~",  operand: var o } => {
-            var { result: operand, block: nextBlock } = check codeGenExprForInt(cx, bb, env, o);
+            var { operand, nextBlock, ifNilBlock } = check codeGenUnaryNilLift(cx, env, o, bb, pos);
             bir:Register result = cx.createTmpRegister(t:INT, pos);
-            bir:IntBitwiseBinaryInsn insn = { op: "^", pos, operands: [-1, operand], result };
-            nextBlock.insns.push(insn);
-            return { result, block: nextBlock };
+            if operand is bir:IntOperand {
```
jclark: Should have a function that takes an Operand and converts to an IntOperand|error.

jclark (changes requested):

compiler/modules/front/codeGen.bal

compiler/modules/front/codeGen.bal

compiler/modules/front/codeGen.bal
```diff
@@ -1216,6 +1215,105 @@ function codeGenExpr(CodeGenContext cx, bir:BasicBlock bb, Environment env, s:Ex
     panic err:impossible("unrecognized expression type in code gen: " +  s:exprToString(expr));
 }
 
+function intOperand(CodeGenContext cx, bir:Operand operand, Position pos) returns bir:IntOperand|CodeGenError {
+    if operand is bir:IntOperand {
+        return operand;
+    }
+    return cx.semanticErr("invalid operand type", pos);
```
jclark: Make the error more specific: we need an int.

compiler/modules/front/codeGen.bal
```diff
@@ -1116,10 +1115,10 @@ function codeGenExpr(CodeGenContext cx, bir:BasicBlock bb, Environment env, s:Ex
             return codeGenCheckingExpr(cx, bb, env, checkingKeyword, operand, pos);
         }
         var { opPos: pos, bitwiseOp: op, left, right } => {
-            var { result: l, block: block1} = check codeGenExprForInt(cx, bb, env, left);
-            var { result: r, block: nextBlock } = check codeGenExprForInt(cx, block1, env, right);
-            // We evaluate the operands here, so we can reuse the function for compound assignment.
-            return codeGenBitwiseBinaryExpr(cx, nextBlock, op, pos, l, r);
+            var { lhs, rhs, nextBlock, ifNilBlock } = check codeGenBinaryNilLift(cx, env, left, right, bb, pos);
+            bir:IntOperand l = check intOperand(cx, lhs, pos);
```
jclark: Using the opPos here isn't going to give a very good message. Maybe use Range for left/right expression.

jclark (changes requested):

compiler/modules/front/codeGen.bal

compiler/modules/front/codeGen.bal

compiler/testSuite/12-nillift/1-v.bal
```diff
@@ -0,0 +1,41 @@
...
+    io:println(v5); // @output 2
+
+    int? v6 = -a;
+    io:println(v6); // @output -5
+    io:println(-c); // @output -1
+
+    int d = 13;
+    int? v7 = a + d;
+    io:println(v7); // @output 18
+    io:println(a+b+c+d); // @output 25
```
manuranga: minor: let's format with spaces.

heshanpadmasiri: fixed

compiler/modules/front/codeGen.bal
```diff
@@ -1216,6 +1215,109 @@ function codeGenExpr(CodeGenContext cx, bir:BasicBlock bb, Environment env, s:Ex
...
+function codeGenNilLift(CodeGenContext cx, Environment env, s:Expr[] operands, bir:BasicBlock bb, Position pos) returns NilLiftResult|CodeGenError {
+    bir:BasicBlock? ifNilBlock = ();
+    bir:BasicBlock nextBlock;
+    bir:BasicBlock currentBlock = bb;
+    bir:Operand[] newOperands = [];
+    foreach int i in 0 ..< operands.length() {
+        s:Expr operandExpr = operands[i];
+        var { result: operand, block } = check codeGenExpr(cx, currentBlock, env, operandExpr);
+        currentBlock = block;
+        if operand is bir:Register && t:containsNil(operand.semType) && t:singleNumericType(operand.semType) != () {
```
jclark: What is the call to singleNumericType doing here?

heshanpadmasiri: I added it to make sure only number types will be lifted. But removing doesn't seems to have an effect, so that check is removed.

jclark (changes requested):
Please have a test that checks that all operands are evaluated even if one of them is nil (i.e. it does not short-circuit like `&&` or `||`).

jclark:
> I added it to make sure only number types will be lifted. But removing doesn't seems to have an effect, so that check is removed. 

Please read the spec https://ballerina.io/spec/lang/master/#nil_lifting

It is not restricted to numbers: result of `string? + string?` is `string?`. Please have a test for that.

jclark:
Make sure there is a test for every point mentioned here:

https://github.com/ballerina-platform/ballerina-spec/issues/191#issuecomment-997324452

heshanpadmasiri:
added additional tests and fixed compound assignment not causing an error

compiler/modules/front/codeGen.bal
```diff
@@ -917,6 +918,7 @@ function codeGenAssignToMember(CodeGenContext cx, bir:BasicBlock startBlock, Env
 }
 
 function codeGenCompoundAssignStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environment env, s:CompoundAssignStmt stmt) returns CodeGenError|StmtEffect {
+    cx.skipNilLiftig = true;
```
jclark: This is not OK. It looks like you are using this to deal with not doing nil lifting in compound assignment.

The way to handle this is to have a clean layering between the functions that do nil lifting and the functions that do the underlying operation, and then use the latter when doing compound assignment.

You may also need to refactor the compound assignment to not use codeGenExpr for the operator in the compound assignement, but instead use a lower-level function.

heshanpadmasiri: fixed.

compiler/modules/front/codeGen.bal
```diff
@@ -1229,6 +1245,108 @@ function codeGenExpr(CodeGenContext cx, bir:BasicBlock bb, Environment env, s:Ex
...
 }
 
+function intOperand(CodeGenContext cx, bir:Operand operand, Range subExprRange) returns bir:IntOperand|CodeGenError {
+    if operand is bir:IntOperand {
+        return operand;
+    }
+    return cx.semanticErr("expected an int operand", subExprRange);
+}
+
+function exprRange(s:Expr expr) returns Range {
```
jclark: There's already an `s:range` function to do this.

compiler/modules/front/codeGen.bal
```diff
@@ -1229,6 +1245,108 @@ function codeGenExpr(CodeGenContext cx, bir:BasicBlock bb, Environment env, s:Ex
     panic err:impossible("unrecognized expression type in code gen: " +  s:exprToString(expr));
 }
 
+function intOperand(CodeGenContext cx, bir:Operand operand, Range subExprRange) returns bir:IntOperand|CodeGenError {
```
jclark: Put this function down the bottom. Make the argument be `Position|Range pos`.

jclark: Use this function in codeGenExprForInt.

jclark: Actually better to pass in Expr rather than Range.  We don't want to create the Range in the normal case where there's no error.

compiler/modules/front/codeGen.bal
```diff
@@ -1229,6 +1245,108 @@ function codeGenExpr(CodeGenContext cx, bir:BasicBlock bb, Environment env, s:Ex
...
+    return { startPos: expr.startPos, endPos: expr.endPos };
+}
+
+function codeGenNilLiftResult(CodeGenContext cx, ExprEffect nonNilEffect, bir:BasicBlock? ifNilBlock, Position pos) returns ExprEffect {
+    if ifNilBlock == () {
+        return nonNilEffect;
+    }
+    else {
+        bir:BasicBlock block = cx.createBasicBlock();
+        bir:Register nonNilResult = <bir:Register> nonNilEffect.result;
```
jclark: How do you know this cast cannot fail?

jclark (changes requested):

jclark:
Overall this is looking good.

compiler/modules/front/codeGen.bal

compiler/modules/front/codeGen.bal

compiler/modules/front/codeGen.bal
heshanpadmasiri: `nonNilEffect` is the `ExprEffect` we are getting from the non lifted operation (such as `codeGenArithmeticBinaryExpr`) and they always create a temporary register to store their result (which we don't codeGen to be a constant value).

jclark (changes requested):
> `nonNilEffect` is the `ExprEffect` we are getting from the non lifted operation (such as `codeGenArithmeticBinaryExpr`) and they always create a temporary register to store their result (which we don't codeGen to be a constant value).  

This used to be true, but is no longer true because of #804

heshanpadmasiri:
> > `nonNilEffect` is the `ExprEffect` we are getting from the non lifted operation (such as `codeGenArithmeticBinaryExpr`) and they always create a temporary register to store their result (which we don't codeGen to be a constant value).
> 
> This used to be true, but is no longer true because of #804

fixed

compiler/modules/front/codeGen.bal
```diff
@@ -868,6 +868,12 @@ function codeGenAssignStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environ
 }
 
 function codeGenAssignToVar(CodeGenContext cx, bir:BasicBlock startBlock, Environment env, string varName, s:Expr expr, Position pos) returns CodeGenError|StmtEffect {
+    var [unnarrowedReg, assignments] = check assignmentRegister(cx, env, varName, pos);
+    bir:BasicBlock nextBlock = check codeGenAssign(cx, env, startBlock, unnarrowedReg, expr, unnarrowedReg.semType, pos);
+    return { block: nextBlock, assignments };
+}
+
+function assignmentRegister(CodeGenContext cx, Environment env, string varName, Position pos) returns CodeGenError|[bir:Register, int[]] {
```
jclark: I would call this something like `lookupVarRefForAssign`

compiler/modules/front/codeGen.bal
```diff
@@ -983,16 +988,28 @@ function codeGenCompoundAssignToVar(CodeGenContext cx,
                                     s:Expr rexpr,
                                     s:BinaryArithmeticOp|s:BinaryBitwiseOp op,
                                     Position pos) returns CodeGenError|StmtEffect {
-    s:Expr expr;
-    Position startPos = rexpr.startPos;
-    Position endPos = rexpr.endPos;
+    var [unnarrowedReg, assignments] = check assignmentRegister(cx, env, lValue.name, pos);
```
jclark: I would expect this function to look more like the code for codeGenCompoundAssignToListMember/MappingMember.

jclark (changes requested):

compiler/modules/front/codeGen.bal

compiler/modules/front/codeGen.bal

compiler/modules/front/codeGen.bal
```diff
@@ -983,16 +988,18 @@ function codeGenCompoundAssignToVar(CodeGenContext cx,
                                     s:Expr rexpr,
                                     s:BinaryArithmeticOp|s:BinaryBitwiseOp op,
                                     Position pos) returns CodeGenError|StmtEffect {
-    s:Expr expr;
-    Position startPos = rexpr.startPos;
-    Position endPos = rexpr.endPos;
-    if op is s:BinaryArithmeticOp {
-        expr = { startPos, endPos, opPos: pos, arithmeticOp: op, left: lValue, right: rexpr };
+    var { block: nextBlock, result: lhs } = check codeGenExpr(cx, startBlock, env, lValue);
```
jclark: Don't understand what you're doing here.  lookupVarRefForAssign will give you the register.  So why can't you use that?

The codeGenExprForInt in the codeGenCompoundAssignToMember is for the index expression: there's nothing like that here.

jclark (changes requested):

compiler/modules/front/codeGen.bal

compiler/modules/front/codeGen.bal
```diff
@@ -1099,26 +1095,25 @@ function codeGenExprForString(CodeGenContext cx, bir:BasicBlock bb, Environment
...
-            return codeGenArithmeticBinaryExpr(cx, nextBlock, op, pos, l, r);
+            var { lhs, rhs, nextBlock, ifNilBlock } = check codeGenBinaryNilLift(cx, env, left, right, bb, pos);
+            return codeGenNilLiftResult(cx, check codeGenArithmeticBinaryExpr(cx, nextBlock, op, pos, lhs, rhs), ifNilBlock, pos);
         }
         // Negation
         { opPos: var pos, op: "-",  operand: var o } => {
-            var { result: operand, block: nextBlock } = check codeGenExpr(cx, bb, env, o);
-            return codeGenNegateExpr(cx, nextBlock, pos, operand);
+            var { operand, nextBlock, ifNilBlock } = check codeGenUnaryNilLift(cx, env, o, bb, pos);
+            return codeGenNegateExpr(cx, nextBlock, ifNilBlock, pos, operand);
```
jclark: For uniformity, codeGenNilLiftResult should be here not in codeGenNegateExpr.

jclark (changes requested):

compiler/modules/front/codeGen.bal

PR #802, opened 2021-12-22
Support decimal to int conversion

### LINKED ISSUE #778: Implement conversion from decimal to int
This involves making the ConvertToInt instruction work.

KavinduZoysa:
## Purpose
Fixes #778

### DIFF
--- a/compiler/modules/nback/number.bal
+++ b/compiler/modules/nback/number.bal
@@ -3,10 +3,12 @@ import wso2/nballerina.bir;
 import wso2/nballerina.types as t;
 import wso2/nballerina.print.llvm;
 
+final llvm:StructType LLVM_INT_WITH_OVERFLOW = llvm:structType(["i64", "i1"]);
+
 final RuntimeFunction floatToIntFunction = {
     name: "float_to_int",
     ty: {
-        returnType: llvm:structType(["i64", "i1"]),
+        returnType: LLVM_INT_WITH_OVERFLOW,
         paramTypes: ["double"]
     },
     attrs: ["nounwind", "readnone", "speculatable", "willreturn"]
@@ -25,6 +27,24 @@ final RuntimeFunction convertToFloatFunction = {
     name: "convert_to_float",
     ty: {
         returnType: LLVM_TAGGED_PTR,
+        paramTypes: [LLVM_TAGGED_PTR]
+    },
+    attrs: ["readonly"]
+};
+
+final RuntimeFunction decimalToIntFunction = {
+    name: "decimal_to_int",
+    ty: {
+        returnType: LLVM_INT_WITH_OVERFLOW,
+        paramTypes: [LLVM_TAGGED_PTR]
+    },
+    attrs: ["readonly"]
+};
+
+final RuntimeFunction convertToIntFunction = {
+    name: "convert_to_int",
+    ty: {
+        returnType: llvm:structType([LLVM_TAGGED_PTR, "i1"]),
         paramTypes: [LLVM_TAGGED_PTR]
     },
     attrs: ["readonly"]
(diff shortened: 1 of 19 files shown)


compiler/modules/nback/number.bal
```diff
@@ -156,41 +175,32 @@ function buildConvertToInt(llvm:Builder builder, Scaffold scaffold, bir:ConvertT
...
-    buildStoreTagged(builder, scaffold, tagged, insn.result);
-    builder.br(joinBlock);
-    builder.positionAtEnd(joinBlock);
+function buildConvertDecimalToInt(llvm:Builder builder, Scaffold scaffold, llvm:Value decimalVal, bir:ConvertToIntInsn insn) {
+    buildStoreInt(builder, scaffold, convertedInt(builder, scaffold, decimalVal, insn, decimalToIntFunction), insn.result);
 }
 
-function buildConvertFloatToInt(llvm:Builder builder, Scaffold scaffold, llvm:Value floatVal, bir:ConvertToIntInsn insn) {
-    llvm:Value resultWithErr = <llvm:Value>builder.call(scaffold.getRuntimeFunctionDecl(floatToIntFunction), [floatVal]);
+function convertedInt(llvm:Builder builder, Scaffold scaffold, llvm:Value val, bir:ConvertToIntInsn insn, RuntimeFunction func) returns llvm:Value {
```
jclark: Move the call to the RuntimeFunction out of this function. The job of this function then becomes to check the overflow, where the value can be either a tagged or int.  Call it something like buildCheckOverflow. It takes the resultWithErr as an argument, and returns the first member.

compiler/modules/nback/number.bal
```diff
@@ -20,6 +21,24 @@ final RuntimeFunction decimalNegFunction = {
     attrs: ["readonly"]
 };
 
+final RuntimeFunction decimalToIntFunction = {
+    name: "decimal_to_int",
+    ty: {
+        returnType: llvm:structType(["i64", "i1"]),
```
jclark: This struct is used here and in floatToInt, so construct it once and stick it in a final module-level variable.

jclark (changes requested):

compiler/modules/nback/number.bal

compiler/modules/nback/number.bal

runtime/convert.c
```diff
@@ -14,3 +15,35 @@ TaggedPtr _bal_convert_to_float(TaggedPtr tp) {
...
 }
+
+TaggedIntWithOverflow _bal_convert_to_int(TaggedPtr tp) {
+    int64_t val;
+    bool overflowed;
+    TaggedIntWithOverflow res;
+    switch (getTag(tp) & UT_MASK) {
+        case TAG_FLOAT:
+            {
+                struct FloatToIntResult intRes = _bal_float_to_int(taggedToFloat(tp));
```
jclark: Get rid of `struct FloatToIntResult` and use `IntWithOverflow` instead. (`FloatToIntResult` should be using `int64_t` not `uint64_t`).

Get rid of `overflowed` variable.

`TaggedIntWithOverflow` is misnamed because the tagged value isn't necessarily an `int`. Fix the name of the struct and the name of the member accordingly.

jclark (changes requested):

runtime/convert.c
KavinduZoysa: > Get rid of `struct FloatToIntResult` and use `IntWithOverflow` instead. (`FloatToIntResult` should be using `int64_t` not `uint64_t`).
> 

@jclark, to be more clear about this suggestion, do you expect to change all places where `FloatToIntResult` is used into `IntWithOverflow`?

jclark: Yes - we don't need both of them.

jclark: You can do this in a separate clean-up PR if you prefer.

KavinduZoysa: @jclark, I will do it in a separate PR. 

What I have to do is,
1. Create struct like
```
typedef struct {
    TaggedPtr tp;
    bool overflowed;
} TaggedPtrWithOverflow;
```

2. Change `_bal_float_to_int` and `_bal_decimal_to_int` return above struct.
3. In number.bal if repr of reg is BASE_REPR_TAGGED  to store `tp`, otherwise store untagged value of `tp`.

jclark: No, that's not right. First member of FloatToIntResult is not a TaggedPtr.

```
struct FloatToIntResult {
    uint64_t value;
    // LLVM type is i1, no equalent in C. i1 gets lowered to 8-bits.
    // So using bool (which is also 8-bit) works.
    bool overflowed;
};
```

You don't need to change _bal_decimal_with_int. You just replace `struct FloatToIntResult` with `IntWithOverflow`.

compiler/modules/nback/number.bal
```diff
@@ -3,10 +3,12 @@ import wso2/nballerina.bir;
 import wso2/nballerina.types as t;
 import wso2/nballerina.print.llvm;
 
+final llvm:StructType LLVM_STRUCT_WITH_ERROR = llvm:structType(["i64", "i1"]);
```
jclark: Call this LLVM_INT_WITH_OVERFLOW.

runtime/balrt.h
```diff
@@ -294,6 +294,11 @@ typedef struct {
     bool overflow;
 } IntWithOverflow;
 
+typedef struct {
+    TaggedPtr taggedInt;
```
jclark: Call this `ptr`.

runtime/balrt.h
```diff
@@ -294,6 +294,11 @@ typedef struct {
...
+    bool overflowed;
+} TaggedIntWithOverflow;
```
jclark: Call this `TaggedWithOverflow`.

jclark (changes requested):

runtime/balrt.h

runtime/balrt.h

compiler/modules/nback/number.bal

runtime/convert.c
KavinduZoysa: Fixed in #803

runtime/convert.c
```diff
@@ -14,3 +15,33 @@ TaggedPtr _bal_convert_to_float(TaggedPtr tp) {
...
     return floatToTagged(d);
 }
+
+TaggedWithOverflow _bal_convert_to_int(TaggedPtr tp) {
+    int64_t val;
+    TaggedWithOverflow res;
+    switch (getTag(tp) & UT_MASK) {
+        case TAG_FLOAT:
+            {
+                IntWithOverflow intRes = _bal_float_to_int(taggedToFloat(tp));
```
jclark: Declare intRes just before res. Then get rid of the braces in the cases. Assign res.overflow once. Get rid of val.

jclark (changes requested):

runtime/convert.c
KavinduZoysa: Changed

PR #803, opened 2021-12-23
Refactor _bal_float_to_int

KavinduZoysa:
## Purpose
Refactor _bal_float_to_int to return a value same as _bal_decimal_to_int


jclark:
I slightly prefer using `overflow` to `overflowed` for the field name.

KavinduZoysa:
> I slightly prefer using `overflow` to `overflowed` for the field name.

Changed

PR #826, opened 2022-01-04
fix SourceFile.lineContent crashes for position at end of line

### LINKED ISSUE #785: SourceFile.lineContent crashes for position at end of line
SourceFile:lineContent crashes in tokenEndCodePointIndex when the position is position following last character in file.

For example, if the line contains just a close brace and the position is 1. In this case, it should return `["}", "", ""]`.

(We probably shouldn't ever do this, but still better not to crash here.)

heshanpadmasiri:
## Purpose

- Fix SourceFile.lineContent crashes for position at end of line
- Provide meaningful error messages for out of range positions

Resolves #785

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/comm.diagnostic/diagnostic.bal
+++ b/compiler/modules/comm.diagnostic/diagnostic.bal
@@ -119,7 +119,11 @@ public function format(Diagnostic d) returns string[] {
     line += ": error: " + d.message;
     lines.push(line);
     Range|Position range = loc.range;
-    lines.push("".'join(...loc.file.lineContent(range)));
+    string[] content = loc.file.lineContent(range);
+    if content[1] == "" {
+        panic error(string `error position ${lc.toString()} is empty`);
+    }
+    lines.push("".'join(...content));
     lines.push(caretLine(loc.file, range));
     return lines;
 }
--- a/compiler/modules/comm.diagnostic/diagnostic.bal
+++ b/compiler/modules/comm.diagnostic/diagnostic.bal
@@ -141,7 +141,7 @@ function validatePosition(Position pos, File file) {
     if lineNum > file.maxLineNum() || lineNum < 1 {
         panic error(string `line number ${lineNum} is out of range in ${file.filename()}`);
     }
-    if columnNum >= file.maxColumnNum(lineNum) || columnNum < 0 {
+    if columnNum > file.maxColumnNum(lineNum) || columnNum < 0 {
         panic error(string `column number ${columnNum} is out of range for ${file.filename()}`);
     }
 }
--- a/compiler/modules/comm.diagnostic/diagnostic.bal
+++ b/compiler/modules/comm.diagnostic/diagnostic.bal
@@ -28,6 +28,8 @@ public type File readonly & object {
     public function lineColumn(Position pos) returns LineColumn;
     public function lineContent(Position|Range range) returns [string, string, string];
     public function qNameRange(Position startPos) returns Range;
+    public function maxLineNum() returns int;
+    public function maxColumnNum(int lineNum) returns int;
 };
 
 public type Location readonly & record {|
@@ -108,6 +110,7 @@ public function messageFormat(Template t) returns string {
 
 public function format(Diagnostic d) returns string[] {
     Location loc = d.location;
+    validateLocation(loc);
     string[] lines = [];
     string line = loc.file.filename();
     LineColumn lc = locationLineColumn(loc);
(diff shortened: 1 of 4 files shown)


compiler/modules/front.syntax/token.bal
```diff
@@ -569,6 +571,19 @@ public readonly class SourceFile {
     function scannedLine(int lineNumber) returns ScannedLine {
         return self.lines[lineNumber - 1];
     }
+
+    function validatePosition(Position pos) {
```
jclark: This is not the right way to solve this problem.  Please do something more like what I suggested in the issue.

jclark (changes requested):

compiler/modules/front.syntax/token.bal

jclark:
There seems to be a misunderstanding about what the issue is.

The issue is #785, which is "SourceFile.lineContent crashes for position at end of line".  The issue  is not "Provide meaningful error messages for out of range positions".

Fixing #785 means fixing lineContent so that e.g. if the line contains just a close brace and the position is 1, the lineContent should return ["}", "", ""].

The 2nd part of this is that we should not be constructing errors wiith locations where lineContent would return a 2nd string that is empty. So when d.format gets such a return value from lineContent, it should do so in a way that ensures we know that there's a problem and can fix it (that is related to the subject of this PR, but is a specific case).

jclark (changes requested):
Please also fix the PR title.

heshanpadmasiri:
> There seems to be a misunderstanding about what the issue is.
> 
> The issue is #785, which is "SourceFile.lineContent crashes for position at end of line". The issue is not "Provide meaningful error messages for out of range positions".
> 
> Fixing #785 means fixing lineContent so that e.g. if the line contains just a close brace and the position is 1, the lineContent should return ["}", "", ""].
> 
> The 2nd part of this is that we should not be constructing errors wiith locations where lineContent would return a 2nd string that is empty. So when d.format gets such a return value from lineContent, it should do so in a way that ensures we know that there's a problem and can fix it (that is related to the subject of this PR, but is a specific case).

fixed. (`lineContent` will no longer fail in such cases and d:format will panic with a meaningful error message in such cases (in addition to out of range positions)

compiler/modules/comm.diagnostic/diagnostic.bal
```diff
@@ -28,6 +28,8 @@ public type File readonly & object {
     public function lineColumn(Position pos) returns LineColumn;
     public function lineContent(Position|Range range) returns [string, string, string];
     public function qNameRange(Position startPos) returns Range;
+    public function maxLineNum() returns int;
```
jclark: This is not part of the issue.

compiler/modules/comm.diagnostic/diagnostic.bal
```diff
@@ -108,6 +110,7 @@ public function messageFormat(Template t) returns string {
 
 public function format(Diagnostic d) returns string[] {
     Location loc = d.location;
+    validateLocation(loc);
```
jclark: This is not part of the issue.

compiler/modules/comm.diagnostic/diagnostic.bal
```diff
@@ -116,11 +119,37 @@ public function format(Diagnostic d) returns string[] {
...
+    string[] content = loc.file.lineContent(range);
+    if content[1] == "" {
+        panic error(string `error range ${lc.toString()} is empty in ${loc.file.filename()}`);
+    }
+    lines.push("".'join(...content));
     lines.push(caretLine(loc.file, range));
     return lines;
 }
 
+function validateLocation(Location loc) {
```
jclark: This is not part of the issue.

jclark (changes requested):

compiler/modules/comm.diagnostic/diagnostic.bal

compiler/modules/comm.diagnostic/diagnostic.bal

compiler/modules/comm.diagnostic/diagnostic.bal

compiler/modules/front.syntax/token.bal
```diff
@@ -547,6 +547,9 @@ public readonly class SourceFile {
                 endColumnNum = lineContent.length();
             }
         }
+        else if startColumnNum == lineContent.length() {
```
jclark: I would have thought it would be cleaner to fix tokenEndCodePointIndex...

jclark (changes requested):

compiler/modules/front.syntax/token.bal

PR #827, opened 2022-01-04
Fix lengthy array iterations

### LINKED ISSUE #820: Semtype representation of fixed-length array grows linearly while performing `listInhabited`
**Description:**
Fixed length array representation expands the `initial` components to `N` when the fixed length is `N` during `listInhabited`.

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**Steps to reproduce:**

**Related Issues:**

rdhananjaya:
## Purpose
Fixes https://github.com/ballerina-platform/nballerina/issues/820

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/types/list.bal
+++ b/compiler/modules/types/list.bal
@@ -270,7 +270,11 @@ function listInhabited(Context cx, FixedLengthArray members, SemType rest, Conju
         // SemType d1 = diff(s[1], t[1]);
         // return !isEmpty(cx, d1) &&  tupleInhabited(cx, [s[0], d1], neg.rest);
         // We can generalize this to tuples of arbitrary length.
-        foreach int i in 0 ..< len {
+        int maxInitialLength = members.initial.length();
+        if maxInitialLength != 0 {
+            maxInitialLength = int:max(maxInitialLength, nt.members.initial.length());
+        }
+        foreach int i in 0 ..< maxInitialLength {
             SemType d = diff(fixedArrayGet(members, i), listMemberAt(nt.members, nt.rest, i));
             if !isEmpty(cx, d) {
                 FixedLengthArray s = fixedArrayShallowCopy(members);
@@ -278,7 +282,7 @@ function listInhabited(Context cx, FixedLengthArray members, SemType rest, Conju
                 if listInhabited(cx, s, rest, neg.next) {
                     return true;
                 }
-            }     
+            }
         }
         if !isEmpty(cx, diff(rest, nt.rest)) {
             return true;
--- a/compiler/testSuite/semtype/fixed-length-array-t.bal
+++ b/compiler/testSuite/semtype/fixed-length-array-t.bal
@@ -66,3 +66,13 @@ type Array7OfArray2OfInt5 Array2OfInt5[7];
 
 // @type Array7x2x5 = Array7OfArray2OfInt5
 type Array7x2x5 int[7][2][5];
+
+public const int MAX_VALUE = 9223372036854775807;
+public const int MAX_VALUE_M_1 = 9223372036854775806;
+
+// @type LargeArray < IntArray
+type LargeArray int[MAX_VALUE];
+
+// @type LargeArray2 < IntArray
+// @type LargeArray <> LargeArray2
+type LargeArray2 int[MAX_VALUE_M_1];


compiler/modules/types/list.bal
```diff
@@ -289,6 +291,18 @@ function listInhabited(Context cx, FixedLengthArray members, SemType rest, Conju
     }
 }
 
+function listMemberInhabited(Context cx, Conjunction neg, FixedLengthArray members, SemType rest, ListAtomicType nt, int i) returns boolean {
```
rdhananjaya: @jclark I'm having a hard time naming this function, this is just a temporary name.

compiler/modules/types/list.bal
```diff
@@ -270,15 +270,17 @@ function listInhabited(Context cx, FixedLengthArray members, SemType rest, Conju
...
         // return !isEmpty(cx, d1) &&  tupleInhabited(cx, [s[0], d1], neg.rest);
         // We can generalize this to tuples of arbitrary length.
-        foreach int i in 0 ..< len {
-            SemType d = diff(fixedArrayGet(members, i), listMemberAt(nt.members, nt.rest, i));
-            if !isEmpty(cx, d) {
-                FixedLengthArray s = fixedArrayShallowCopy(members);
-                fixedArraySet(s, i, d);
-                if listInhabited(cx, s, rest, neg.next) {
+        foreach int i in 0 ..< members.initial.length() {
+            if listMemberInhabited(cx, neg, members, rest, nt, i) {
```
jclark: Can we not combine these two loops? The first goes from 0 ..< members.initial.length() and the second goes from members.initial.length() ..< nt.members.initial.length()?

Why isn't it just a matter of changing `..< len` in the original code to `int:max(members.initial.length(), nt.members.initial.length())`

jclark (changes requested):

compiler/modules/types/list.bal
jclark: Not clear to me that we need this function.

compiler/modules/types/list.bal
rdhananjaya: Yes I missed that. Will merge the loops.

compiler/modules/types/list.bal
```diff
@@ -270,15 +270,17 @@ function listInhabited(Context cx, FixedLengthArray members, SemType rest, Conju
...
         // We can generalize this to tuples of arbitrary length.
-        foreach int i in 0 ..< len {
-            SemType d = diff(fixedArrayGet(members, i), listMemberAt(nt.members, nt.rest, i));
-            if !isEmpty(cx, d) {
-                FixedLengthArray s = fixedArrayShallowCopy(members);
-                fixedArraySet(s, i, d);
-                if listInhabited(cx, s, rest, neg.next) {
-                    return true;
+        if len != 0 {
+            foreach int i in 0 ..< int:max(members.initial.length(), nt.members.initial.length()) {
```
rdhananjaya: We can remove the `if` and bring that condition into the upper limit of the int-range, I'm not sure which way to go.

compiler/modules/types/list.bal
jclark: If you combine the loops, you don't need to factor out the listMemberInhabited function (which means you don't need to think of a good name for it). Also it will be easier to see how the logic is changed.

compiler/modules/types/list.bal
```diff
@@ -270,15 +270,17 @@ function listInhabited(Context cx, FixedLengthArray members, SemType rest, Conju
...
         // return !isEmpty(cx, d1) &&  tupleInhabited(cx, [s[0], d1], neg.rest);
         // We can generalize this to tuples of arbitrary length.
-        foreach int i in 0 ..< len {
-            SemType d = diff(fixedArrayGet(members, i), listMemberAt(nt.members, nt.rest, i));
-            if !isEmpty(cx, d) {
-                FixedLengthArray s = fixedArrayShallowCopy(members);
-                fixedArraySet(s, i, d);
-                if listInhabited(cx, s, rest, neg.next) {
-                    return true;
+        if len != 0 {
```
jclark: Are you sure you need this test?

I would just compute separately the upper limit (which might be 0) and then do the foreach up to that limit (which will do nothing if the upper limit is 0).

jclark:
> We can remove the `if` and bring that condition into the upper limit of the int-range, I'm not sure which way to go. 

We should do that if we need the test.

compiler/modules/types/list.bal
rdhananjaya: No, we don't need it (https://github.com/ballerina-platform/nballerina/pull/827/files#r778093687) I just don't have a name for the variable.
Would something like `distinctLen` work?

compiler/modules/types/list.bal
```diff
@@ -270,15 +270,16 @@ function listInhabited(Context cx, FixedLengthArray members, SemType rest, Conju
         // SemType d1 = diff(s[1], t[1]);
         // return !isEmpty(cx, d1) &&  tupleInhabited(cx, [s[0], d1], neg.rest);
         // We can generalize this to tuples of arbitrary length.
-        foreach int i in 0 ..< len {
+        int distinctMemberCount = members.initial.length() == 0 ? 0 : int:max(members.initial.length(), nt.members.initial.length());
```
jclark: You have two uses of members.initial.length() here.

I think I would call the variable `maxInitialLength` and write it like this:

```
int maxInitialLength = members.initial.length();
if maxInitialLength != 0 {
    maxInitialLength = int:max(maxInitialLength, nt.members.initial.length());
}
```

jclark (changes requested):

PR #831, opened 2022-01-05
Fix assignment to variables narrowed inside a loop

### LINKED ISSUE #825: Assignment to variables narrowed inside a loop is not supported
Assignment to variables narrowed outside loops shouldn't be supported (ballerina-platform/ballerina-spec#904). But the current implementation rejects even the assignments to variables defined outside but narrowed inside the loop.

```ballerina
import ballerina/io;

type IntList record {
   int i;
   IntList? next;
};

public function main() {
    IntList l = { i: 42, next: { i: 20, next: () } };
    int s = sum(l);
    io:println(s);
}

function sum(IntList? intList) returns int {
   IntList? temp = intList;
   int total = 0;
   while true {
      if temp is () {
        break;
      }
      else {
        total += temp.i;
        temp = temp.next;
      }
  }
  return total;
}
```

```
assignment to narrowed variable 'temp' in loop
   IntList? temp = intList;
            ^^^^
```

manuranga:
Fixes #825

### DIFF
--- a/compiler/modules/front/codeGen.bal
+++ b/compiler/modules/front/codeGen.bal
@@ -172,10 +172,6 @@ class CodeGenContext {
         return (<LoopContext>self.loopContext).onBreak;
     }
 
-    function loopStartRegister() returns int {
-        return (<LoopContext>self.loopContext).startRegister;
-    }
-
     function popLoopContext() {
         self.loopContext = (<LoopContext>self.loopContext).enclosing;
     }
@@ -427,6 +423,7 @@ function environmentCopy(Environment env) returns Environment {
 }
 
 function codeGenForeachStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environment env, s:ForeachStmt stmt) returns CodeGenError|StmtEffect {
+    Binding? startBindings = env.bindings;
     string varName = stmt.name;
     if lookup(varName, env) !== () {
         return cx.semanticErr(`duplicate declaration of ${varName}`, stmt.namePos);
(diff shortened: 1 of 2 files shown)


manuranga:
This fixes the bug, but still does not improve the error message location. For that we'll have to:
* Give an error during assignment. `LoopContext` needs to remember the `Binding`s at the beginning of the loop (not when `pushLoopContext` is currently called, but before the conditional is generated, since its valid to reassign to a var narrowed by the loop condition). We shouldn't give errors when the loop doesn't complete normally, it'll be difficult in this approach.

Or

* Add location information to the assignments array.

I'll try latter.

jclark:
I think we should add location info to the assignments array. But if we are considering making the assignments work more like bindings, then perhaps we should do this at the same time.

jclark:
We discussed creating an issue for reworking assignments array to be part of bindings, but I don't remember if there's an issue. If there isn't yet please create one.

compiler/modules/front/codeGen.bal
```diff
@@ -495,12 +497,19 @@ function codeGenWhileStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environm
     }
 }
 
-function validLoopAssignments(CodeGenContext cx, int[] assignments) returns CodeGenError? {
+function validLoopAssignments(CodeGenContext cx, Binding? binding, int[] assignments) returns CodeGenError? {
+    // pr-todo: I think swapping the loop order will be more efficient
     foreach int r in assignments {
-        if r < cx.loopStartRegister() {
```
jclark, line 532: I think this means we are not using loopStartRegister anywhere.

compiler/modules/front/codeGen.bal
```diff
@@ -495,12 +497,19 @@ function codeGenWhileStmt(CodeGenContext cx, bir:BasicBlock startBlock, Environm
...
-            return cx.semanticErr(`assignment to narrowed variable ${<string>cx.registerVarName(r)} in loop`, <Position>cx.registerPosition(r));
+        Binding? tem = binding;
+        while tem != () {
```
jclark: This feels like we don't have the right data structure any more. Should we be keeping more information about each assignment?

compiler/modules/front/codeGen.bal
manuranga: Yes, I removed it but have pushed to the wrong branch, I have accidentally opened this with my tmp branch.

compiler/modules/front/codeGen.bal
manuranga: Yes, since we'll have to keep the position anyway, we should also keep narrowed binding to which the assignment happens (currently we keep unnarrowed binding's reg number). Then we'll have original logic here, just with narrowed reg number instead unnarrowed reg number.

PR #835, opened 2022-01-05
Narrow types with while statement

### LINKED ISSUE #270: While statement should narrow types
Specified here https://github.com/ballerina-platform/ballerina-spec/issues/840

Follow on from #129

manuranga:
Fixes #270

### DIFF
--- a/compiler/testSuite/12-narrow/while02-v.bal
+++ b/compiler/testSuite/12-narrow/while02-v.bal
@@ -1,0 +1,21 @@
+import ballerina/io;
+
+public function main() {
+   int|boolean count = 3;
+   while count is int {
+      int i = count;
+      io:println(i); // @output 3
+                     // @output 2
+                     // @output 1
+      if count > 1 {
+         count = count - 1;
+      }
+      else {
+         count = false;
+      }
+   }
+   if count is int {
+      // Should be removed if #ballerina-spec/1019 is implemented.
+      io:println("unreached");
+   }
+}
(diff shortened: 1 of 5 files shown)


compiler/testSuite/12-narrow/while02-v.bal
```diff
@@ -0,0 +1,18 @@
...
+                     // @output 2
+                     // @output 1
+      if count > 1 {
+         count = count - 1;
+      }
+      else {
+         count = false;
+      }
+   }
+   // boolean _ = count;
```
manuranga: @jclark Shouldn't `count` be narrowed to `false` here ?

jclark (changes requested):
This is including #831 which I don't think is ready yet.

compiler/testSuite/12-narrow/while02-v.bal
jclark: I don't think the spec says that yet.

Please add a spec issue with this example and subject "Normal completion of while-stmt should be able to cause variable type narrowing".

If you don't have a `break`, then after the `while` completes normally the loop condition while be false.

manuranga: crated https://github.com/ballerina-platform/ballerina-spec/issues/1019

manuranga:
Please review #839 first.

jclark (changes requested):
Can you rebase this? Looks like it is straightforward given #839.

PR #838, opened 2022-01-06
Add parser support for map method

### LINKED ISSUE #823: Make parser support `map` as a method name
This is all we need to do right now to support https://github.com/ballerina-platform/ballerina-spec/issues/958

ushirask:
Fixes #823 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/front.syntax/tests/data/testParser.json
```diff
@@ -238,8 +239,6 @@
     ["V", "expr", "a(x).b.c[y]", "((a(x).b).c)[y]"],
     ["V", "expr", "a[b.c]"],
     ["E", "expr", "a.1", ""],
-    { "comment" : "may get allowed under  ballerina-spec#34" },
```
ushirask: @jclark Is there no need for `map` getting allowed for field access to be handled in the parser?

jclark: `a.map` is an error.

compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -340,7 +340,7 @@ function finishPrimaryExpr(Tokenizer tok, Expr expr, Position startPos) returns
         opPos = tok.currentStartPos();
         check tok.advance();
         Position qnamePos = tok.currentStartPos();
-        string name = check tok.expectIdentifier();
+        string name = check tok.expectMethodName();
```
jclark: This isn't really right, because we are expecting either
- an identifier, or
- a special method name followed by `(`

So I don't think introducing `expectMethodName` is the right approach.

jclark (changes requested):
expectMethodName is not quite the right function to handle this:

We want a function (not a method on Tokenizer) that will parse either

- method-name plus "(", or
- identifier

compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -340,6 +340,15 @@ function finishPrimaryExpr(Tokenizer tok, Expr expr, Position startPos) returns
         opPos = tok.currentStartPos();
         check tok.advance();
         Position qnamePos = tok.currentStartPos();
+        if tok.current() == "map" {
+            check tok.advance();
+            if tok.current() == "(" {
+                return finishPrimaryExpr(tok, check finishMethodCallExpr(tok, expr, "map", startPos, qnamePos, opPos), startPos);
```
jclark: Too much duplicated code.

compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -340,6 +340,15 @@ function finishPrimaryExpr(Tokenizer tok, Expr expr, Position startPos) returns
         opPos = tok.currentStartPos();
         check tok.advance();
         Position qnamePos = tok.currentStartPos();
+        if tok.current() == "map" {
```
jclark: Having a type `SpecialMethodName` as you previously did is good, since these names will grow.

jclark (changes requested):

compiler/modules/front.syntax/parseExpr.bal
ushirask: Fixed.

jclark (changes requested):
You need to deal with a method call stmt also:

```
x.map();
```

Even though it is not semantically valid, we should parse it correctly.

compiler/modules/front.syntax/tests/data/testParser.json
```diff
@@ -228,6 +228,8 @@
     ["V", "expr", "false.length()"],
     ["V", "expr", "-1.hexStr()", "-(1.hexStr())"],
     ["V", "expr", "-0xf.max()", "-(0xF.max())"],
+    ["V", "expr", "x.map()"],
```
jclark, line 231: Try some more complex expressions with multiple `.`s and `[]` also.

compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -355,6 +357,18 @@ function finishPrimaryExpr(Tokenizer tok, Expr expr, Position startPos) returns
     }
 }
 
+function parseIdOrSpecialMethodName(Tokenizer tok) returns string|err:Syntax {
```
jclark: Rename to `parseIdentifierOrMethodName`

jclark (changes requested):

compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -355,6 +357,18 @@ function finishPrimaryExpr(Tokenizer tok, Expr expr, Position startPos) returns
     }
 }
 
+function parseIdentifierOrMethodName(Tokenizer tok) returns string|err:Syntax {
+    Token? t = tok.current();
+    if t is SpecialMethodName {
+        check tok.advance();
+        if tok.current() == "(" {
```
manuranga, line 364: Can't we use `check tok.expect("(");`

ushirask: If we do so we won't be able to check the token in line 346 since the tokenizer will advance right?

jclark:
> Can't we use `check tok.expect("(");` 

No.

PR #844, opened 2022-01-11
Fix bugs with negative list semtypes

### LINKED ISSUE #828: `[T, T...]` is not a proper subtype of `[T...]`
**Description:**
`[T, T...]` is expected to be a proper subtype of `[T...]` but they are given as both subtypes of each other with `isSubtype`

ex: 
```
// @type T2 < T1
type T1 [int...];
type T2 [int, int...];
```
test case fails with `T2 is not a proper subtype of T1`

### LINKED ISSUE #842: Fixed length array subtyping bug when subtypes of array member type is involved
**Description:**
```ballerina
type IntArray int[];
type ISArray (int|string)[];

type Int4 int[4];
type Int1 int[1];
type Int14 Int4|Int1;
type NegInt14 (!Int14 & IntArray);

// @type I4A = IntArray
// @type I4A < ISArray
type I4A Int4|(!Int4 & IntArray);

// @type IA = IntArray
// @type IA < ISArray //// No type relationship is found here
type IA Int14|NegInt14;
```

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**Steps to reproduce:**

**Related Issues:**

rdhananjaya:
## Purpose
Fixes https://github.com/ballerina-platform/nballerina/issues/828#issuecomment-1006222680
Fixes https://github.com/ballerina-platform/nballerina/issues/842

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


jclark:
How does this relate to #842?

rdhananjaya:
> How does this relate to #842?

Found it while testing this PR. Thought maybe it should be fixed separately.

compiler/modules/types/list.bal
```diff
@@ -359,6 +358,17 @@ function fixedArrayShallowCopy(FixedLengthArray array) returns FixedLengthArray
     return { initial: shallowCopyTypes(array.initial), fixedLength: array.fixedLength };
 }
 
+function maxInitialLength(Context cx, Conjunction? con) returns int {
```
jclark: `listConjunctionMaxInitialLength`

compiler/modules/types/list.bal
```diff
@@ -241,17 +244,17 @@ function listInhabited(Context cx, FixedLengthArray members, SemType rest, Conju
         int negLen = nt.members.fixedLength;
         if len < negLen {
```
jclark, line 249: I am suspicious that `negLen` is left unchanged here but is changed below in the `foreach`. Are you sure that's right?

jclark (commented):
I need to think about this some more.

jclark:
> > How does this relate to #842?
> 
> Found it while testing this PR. Thought maybe it should be fixed separately.

Why? This PR is fixing up the logic of fixed length arrays, so if #842 is a fixed length array problem, we should fix as part of this.

compiler/modules/types/list.bal
```diff
@@ -241,17 +244,17 @@ function listInhabited(Context cx, FixedLengthArray members, SemType rest, Conju
...
+                return listInhabited(cx, members, rest, neg.next, maxInitialLen);
             }
             // For list shapes with length less than negLen,
             // this neg type is not relevant.
-            if listInhabited(cx, members, NEVER, neg.next) {
+            if listInhabited(cx, members, NEVER, neg.next, maxInitialLen) {
                 return true;
             }
-            foreach int i in len + 1 ..< negLen {
+            foreach int i in len + 1 ..< maxInitialLen {
```
jclark: I think this should be negLen

compiler/modules/types/list.bal
```diff
@@ -347,18 +349,31 @@ function fixedArraySet(FixedLengthArray members, int setIndex, SemType m) {
...
-    foreach int i in initCount ..< setIndex + 1 {
-        members.initial.push(lastMember);
+    if lastMember != m {
+        foreach int i in initCount ..< setIndex + 1 {
+            members.initial.push(lastMember);
+        }
+        members.initial[setIndex] = m;
     }
-    members.initial[setIndex] = m;
+    members.fixedLength = int:max(members.fixedLength, setIndex + 1);
```
jclark: Is it valid to call this function with setIndex >= members.fixedLength?

compiler/modules/types/list.bal
```diff
@@ -284,22 +287,21 @@ function listInhabited(Context cx, FixedLengthArray members, SemType rest, Conju
...
-                if listInhabited(cx, s, rest, neg.next) {
+                if listInhabited(cx, s, rest, neg.next, maxInitialLen) {
                     return true;
                 }
             }
         }
-        if !isEmpty(cx, diff(rest, nt.rest)) {
-            return true;
+        SemType rd = diff(rest, nt.rest);
+        if !isEmpty(cx, rd) {
```
jclark: Combine with `&&`

jclark (changes requested):
Looks good

compiler/modules/types/list.bal
rdhananjaya: There are places where we need to go over the `fixedLength` and set some individual member.

jclark (changes requested):
> There are places where we need to go over the `fixedLength` and set some individual member. 

What places exactly?

The semantics of FixedLengthArray is that it's an array, but with an optimized representation for certain cases.  It's not coherent to set past the length, unless it is 1 past the length, because that would leave a hole.  If it's just one then that's OK, but the code should be clear about it.

compiler/modules/types/list.bal
```diff
@@ -359,6 +362,17 @@ function fixedArrayShallowCopy(FixedLengthArray array) returns FixedLengthArray
     return { initial: shallowCopyTypes(array.initial), fixedLength: array.fixedLength };
 }
 
+function listConjunctionMaxInitialLength(Context cx, Conjunction? con) returns int {
```
jclark: Would it be an improvement to change this function to create a record ListConjunction, which is like Conjunction but contains a ListAtomicType instead of an atom, and has a maxInitialLen field which is the maximum of the initial length of this and all conjunctions? (This would be easiest to write recursively.) Then you don't need the maxInitialLen argument to listInhabited and you don't need the do to Atom->ListAtomicType conversion twice. You would then compute maxInitialLen in listInhabited from members and neg.maxInitialLen.

compiler/modules/types/list.bal
```diff
@@ -195,7 +195,10 @@ function listFormulaIsEmpty(Context cx, Conjunction? pos, Conjunction? neg) retu
             rest = NEVER;
         }
     }
-    return !listInhabited(cx, members, rest, neg);
+    // We only need to fill `maxInitialLen` number of places in any FixedLenthArray
```
jclark: Spelling "Lenth".

Put some punctuation between the two comment lines.

compiler/modules/types/list.bal
```diff
@@ -284,21 +291,19 @@ function listInhabited(Context cx, FixedLengthArray members, SemType rest, Conju
         // SemType d1 = diff(s[1], t[1]);
         // return !isEmpty(cx, d1) &&  tupleInhabited(cx, [s[0], d1], neg.rest);
         // We can generalize this to tuples of arbitrary length.
-        int maxInitialLength = members.initial.length();
-        if maxInitialLength != 0 {
```
jclark, line 288: Why isn't this check needed now?

jclark:
Looks good

compiler/modules/types/list.bal
rdhananjaya: As I remember, this was added to protect the call to `fixedArrayGet` in line:292, it was replaced by a call to `listMemberAt`.

Issue #101, opened 2021-06-20
Implement logical expressions `&&` and `||` in the front-end


jclark:
Specified in

https://ballerina.io/spec/lang/2021R1/#section_6.31

I don't think they need anything from the backend.

jclark:
Better to do #129 first.

jclark:
This will need doc added to subset13.

manuranga:
@jclark Should following code compile ?
```ballerina
function f(int|string x) {
    if false || x is string {
        string _ = x;
        return;
    }
    int _ = x;
}
```

** Edit: There was a mistake in the sample, changed `true` to `false`

jclark:
Please explain the issue.

manuranga:
Following shouldn't compile, but I think the above should. Since spec doesn't explicitly mention constants, can I still infer above should compile, or do I need to create a spec issue.

```ballerina
function f(boolean b, int|string x) {
    if b || x is string {
        string _ = x; // @error
        return;
    }
    int _ = x;
}
```

jclark:
By explain the issue, I mean explain why you think it should or should not compile.

Expression typing is based on singleton types not constants.

manuranga:
Created [ballerina-spec#1029](https://github.com/ballerina-platform/ballerina-spec/issues/1029)  to track above.

manuranga:
I need to move `env` back out of the the `ExprContext` into a parameter.
Consider `x is string && isEmptyString(x)`, env created by `x is string` needs to pass to `isEmptyString(x)`. Another option is to update (push/pop) in the `ExprContext`, but extra state is undesirable.
I'll send this as a separate pr.

jclark:
You don't need to do that: just create a new ExprContext with different env.

PR #848, opened 2022-01-12
Implement logical expression parsing

manuranga:
Part of #101


jclark (commented):
I had a quick look now and it looks good. Will look in more detail tomorrow.

PR #869, opened 2022-01-19
Ensure start positions at not at end of line or a comment 

### LINKED ISSUE #786: Ensure start positions at not at end of line or a comment
The index equal to the length of the line (not including the terminating newline) should only be used as a endPos. It should never be used as a startPos and (probably) should never be used as the position of a BIR insn.

There was a bug for this in #784. We should make sure it does not happen elsewhere.

heshanpadmasiri:
## Purpose
Ensure start positions at not at end of line or a comment 
Resolves #786

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/front/stmt.bal
+++ b/compiler/modules/front/stmt.bal
@@ -268,17 +268,16 @@ function codeGenScope(StmtContext cx, bir:BasicBlock bb, Environment initialEnv,
         applyEffect(env, narrowings, effect);
     }
     else {
-        int lastStmtIndex = scope.stmts.length() - 1;
-        int stmtIndex = 0;
+        StmtEffect? previousEffect = ();
         foreach var stmt in scope.stmts {
+            // add narrowing to previous stmt
+            if previousEffect != () {
+                addNarrowings(cx, <bir:BasicBlock>previousEffect.block, env, previousEffect.narrowings, stmt.startPos);
+            }
             StmtEffect effect = check codeGenStmt(cx, curBlock, env, stmt);
             curBlock = effect.block;
+            previousEffect = curBlock != () ? effect : ();
             applyEffect(env, narrowings, effect);
-            // Compound statements will gen narrowings post-block, no need to narrow after last stmt
-            if curBlock != () && stmtIndex != lastStmtIndex {
-                addNarrowings(cx, curBlock, env, effect.narrowings, stmt.endPos);
-            }
-            stmtIndex += 1;
         }
     }
     check unusedLocalVariables(cx, env, initialEnv.bindings);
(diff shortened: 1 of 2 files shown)


compiler/modules/front/stmt.bal
```diff
@@ -276,7 +276,7 @@ function codeGenScope(StmtContext cx, bir:BasicBlock bb, Environment initialEnv,
             applyEffect(env, narrowings, effect);
             // Compound statements will gen narrowings post-block, no need to narrow after last stmt
             if curBlock != () && stmtIndex != lastStmtIndex {
-                addNarrowings(cx, curBlock, env, effect.narrowings, stmt.endPos);
+                addNarrowings(cx, curBlock, env, effect.narrowings, stmt.startPos);
```
jclark: Not sure about this. See what @manuranga thinks.

manuranga: I don't think this is correct, @heshanpadmasiri does this break the validation? May I have a chat with you?

manuranga: Had a chat,
Option 1) We'll leave it as endPos, since it doesn't highlight nor should it be debugable via breakpoints
Option 2) Rotate the loop such that each iteration generate narrowings for the previous, then we can use the startPos of current statement as the narrowing pos of the previous.

I think let's go with Option 1.

jclark: It's not OK to use an endPos as the position of a BIR insn.

jclark: If you what you want is the position of the close brace, then you should use that.

manuranga: We (potentially) add narrowing between each statement, so close brace is not a good fit. @heshanpadmasiri shall we explore above options 2 in that case?

(I have a minor concern with that though, if feels more correct for the narrowing to be part of the previous statement, therefore derive position from that)

compiler/modules/front/stmt.bal
```diff
@@ -268,17 +268,16 @@ function codeGenScope(StmtContext cx, bir:BasicBlock bb, Environment initialEnv,
         applyEffect(env, narrowings, effect);
     }
     else {
-        int lastStmtIndex = scope.stmts.length() - 1;
-        int stmtIndex = 0;
+        StmtEffect? previousEffect = ();
         foreach var stmt in scope.stmts {
+            // add narrowing to previous stmt
+            if previousEffect != () {
+                addNarrowings(cx, <bir:BasicBlock>previousEffect.block, env, previousEffect.narrowings, stmt.startPos);
```
jclark, line 275: Why is there such a big change just to fix the position passed into addNarrowings?  Can't you just keep track of the previous stmt and use its endPos?

jclark (changes requested):

compiler/modules/front/stmt.bal
manuranga: Isn't it simpler this way since we got rid of `lastStmtIndex` and `stmtIndex`.

jclark: OK.

PR #870, opened 2022-01-20
Find semantic errors related to functions in front-end instead of BIR verify

### LINKED ISSUE #812: Front-end should not rely on BIR verification
Currently front-end relies on BIR verification to find and report some semantic errors.

The goal of this issue is for the front-end to find and report all semantic errors itself, so that BIR verification can be optional.

Reasons to change our strategy now:
- front-end has richer location information and so can give better error messages
- we are doing unnecessary duplicate work in the verifier
- need for BIR verification is less urgent, since likely to be some time before jBallerina can produce our BIR
- when doing #804, we have to do more semantic checking in codeGen (seeing this already here https://github.com/ballerina-platform/nballerina/commit/259b92297d5f3021de96730862330e1ef56f7272)

ushirask:


compiler/modules/front/expr.bal
```diff
@@ -1018,6 +1020,21 @@ function validArgumentCount(ExprContext cx, bir:FunctionRef func, int nSuppliedA
     }
 }
 
+function validArgumentTypes(ExprContext cx, bir:FunctionRef func, bir:Operand[] suppliedArgs, s:MethodCallExpr|s:FunctionCallExpr expr) returns CodeGenError? {
+    foreach int i in 0 ..< suppliedArgs.length() {
+        if operandHasType(cx.mod.tc, suppliedArgs[i], func.signature.paramTypes[i]) {
+            continue;
+        }
+        return cx.semanticErr(`wrong argument type for parameter ${i + 1} in call to function ${symbolToString(cx.mod, cx.defn.part.partIndex, func.symbol)}`, expr.args[i].startPos);
```
jclark: This won't work because method arg expression index is one less than expression arg expression index.

compiler/modules/front/expr.bal
```diff
@@ -1018,6 +1020,21 @@ function validArgumentCount(ExprContext cx, bir:FunctionRef func, int nSuppliedA
...
+    }
+    return ();
+}
+ 
+
+public function operandHasType(t:Context tc, bir:Operand operand, t:SemType semType) returns boolean {
```
jclark: This function should be further down the file.

compiler/modules/front/expr.bal
jclark: Also use a range for the argument expression.

jclark (changes requested):

compiler/modules/bir/verify.bal
```diff
@@ -159,7 +159,7 @@ function verifyCall(VerifyContext vc, CallInsn insn) returns err:Semantic? {
         }
     }
     foreach int i in 0 ..< nSuppliedArgs {
-        check verifyOperandType(vc, insn.args[i], sig.paramTypes[i], `wrong argument type for parameter ${i + 1} in call to function ${vc.symbolToString(func.symbol)}`, vc.qNameRange(insn.pos));
+        check verifyOperandType(vc, insn.args[i], sig.paramTypes[i], `bad BIR: wrong argument type for parameter ${i + 1} in call to function ${vc.symbolToString(func.symbol)}`, vc.qNameRange(insn.pos));
```
jclark: This needs updating.

compiler/modules/front/expr.bal
```diff
@@ -1018,6 +1020,26 @@ function validArgumentCount(ExprContext cx, bir:FunctionRef func, int nSuppliedA
     }
 }
 
+function validArgumentTypes(ExprContext cx, bir:FunctionRef func, bir:Operand[] suppliedArgs, s:MethodCallExpr|s:FunctionCallExpr expr) returns CodeGenError? {
+    foreach int i in 0 ..< suppliedArgs.length() {
+        if operandHasType(cx.mod.tc, suppliedArgs[i], func.signature.paramTypes[i]) {
+            continue;
+        }
+        Range range;
```
jclark: First identify the relevant expr. Then get its range (once)

jclark (changes requested):

compiler/modules/front/expr.bal
```diff
@@ -1158,6 +1160,26 @@ function validArgumentCount(ExprContext cx, bir:FunctionRef func, int nSuppliedA
...
+        }
+        s:Expr argExpr;
+        if expr is s:FunctionCallExpr {
+            argExpr = expr.args[i];
+        }
+        else if i == 0 {
+            argExpr = expr;
+        }
+        else {
+            argExpr = expr.args[i-1];
```
jclark: Spaces around `-`.

jclark (changes requested):

compiler/modules/front/expr.bal
```diff
@@ -1158,6 +1160,26 @@ function validArgumentCount(ExprContext cx, bir:FunctionRef func, int nSuppliedA
...
+        if expr is s:FunctionCallExpr {
+            argExpr = expr.args[i];
+        }
+        else if i == 0 {
+            argExpr = expr;
+        }
+        else {
+            argExpr = expr.args[i - 1];
+        }
+        return cx.semanticErr(`wrong argument type for parameter ${i + 1} in call to function ${symbolToString(cx.mod, cx.defn.part.partIndex, func.symbol)}`, s:range(argExpr));
```
jclark: A better way to solve this problem is to add a `codeGenExprForType` function (like codeGenExprForBoolean/String), which takes a semType argument. Then use that instead of codeGenExpr in codeGenMethodCallExpr/codeGenFunctionCallExpr.

Add test cases which have arguments on different lines to check that the error is for the right line.

For the too many arguments case, the error message should have the location of the first excess argument.

For the too few arguments case, the error message should be the location of closing parenthesis of the function call (may need to fix the AST to do this.)

There's no need for the error message to mention the parameter number or function name give that we have ranges. This applies for both wrong number and wrong type of messages.

jclark (changes requested):

compiler/modules/front/expr.bal
```diff
@@ -1141,12 +1149,12 @@ function codeGenFunctionCallExpr(ExprContext cx, bir:BasicBlock bb, s:FunctionCa
...
+    check validArgumentCount(cx, func, expr);
 
     t:SemType[] paramTypes = func.signature.paramTypes;
     bir:Operand[] args = [target];
     foreach int i in 0 ..< expr.args.length() {
-        var { result: arg, block: nextBlock } = check codeGenExpr(cx, curBlock, paramTypes[i + 1], expr.args[i]);
+        var { result: arg, block: nextBlock } = check codeGenExprForType(cx, curBlock, expr.args[i], paramTypes[i + 1]);
         curBlock = nextBlock;
         args.push(arg);
     }
```
jclark, line 1169: Add this one
```
function codeGenArgument(ExprContext cx, bir:BasicBlock nextBlock, bir:MethodCallExpr|s:FunctionCallExpr callExpr, bir:FunctionRef func, int i) returns ExprEffect|CodeGenError { }
```

Have error saying "incorrect type for argument".
Get rid of codeGenForType.
Handle too many arguments here.

Then do `check sufficientArguments(cx, func, expr);` at the end.

jclark (changes requested):

compiler/modules/front/expr.bal
```diff
@@ -176,6 +176,22 @@ function codeGenExprForString(ExprContext cx, bir:BasicBlock bb, s:Expr expr) re
     return cx.semanticErr("expected string operand", s:range(expr));
 }
 
+function codeGenArgument(ExprContext cx, bir:BasicBlock bb, s:MethodCallExpr|s:FunctionCallExpr callExpr, bir:FunctionRef func, int i) returns ExprEffect|CodeGenError {
+    s:Expr arg = callExpr.args[i];
+    int n = callExpr is s:FunctionCallExpr ? i : i + 1;
+    if n >= func.signature.paramTypes.length() {
+        if symbolToString(cx.mod, cx.defn.part.partIndex, func.symbol) == "io:println" {
```
jclark: Don't convert the name to a string.

jclark (changes requested):

compiler/modules/front/expr.bal
```diff
@@ -176,6 +176,22 @@ function codeGenExprForString(ExprContext cx, bir:BasicBlock bb, s:Expr expr) re
     return cx.semanticErr("expected string operand", s:range(expr));
 }
 
+function codeGenArgument(ExprContext cx, bir:BasicBlock bb, s:MethodCallExpr|s:FunctionCallExpr callExpr, bir:FunctionRef func, int i) returns ExprEffect|CodeGenError {
+    s:Expr arg = callExpr.args[i];
+    int n = callExpr is s:FunctionCallExpr ? i : i + 1;
+    if n >= func.signature.paramTypes.length() {
+        if func.symbol == { module: { org: "ballerina", names: ["io"] }, identifier: "println" } {
```
jclark: Make a `final readonly &` variable (e.g. `IO_PRINTLN_SYMBOL`) for the RHS to avoid constructing it each time

jclark (changes requested):

compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -7,7 +7,7 @@ function parseExpr(Tokenizer tok) returns Expr|err:Syntax {
     Position startPos = tok.currentStartPos();
     if t == "[" {
         check tok.advance();
-        Expr[] members = check parseExprList(tok, "]");
+        ExprList {exprs : members} = check parseExprList(tok, "]");
```
jclark: Do we need ExprList? Can't we use `currentEndPos`?

compiler/modules/front/expr.bal
```diff
@@ -176,6 +176,27 @@ function codeGenExprForString(ExprContext cx, bir:BasicBlock bb, s:Expr expr) re
...
+    module: { org: "ballerina", names: ["io"] }, 
+    identifier: "println"
+};
+
+function codeGenArgument(ExprContext cx, bir:BasicBlock bb, s:MethodCallExpr|s:FunctionCallExpr callExpr, bir:FunctionRef func, int i) returns ExprEffect|CodeGenError {
+    s:Expr arg = callExpr.args[i];
+    int n = callExpr is s:FunctionCallExpr ? i : i + 1;
+    if n >= func.signature.paramTypes.length() {
+        if func.symbol == IO_PRINTLN_SYMBOL {
+            return cx.unimplementedErr("io:println can only take 1 argument", s:range(arg));
```
jclark: "multiple arguments for io:println not implemented"

jclark (changes requested):

compiler/modules/front.syntax/parseExpr.bal
ushirask: @jclark do you mean whether  `currentEndPos` will give the same as `closeParenPos`? I don't think they'll give the same position.

jclark: Yes, you're right: currentEndPos would be one more than closeParenPos.

I suggest you make parseExprList return a tuple `[Expr[], Position]`

compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -411,20 +411,20 @@ function parseIdentifierOrMethodName(Tokenizer tok) returns string|err:Syntax {
...
     Position endPos = tok.previousEndPos();
-    return { startPos, endPos, opPos, namePos, openParenPos, target, methodName, args };
+    return { startPos, endPos, opPos, namePos, openParenPos, closeParenPos: exprList[1], target, methodName, args: exprList[0] };
 }
 
 function finishFunctionCallExpr(Tokenizer tok, string? prefix, string funcName, Position startPos) returns FunctionCallExpr|err:Syntax {
     Position openParenPos = tok.currentStartPos();
     check tok.advance();
-    Expr[] args = check parseExprList(tok, ")");
+    [Expr[], Position] exprList  = check parseExprList(tok, ")");
```
jclark: `var [args, closeParenList] =`

compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -7,9 +7,9 @@ function parseExpr(Tokenizer tok) returns Expr|err:Syntax {
     Position startPos = tok.currentStartPos();
     if t == "[" {
         check tok.advance();
-        Expr[] members = check parseExprList(tok, "]");
+        [Expr[], Position] exprList  = check parseExprList(tok, "]");
```
jclark: `var [members, _] = `

jclark (changes requested):

PR #872, opened 2022-01-20
 Fix re-narrow inside loop bugs

manuranga:
nightly build fails with
```
ERROR [parseStmt.bal:(178:9,178:29)] invalid attempt to assign a value to a variable narrowed outside the loop
ERROR [preparse.bal:(26:9,26:52)] invalid attempt to assign a value to a variable narrowed outside the loop
```

This fixes it.


jclark:
This is an excellent example of why the language design is completely broken here.

PR #873, opened 2022-01-21
Add OptionalTypeDesc to the AST

### LINKED ISSUE #867: Add OptionalTypeDesc to the AST
Currently we represent both T|() and T? similarly in the AST (Both as BinaryTypeDesc with right as BuiltinTypeDesc null) with only difference being for T? case BinaryTypeDesc’s right having same start and end pos. Instead we need to make this difference explicit in the AST.

heshanpadmasiri:
## Purpose
Add OptionalTypeDesc to the AST
Resolves #867

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/front.syntax/ast.bal
```diff
@@ -459,6 +459,12 @@ public type GroupingTypeDesc record {|
     TypeDesc innerTd;
 |};
 
+public type OptionalTypeDesc record {|
```
jclark: OptionalTypeDesc (and GroupingTypeDesc) are not consistent with UnaryTypeDesc below.

jclark (changes requested):

compiler/modules/front.syntax/ast.bal

jclark: The field names need to be distinct.

This still isn't consistent: UnaryTypeDesc has an `op`.

For any change to the AST please propose a design in the issue before implementing.

jclark (changes requested):

compiler/modules/front.syntax/ast.bal

compiler/modules/front.syntax/ast.bal
```diff
@@ -524,9 +518,22 @@ public type BinaryTypeDesc record {|
...
+public type PrefixTypeDesc record {|
     *PositionFields;
-    UnaryTypeOp op;
+    PrefixTypeOp op;
+    TypeDesc td;
+|};
+
+public const PostfixTypeOp = "?";
+
+public type PostfixTypeDesc record {|
```
jclark: Remember that type names are just aliases.  So what you have defined is completely equivalent to having a UnaryTypeDesc record with an `op` field of `"!"|"?"|"("`. That is OK as a design but you can be written much more simply than what you have here.

jclark (changes requested):

compiler/modules/front.syntax/ast.bal

PR #874, opened 2022-01-21
Implement logical exprs with a single narrowing

manuranga:
Part of #101

### DIFF
--- a/compiler/modules/front/expr.bal
+++ b/compiler/modules/front/expr.bal
@@ -105,6 +105,25 @@ class ExprContext {
     function stmtContext() returns StmtContext|CodeGenError {
         return self.sc;
     }
+
+    function createNarrowedExprContext(bir:BasicBlock bb, Binding binding, t:SemType narrowedType, bir:Result basis, Position pos) returns ExprContext {
+        bir:Register narrowed = self.sc.createVarRegister(narrowedType, binding.name, pos);
+        bir:CondNarrowInsn insn = {
+            result: narrowed,
+            operand: binding.reg,
+            basis,
+            pos
+        };
+        bb.insns.push(insn);
+        Binding bindings = {
+            name: binding.name,
+            reg: narrowed,
+            isFinal: binding.isFinal,
+            prev: self.env.bindings,
+            unnarrowed: unnarrowBinding(binding)
+        };
+        return new(self.sc, { bindings, assignments: self.env.assignments });
+    }
 }
 
 function codeGenExprForBoolean(ExprContext cx, bir:BasicBlock bb, s:Expr expr) returns CodeGenError|BooleanExprEffect {
@@ -179,9 +198,7 @@ function codeGenExpr(ExprContext cx, bir:BasicBlock bb, t:SemType? expected, s:E
             return codeGenCheckingExpr(cx, bb, expected, checkingKeyword, operand, pos);
         }
         var { opPos: pos, logicalOp: op, left, right } => {
-            var { result: l, block: block1 } = check codeGenExprForBoolean(cx, bb, left);
-            var { result: r, block: nextBlock } = check codeGenExprForBoolean(cx, block1, right);
-            return codeGenLogicalBinaryExpr(cx, nextBlock, op, pos, l, r);
+            return codeGenLogicalBinaryExpr(cx, bb, op, pos, left, right);
         }
         var { opPos: pos, bitwiseOp: op, left, right } => {
             var { lhs, rhs, nextBlock, ifNilBlock } = check codeGenBinaryNilLift(cx, expected, left, right, bb, pos);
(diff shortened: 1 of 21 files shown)


compiler/modules/front/expr.bal
```diff
@@ -533,11 +551,110 @@ function codeGenArithmeticBinaryExpr(ExprContext cx, bir:BasicBlock bb, bir:Arit
...
+        }
+    }
+    bir:AssignInsn assignInsn = { result, operand: rhs, pos };
+    evalRightBlock2.insns.push(assignInsn);
+    bir:BranchInsn branchInsn = { dest: contBlock.label, pos };
+    evalRightBlock2.insns.push(branchInsn);
+    return { result, block: contBlock, narrowing };
+}
+
+function truth(ExprNarrowing n) returns t:SemType {
```
jclark: This needs a longer name. Maybe something like typeImpliedByTruth (if that's what it's doing.)

compiler/modules/front/expr.bal
```diff
@@ -533,11 +551,110 @@ function codeGenArithmeticBinaryExpr(ExprContext cx, bir:BasicBlock bb, bir:Arit
     return { result, block: bb };
 }
 
-function codeGenLogicalBinaryExpr(ExprContext cx, bir:BasicBlock bb, s:BinaryLogicalOp op, Position pos, bir:BooleanOperand lhs, bir:BooleanOperand rhs) returns CodeGenError|ExprEffect {
-    if lhs is boolean && rhs is boolean {
-        return { result: logicalEval(op, lhs, rhs), block: bb };
+function codeGenLogicalBinaryExpr(ExprContext cx, bir:BasicBlock bb, s:BinaryLogicalOp op, Position pos, s:Expr left, s:Expr right) returns CodeGenError|ExprEffect {
+    var { result: lhs, block: block1, narrowing: lhsNarrowing } = check codeGenExprForBoolean(cx, bb, left);
+    if lhs is boolean {
+        boolean shortCircuit = op == "||";
```
jclark: You use `op == "||"` several times, so I would define `boolean isOr` at the outermost level.

compiler/modules/front/expr.bal
```diff
@@ -56,7 +56,8 @@ type ExprNarrowing record {|
     Binding binding;
     t:SemType ifTrue;
     t:SemType ifFalse;
-    bir:InsnRef testInsn;
+    bir:Result trueBasis;
+    bir:Result falseBasis;
     boolean negated = false;
```
jclark: IIRC I had `negated` so that I could generated the right basis from the InsnRef.  If you put the basis in, then I think you may be able to negate just by swapping the true/false fields, which would avoid the need for the smaller helper functions that are in effect just dealing with `negated`.

jclark (changes requested):

compiler/modules/front/expr.bal
```diff
@@ -533,11 +551,110 @@ function codeGenArithmeticBinaryExpr(ExprContext cx, bir:BasicBlock bb, bir:Arit
...
+    if lhs is boolean {
+        boolean shortCircuit = op == "||";
+        if lhs == shortCircuit {
+            return <ExprEffect>{ result: shortCircuit, block: bb };
+        }
+        else {
+            return codeGenExprForBoolean(cx, block1, right);
+        }
+    }
+    final bir:Register result = lhs; // reuse lhs register as result
```
jclark: Make a new one.

manuranga: Fixed, but I had to add an additional BasicBlock.

compiler/modules/front/expr.bal
```diff
@@ -533,11 +552,102 @@ function codeGenArithmeticBinaryExpr(ExprContext cx, bir:BasicBlock bb, bir:Arit
...
+    }
+    else {
+        if lhsNarrowing != () && rhsNarrowing != () {
+            if lhsNarrowing.binding.name != rhsNarrowing.binding.name {
+                return cx.unimplementedErr("not implemented: logical expr that narrows multiple variables", pos=pos);
+            }
+            narrowing = {
+                binding: lhsNarrowing.binding,
+                ifTrue: t:intersect(lhsNarrowing.ifTrue, rhsNarrowing.ifTrue),
+                ifFalse: t:union(lhsNarrowing.ifFalse, t:intersect(lhsNarrowing.ifTrue, rhsNarrowing.ifFalse)),
```
manuranga: @jclark 
Spec says this, but `t:union(lhsNarrowing.ifFalse, rhsNarrowing.ifFalse)` also works.

manuranga: narrowing(1)

compiler/modules/front/expr.bal
```diff
@@ -56,8 +56,8 @@ type ExprNarrowing record {|
     Binding binding;
     t:SemType ifTrue;
```
jclark: If you are using `basisIfTrue`, then this should be `typeIfTrue`.

compiler/modules/front/expr.bal
```diff
@@ -165,17 +184,17 @@ function codeGenExpr(ExprContext cx, bir:BasicBlock bb, t:SemType? expected, s:E
             bir:BooleanNotInsn insn = { operand, result, pos };
             nextBlock.insns.push(insn);
             if narrowing != () {
-                narrowing.negated = !narrowing.negated;
+                t:SemType tmp = narrowing.ifTrue;
```
jclark: Don't you have to swap the basis?

jclark (changes requested):

compiler/modules/front/expr.bal
manuranga: Ah yes.

compiler/modules/front/expr.bal
manuranga: This would cut down the amount of code since or/and becomes dual of each other.

PR #875, opened 2022-01-24
Extend semtype projection to allow arbitrary string subtypes as keys

### LINKED ISSUE #845: Extend semtype projection to allow arbitrary string subtypes as keys
This is the last bit of #522 and is needed for #575.

rdhananjaya:
## Purpose
Fixes https://github.com/ballerina-platform/nballerina/issues/845

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/types/core.bal
```diff
@@ -960,7 +960,7 @@ function bddMappingAtomicType(Env env, Bdd bdd, MappingAtomicType top) returns M
 // for when T is a subtype of mapping, and K is either `string` or a singleton string.
 // This is what Castagna calls projection.
 // We will extend this to allow `key` to be a SemType, which will turn into a StringSubtype.
-public function mappingMemberType(Context cx, SemType t, string? k = ()) returns SemType {
+public function mappingMemberType(Context cx, SemType t, string|SemType? k = ()) returns SemType {
```
jclark: Type of `k` should just be SemType.

You need to fix `mappingMemberRequired` also (also change `k` to `SemType`)

compiler/modules/types/mapping.bal
```diff
@@ -393,7 +393,7 @@ class MappingPairing {
     private function curName2() returns string => self.names2[self.i2];
 }
 
-function bddMappingMemberType(Context cx, Bdd b, string? key, SemType accum) returns SemType {
+function bddMappingMemberType(Context cx, Bdd b, string|SemType? key, SemType accum) returns SemType {
```
jclark: key should have type `StringSubtype?`

jclark (changes requested):

compiler/modules/types/core.bal
rdhananjaya: **All** string to represent the present `nil`?
Edit: I think answer to this is YES

rdhananjaya: @jclark what should `mappingMemberRequired` do when `k` contains more than one string singleton?

rdhananjaya: Made it so that if all singleton strings in the string subtype are field names then to return `true` from `mappingMemberRequired`

compiler/modules/bir/verify.bal
```diff
@@ -201,7 +201,7 @@ function verifyMappingConstruct(VerifyContext vc, MappingConstructInsn insn) ret
     }
     t:MappingAtomicType? mat = t:mappingAtomicTypeRw(vc.typeContext(), ty);
     foreach int i in 0 ..< insn.operands.length() {
-        t:SemType memberType = t:mappingMemberType(vc.typeContext(), ty, insn.fieldNames[i]);
+        t:SemType memberType = t:mappingMemberType(vc.typeContext(), ty, t:stringConst(insn.fieldNames[i]));
```
jclark: Should use t:singleton here, so that it gets memoized.

compiler/modules/front/expr.bal
```diff
@@ -381,17 +381,17 @@ type MappingAccessType "."|"["|"fill";
...
-        if t:mappingMemberRequired(cx.mod.tc, mapping.semType, kVal) {
-            maybeMissing = false;
-        }
-        else if accessType == "." {
-            return cx.semanticErr(`field access to ${kVal} is invalid because field may not be present`, pos=pos);
+    if t:mappingMemberRequired(cx.mod.tc, mapping.semType, k.semType) {
+        maybeMissing = false;
+    }
+    else if accessType == "." {
+        string? fieldName = t:singleStringShape(k.semType);
```
jclark: The field name should be in k.value.

compiler/modules/front/expr.bal
```diff
@@ -381,17 +381,14 @@ type MappingAccessType "."|"["|"fill";
...
-        if t:mappingMemberRequired(cx.mod.tc, mapping.semType, kVal) {
-            maybeMissing = false;
-        }
-        else if accessType == "." {
-            return cx.semanticErr(`field access to ${kVal} is invalid because field may not be present`, pos=pos);
-        }
+    if t:mappingMemberRequired(cx.mod.tc, mapping.semType, k.semType) {
+        maybeMissing = false;
+    }
+    else if accessType == "." && k is bir:StringConstOperand {
```
jclark: Cast k: all the cases that call with accessType "." should pass it as a StringConstOperand.

compiler/modules/types/core.bal
```diff
@@ -960,23 +960,31 @@ function bddMappingAtomicType(Env env, Bdd bdd, MappingAtomicType top) returns M
...
+    if k !is ComplexSemType {
+        return false;
+    }
     if t is UniformTypeBitSet {
         return false;
     }
     else {
-        return bddMappingMemberRequired(cx, <Bdd>getComplexSubtypeData(t, UT_MAPPING_RW), k, false)
-               && bddMappingMemberRequired(cx, <Bdd>getComplexSubtypeData(t, UT_MAPPING_RO), k, false);
+        StringSubtype key = <StringSubtype>getComplexSubtypeData(k, UT_STRING);
```
jclark: Use stringSubtype

compiler/modules/types/core.bal
```diff
@@ -960,23 +960,31 @@ function bddMappingAtomicType(Env env, Bdd bdd, MappingAtomicType top) returns M
...
+            key = <StringSubtype>getComplexSubtypeData(k, UT_STRING);
+        }
+        return union(bddMappingMemberType(cx, <Bdd>getComplexSubtypeData(t, UT_MAPPING_RO), key, TOP),
+                     bddMappingMemberType(cx, <Bdd>getComplexSubtypeData(t, UT_MAPPING_RW), key, TOP));
     }
 }
 
-public function mappingMemberRequired(Context cx, SemType t, string k) returns boolean {
+public function mappingMemberRequired(Context cx, SemType t, SemType k) returns boolean {
+    if k !is ComplexSemType {
```
jclark: Can we combine this with the next test?

compiler/modules/types/string.bal
```diff
@@ -65,6 +65,41 @@ function stringSubtypeContains(SubtypeData d, string s) returns boolean {
     return nonChar.values.indexOf(s) != () ? nonChar.allowed : !nonChar.allowed;
 }
 
+function stringSubtypeContainedIn(SubtypeData d, string[] values) returns boolean {
```
jclark: This is testing whether the d is a subtype of the set of strings in values, right?

A simpler way to do this is to construct a StringSubtype for the values and then use diff and see if you get false back, which would be O(log(N)) vs what you have here which is O(N^2).

compiler/modules/types/mapping.bal
```diff
@@ -407,13 +407,18 @@ function bddMappingMemberType(Context cx, Bdd b, string? key, SemType accum) ret
...
 
-function mappingAtomicMemberType(MappingAtomicType atomic, string? key) returns SemType {
+function mappingAtomicMemberType(MappingAtomicType atomic, StringSubtype? key) returns SemType {
     if key != () {
-        int? i = atomic.names.indexOf(key);
-        if i != () {
-            return atomic.types[i];
+        SemType m = NEVER;
+        foreach var i in 0 ..< atomic.names.length() {
+            if stringSubtypeContains(key, atomic.names[i]) {
```
jclark: This is O(N^2). atomic.names is sorted, so we can do much better.

I would add a function in string.bal that takes a StringSubtype and a sorted array of strings and returns a list of the indices of the strings that are in StringSubtype. Do this in a way that is O(log(N)). As a subroutine have a function that takes two sorted lists of strings and returns the list of the indices in the second array and that are in the first array. You can use enumerableListIntersect as the basis.

jclark (changes requested):

compiler/modules/types/string.bal
jclark: Need a function that creates a StringSubtype from a sorted list of strings by splitting into Char and NonChar

jclark: > This is testing whether the d is a subtype of the set of strings in values, right?
> 
> A simpler way to do this is to construct a StringSubtype for the values and then use diff and see if you get false back, which would be O(log(N)) vs what you have here which is O(N^2).

I mean O(N)

compiler/modules/types/core.bal
rdhananjaya: I assumed that we might not get the narrowing, I was wrong.

compiler/modules/types/mapping.bal
```diff
@@ -422,19 +426,18 @@ function mappingAtomicMemberType(MappingAtomicType atomic, string? key) returns
...
 
-function bddMappingMemberRequired(Context cx, Bdd b, string k, boolean requiredOnPath) returns boolean {
+function bddMappingMemberRequired(Context cx, Bdd b, StringSubtype k, boolean requiredOnPath) returns boolean {
     if b is boolean {
         return b ? requiredOnPath : true;
     }
     else {
         return bddMappingMemberRequired(cx, b.left, k,
-                                        requiredOnPath || cx.mappingAtomType(b.atom).names.indexOf(k) != ())
+                                        requiredOnPath || stringSubtypeFindIn(k, cx.mappingAtomType(b.atom).names)[1])
```
rdhananjaya: @jclark should we extract a variable here? Not sure if it's ok just get the second element from the tuple.

compiler/modules/types/string.bal
```diff
@@ -65,6 +65,54 @@ function stringSubtypeContains(SubtypeData d, string s) returns boolean {
     return nonChar.values.indexOf(s) != () ? nonChar.allowed : !nonChar.allowed;
 }
 
+function stringSubtypeFindIn(StringSubtype subtype, string[] sorted) returns [int[], boolean] {
```
jclark: The name does not tell me what this function does.  Needs a comment.

compiler/modules/types/string.bal
```diff
@@ -65,6 +65,54 @@ function stringSubtypeContains(SubtypeData d, string s) returns boolean {
...
+        findStringIntersections(sorted, nonChar.values, indexes);
+        stringConsts += nonChar.values.length();
+    }
+    else if nonChar.values.length() == 0 {
+        return [[], false];
+    }
+    return [indexes, stringConsts == indexes.length()];
+}
+
+function findStringIntersections(string[] values, string[] target, int[] indexes) {
```
jclark: The name of every other function in this file starts with `string`.

PR #876, opened 2022-01-24
Apply fixes for JBUG upto Beta6

### LINKED ISSUE #765: Update JBUGs for slbeta6
Update all JBUG comments to remove any that are fixed for beta6 and add bug numbers for other.

gimantha:
Partially fixes #765


compiler/main.bal
```diff
@@ -64,10 +64,8 @@ public function main(string[] filenames, *Options opts) returns error? {
                 errorFileCount += 1;
                 dPrinter.print(err.detail());
             }
-            // JBUG: #34014
-            // can't use else { check err; }
-            else if err != () {
-                return err;
+            else {
+                _ = check err;
```
jclark, line 68: Should be just `check err;`

jclark (commented):
Can you fix this last point separately?

compiler/main.bal
gimantha: @jclark I have a little confusion here. In spec, `checking-keyword` is only present in [call-stmt](https://ballerina.io/spec/lang/master/#call-stmt) under `statement`. So can this (`check err;`) be syntactically valid?

jclark: Oh yes. You're right.  Can you create a spec issue to allow `check variable-name;` as a stmt?

gimantha: sure

gimantha: Can we use https://github.com/ballerina-platform/ballerina-spec/issues/588 this to track it?

jclark: Yes. That's perfect.

PR #880, opened 2022-01-27
Make SingletonTypeDesc use SimpleConstExpr

### LINKED ISSUE #862: Make SingletonTypeDesc use SimpleConstExpr
SingletonTypeDesc just records the single value corresponding to the type, rather than fully representing the syntax tree as elsewhere in the AST. It should instead use SimpleConstExpr (as in match pattern). The work done during parsing would instead be done in resolveTypes.

The current approach has the problem we can't determine the initial token value from float SingletonTypeDesc. For example both `0f` and `0.0` will lead to a SingletonTypeDesc with value 0.0. It would be better for #816 if we can have the ability to differentiate these two

ushirask:


compiler/modules/front.syntax/astString.bal
```diff
@@ -600,6 +600,25 @@ function valueToWords(Word[] w, boolean|int|string|float|decimal|() val) {
     }
 }
 
+function simpleConstExprToWords(Word[] w, SimpleConstExpr expr) {
```
jclark: Why do you need this? SimpleConstExpr should be a subtype of Expr.

compiler/modules/front/resolveTypes.bal
```diff
@@ -258,7 +258,29 @@ function resolveTypeDesc(ModuleSymbols mod, s:ModuleLevelDefn modDefn, int depth
         }
     }
     if td is s:SingletonTypeDesc {
-        return t:singleton(mod.tc, td.value);
+        s:SimpleConstExpr valueExpr = td.valueExpr;
```
jclark: A SimpleConstExpr is an Expr so you should just be able to handle this with resolveConstExpr.

jclark (changes requested):

compiler/modules/front/resolveTypes.bal
ushirask: Fixed.

compiler/modules/front.syntax/tests/syntaxNode.bal
```diff
@@ -575,17 +580,20 @@ function syntaxNodeFromTerminalTypeDesc(TerminalTypeDesc td) returns SyntaxNode
     string token;
     if td is BuiltinTypeDesc {
         token = td.builtinTypeName == "null" ? "()" : td.builtinTypeName;
+        return { token, astNode: td };
     }
     else {
-        var value = td.value;
-        if (value is int && value < 0) || (value is float && value < 0.0) {
-            return nonTerminalSyntaxNode(td, { token: "-" }, { literal: (value * -1).toString() });
+        var valueExpr = td.valueExpr;
```
jclark: Declare type explicitly.

compiler/modules/front.syntax/ast.bal
```diff
@@ -505,7 +505,7 @@ public type TypeDescRef record {|
 
 public type SingletonTypeDesc record {|
     *PositionFields;
-    (string|float|int|boolean|decimal) value;
+    SimpleConstExpr valueExpr;
```
jclark: This is not quite right because it allows VarRefExpr.  So we need to do

```
public type ExtendedLiteralExpr LiteralExpr|NumericLiteralExpr|SimpleConstNegateExpr;
public type SimpleConstExpr ExtendedLiteralExpr|VarRefExpr;
```

and then use ExtendedLiteralExpr here.

jclark (changes requested):

compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -497,7 +497,7 @@ function parseSimpleConstExpr(Tokenizer tok) returns SimpleConstExpr|err:Syntax
     if t == "-" {
         Position opPos = tok.currentStartPos();
         check tok.advance();
-        IntLiteralExpr operand = check parseIntLiteralExpr(tok);
+        NumericLiteralExpr operand = check parseNumericLiteralExpr(tok);
```
jclark, line 500: This was a bug (match pattern with - and floating point literal) There should be a test case.

compiler/modules/front.syntax/parseTypeDesc.bal
```diff
@@ -195,90 +195,13 @@ function parsePrimaryTypeDesc(Tokenizer tok) returns TypeDesc|err:Syntax {
...
-        "false" => {
-            Position endPos = tok.currentEndPos();
-            check tok.advance();
-            return <SingletonTypeDesc>{ startPos, endPos, value: false };
-        }
-    }
-    return parseError(tok);
-}
-
-// It would be cleaner to handle this using const folding, but there are obstacles:
```
jclark, line 227: A nice lot of deletions!

jclark (changes requested):
This is looking nice.

compiler/modules/front.syntax/parseExpr.bal
ushirask: Added match pattern tests cases which passes. Does there need to be any changes made for this?

ushirask: Should we have a condition to use either parseNumericLiteralExpr or parseIntLiteralExpr depending on the use case for now here?

jclark: Why?

ushirask: Sorry I was confused about the comment earlier, added test cases instead.

PR #882, opened 2022-01-28
Redo astString on top of syntaxNode.bal

### LINKED ISSUE #865: Redo astString on top of syntaxNode.bal
We can use the tree produced by syntaxNode.bal to do astString.bal more simply.

First we need to #863.  Then astString can be done in two parts:

- transform the syntax tree by normalizing parentheses
- output the syntax tree inserting white space nicely

heshanpadmasiri:
## Purpose

- Redo astString on top of `syntaxNode.bal`
- Fix `MethodCallExpr` syntax node not adding commas between args

Resolves to #865

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


heshanpadmasiri:
JBUGs I have marked here does not exist in RC2 (only in beta 6).

compiler/modules/front.syntax/tests/canonicalTree.bal
```diff
@@ -0,0 +1,295 @@
+// lhs is expected to be the normalized tree node and rhs to be an un-normalized tree node
+function syntaxNodeEquals(SyntaxNode|RootSyntaxNode lhs, SyntaxNode|RootSyntaxNode rhs) returns boolean {
```
jclark: Why do you need this?

Can't you just normalize and then use `==`?

jclark (changes requested):

compiler/modules/front.syntax/tests/canonicalTree.bal
heshanpadmasiri: Our SyntaxNode's can have ASTNodes that have positions which we don't change as a part of normalization. So it's possible to have `lhs` and `rhs` to be same (in terms of tokens and their place in the tree) but different in terms of position

jclark: So the positions in the two trees you are comparing are different?

Where do these two trees come from and why are you comparing them? I suspect there should be a way to set things up so that the trees you are comparing have the same positions.

compiler/modules/front.syntax/tests/canonicalTree.bal
```diff
@@ -0,0 +1,295 @@
...
+    }
+    else {
+        SyntaxNode[] newChildNodes = [{ token: "(" }];
+        newChildNodes.push(...childNodes);
+        newChildNodes.push({ token: ")" });
+        return newChildNodes;
+    }
+}
+
+function syntaxNodeToString(SyntaxNode|RootSyntaxNode node) returns string {
```
jclark: Put syntaxNodeToString and related functions into syntaxNode. Put the parenthesis normalization/canonicalization into a separate file and make the file naming consistent with function naming.

The recursive functions should append strings to a `string[]`. There should be a top-level function on the root that concatenates this array. (At the moment you are doing lots of concatenations.)

compiler/modules/front.syntax/tests/canonicalTree.bal
```diff
@@ -0,0 +1,295 @@
...
+        return normalizeTypeDescSyntaxNode(node, false);
+    }
+    else {
+        return node;
+    }
+}
+
+function normalizeRootSyntaxNode(RootSyntaxNode node) returns RootSyntaxNode {
+    return { part: node.part, childNodes: from SyntaxNode child in node.childNodes select normalizeSyntaxNode(child) };
+}
```
jclark: Split these with each field on a separate line

```
return {
    part: node.part,
    childNodes: ...
};
```

compiler/modules/front.syntax/tests/canonicalTree.bal
heshanpadmasiri: First tree correspond to the ast of the un-normalized version of the code in [`testParse.json`](https://github.com/ballerina-platform/nballerina/blob/main/compiler/modules/front.syntax/tests/data/testParser.json) and the second one corresponds to the normalized version. Objective is the verify in both cases we get the same tree. Previously what we did was run astToString against the ast of the un-normalized version and see if that is same as the normalized version string wise. We can't do that until step 2 is done. (at the moment we can convert syntax's trees for both cases and compare them directly as well)

compiler/modules/front.syntax/tests/canonicalTree.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/tests/canonicalTree.bal
heshanpadmasiri: fixed

jclark:
> The recursive functions should append strings to a `string[]`. There should be a top-level function on the root that concatenates this array. (At the moment you are doing lots of concatenations.) 

You haven't fixed this.

You will still do multiple concats. You need to pass the `string[]` as a parameter (like astString does).

jclark (changes requested):
>  We can't do that until step 2 is done.

Is this pull request not doing step 1 and step 2?

We should get rid of astString as well.

heshanpadmasiri:
> > The recursive functions should append strings to a `string[]`. There should be a top-level function on the root that concatenates this array. (At the moment you are doing lots of concatenations.)
> 
> You haven't fixed this.
> 
> You will still do multiple concats. You need to pass the `string[]` as a parameter (like astString does).

fixed

heshanpadmasiri:
> > We can't do that until step 2 is done.
> 
> Is this pull request not doing step 1 and step 2?
> 
> We should get rid of astString as well.

No this is just step 1. (and minimum implementation of step 2 so we can get a readable error message by repurposing [`concat`](https://github.com/ballerina-platform/nballerina/blob/4fba86235daa2a905c7d93c0142e02cb3db7d4a4/compiler/modules/print.llvm/llvm.bal#L1567) function in print.llvm. It turn out we can easily extend that function by replacing `string` with something similar to  [`word`](https://github.com/ballerina-platform/nballerina/blob/4fba86235daa2a905c7d93c0142e02cb3db7d4a4/compiler/modules/front.syntax/astString.bal#L10) in `astString`  to add indentations and newlines. I'll send this as the part 2)

jclark (changes requested):
I don't think I can properly review part 1 on its own. Please extend this to do part 2 also. Then I will review in detail.

heshanpadmasiri:
> I don't think I can properly review part 1 on its own. Please extend this to do part 2 also. Then I will review in detail.

Fixed

compiler/modules/front.syntax/syntaxNode.bal
```diff
@@ -339,7 +339,7 @@ function syntaxNodeFromFunctionCallExpr(FunctionCallExpr expr) returns NonTermin
 }
 
 function syntaxNodeFromMethodCallExpr(MethodCallExpr expr) returns NonTerminalSyntaxNode {
-    SyntaxNode[] args = from Expr arg in expr.args select syntaxNodeFromExpr(arg);
+    SyntaxNode[] args = flattenSyntaxNodeList(from int i in 0 ..< expr.args.length() select fixedTokenSeperatedSyntaxNode(i, syntaxNodeFromExpr(expr.args[i]), ","));
```
jclark: Can we please fix this `flatten` approach to what I suggested in the call? (Joining lists with `,`, IIRC.)

compiler/modules/front.syntax/tests/testParser.bal
```diff
@@ -17,26 +17,78 @@ type SingleStringParserTestCase [Kind, string, string, string];
...
+    RootSyntaxNode|SyntaxNode normalizedActualTree = normalizeTree(actualTree);
+    RootSyntaxNode|SyntaxNode expectedTree = normalizeTree(check standardTree(k, rule, expected));
+    string[] actualTreeContent = syntaxNodeToString(normalizedActualTree);
+    string[] expectedTreeContent = syntaxNodeToString(expectedTree);
+    string errMsg = "actualTree : " + "\n".'join(...actualTreeContent) + "is not the same as expectecdTree : " + "\n".'join(...expectedTreeContent);
+    test:assertTrue(validateNormalizedTree(normalizedActualTree, expectedTree), errMsg);
+    test:assertEquals(actualTreeContent, expected);
+}
+
+function standardTree(string k, string rule, string[] content) returns err:Syntax|SyntaxNode|RootSyntaxNode {
```
jclark: `SyntaxNode|RootSyntaxNode` isn't nice.  We should rename `SyntaxNode` to something like `SubSyntaxNode` and then define `SyntaxNode` as `RootSyntaxNode|SubSyntaxNode`.

`standardTree` is not OK: you can't use `Tree` in the function names and `SyntaxNode` in the type names.

Is `content` the lines of the file? If so, call it `lines`.

There should be a type for the production rule.

```
function syntaxNodeFromLines(string[] lines, string filename, ProductionRule rule) returns err:Syntax|SyntaxNode { }
```

compiler/modules/front.syntax/syntaxNode.bal
```diff
@@ -620,3 +621,273 @@ function flattenSyntaxNodeList((SyntaxNode[]|SyntaxNode?)[] arr) returns SyntaxN
...
+            words.push(<Word>LF_OUTDENT);
+        }
+        else if i == lastChild {
+            // JBUG #33335 cast
+            words.push(<Word>LF);
+        }
+    }
+}
+
+function singletonTypedescSyntaxNodeToWords(Word[] words, SyntaxNode node) {
```
jclark: TypeDesc

compiler/modules/front.syntax/syntaxNode.bal
```diff
@@ -620,3 +621,273 @@ function flattenSyntaxNodeList((SyntaxNode[]|SyntaxNode?)[] arr) returns SyntaxN
...
+        if !foundParan && childNode is IdentifierSyntaxNode {
+            // next childNode is paran node
+            words.push(CLING);
+            foundParan = true;
+        }
+    }
+}
+
+function stmtBlockSyntaxNodeToWords(Word[] words, NonTerminalSyntaxNode node) {
+    int lastChild = node.childNodes.length() - 1;
```
jclark: lastChildIndex

compiler/modules/front.syntax/syntaxNode.bal
```diff
@@ -620,3 +621,273 @@ function flattenSyntaxNodeList((SyntaxNode[]|SyntaxNode?)[] arr) returns SyntaxN
...
+        }
+        syntaxNodeToWords(words, childNode);
+        if childNode is FixedSyntaxNode && childNode.token == "/" {
+            words.push(CLING);
+        }
+    }
+}
+
+function functionDefnSyntaxNodeToWords(Word[] words, NonTerminalSyntaxNode node) {
+    boolean foundParan = false;
```
jclark: Paren

compiler/modules/front.syntax/syntaxNode.bal
```diff
@@ -620,3 +621,273 @@ function flattenSyntaxNodeList((SyntaxNode[]|SyntaxNode?)[] arr) returns SyntaxN
...
+    syntaxNodeToWords(words, node);
+    return concat(...words);
+}
+
+function syntaxNodeToWords(Word[] words, SyntaxNode|RootSyntaxNode node) {
+    if node is TerminalSyntaxNode {
+        words.push(terminalSyntaxNodeToString(node));
+    }
+    else {
+        AstNode? astNode = node !is RootSyntaxNode ? node.astNode : ();
```
jclark: This is making stringNodeToString more tightly coupled to the AST than it should be.

I suggest you come up with a set of flags that describe the various differences in how the child nodes of non-terminal need to be formatted.  Set the flags based on the AST node. Then use the flags to output the list of child nodes.

The indent/outdent doesn't feel good to me.  Why can't you just do it based on `{` and `}` terminals?

jclark (changes requested):
This feels too much like the old astString.  There needs to be more thought about how to it right on top of syntax node.

compiler/modules/front.syntax/syntaxNode.bal
```diff
@@ -0,0 +1,830 @@
...
+    return nonTerminalSyntaxNode(expr, false, { token: expr.checkingKeyword, pos: expr.kwPos },
+                                              syntaxNodeFromExpr(expr.operand, false));
+}
+
+function syntaxNodeFromUnaryExpr(UnaryExpr expr, boolean wrap) returns NonTerminalSyntaxNode {
+    return nonTerminalSyntaxNode(expr, wrap, { token: expr.op, pos: expr.opPos, outputFlag: CLING_NEXT },
+                                             syntaxNodeFromExpr(expr.operand, true));
+}
+
+type TerminalExpr LiteralExpr|NumericLiteralExpr;
```
jclark: I don't like `TerminalExpr`.  We should rename LiteralExpr to something else (e.g. NonNumericLiteralExpr) and then define LiteralExpr to be NumericLiteralExpr|NonNumericLiteralExpr.

But NonNumericLiteralExpr is a bit long.  The point is that it is something where we don't need to the contextually expected to resolve it. Let's call it FixedLiteralExpr (unless you can think of something better).

compiler/modules/front.syntax/syntaxNode.bal
heshanpadmasiri: fixed with `joinSyntaxNodesWithSeperator`

compiler/modules/front.syntax/tests/testParser.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/syntaxNode.bal
heshanpadmasiri: function removed

compiler/modules/front.syntax/syntaxNode.bal
heshanpadmasiri: function removed

compiler/modules/front.syntax/syntaxNode.bal
heshanpadmasiri: function removed

compiler/modules/front.syntax/syntaxNode.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/syntaxNode.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/syntaxNode.bal
```diff
@@ -0,0 +1,831 @@
...
+type SubSyntaxNode NonTerminalSyntaxNode|TerminalSyntaxNode;
+type SyntaxNode SubSyntaxNode|RootSyntaxNode;
+
+const CLING = 0;
+const CLING_PREV = 1;
+const CLING_NEXT = 2;
+const NEWLINE_AFTER = 3;
+const NEWLINE_BEFORE = 4;
+const IGNORE_LITERAL_ESCAPE = 5;
+type OutputFlag CLING|CLING_PREV|CLING_NEXT|NEWLINE_AFTER|NEWLINE_BEFORE|IGNORE_LITERAL_ESCAPE;
```
jclark: Flags are things that are bitwise or'ed together (i.e. with value 0x1, 0x2, 0x4, 0x8 etc).

i.e.

```
const CLING_PREV = 0x1;
const CLING_NEXT = 0x2;


const CLING = CLING_NEXT|CLING_PREV;
```

Then use like
```
if (flags & CLING_PREV) != 0 {
}
```

jclark (changes requested):

compiler/modules/front.syntax/syntaxNode.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/syntaxNode.bal
```diff
@@ -0,0 +1,833 @@
...
+const CLING_PREV = 0x1;
+const CLING_NEXT = 0x2;
+const int CLING = CLING_PREV|CLING_NEXT;
+const NEWLINE_AFTER = 0x4;
+const NEWLINE_BEFORE = 0x8;
+const int NEWLINE = NEWLINE_BEFORE|NEWLINE_AFTER;
+const IGNORE_LITERAL_ESCAPE = 0x10;
+const WRAP = 0x20;
+const NONE = 0x0;
+type OutputFlag int;
```
jclark: Call this OutputFlags (with an "s")

compiler/modules/front.syntax/syntaxNode.bal
```diff
@@ -0,0 +1,833 @@
...
+
+type TerminalSyntaxAstNode record {|
+    AstNode astNode;
+    OutputFlag outputFlags = NONE;
```
jclark: Don't store the outputFlags in the node: just compute them when you need them.

jclark (changes requested):

compiler/modules/front.syntax/syntaxNode.bal
heshanpadmasiri: @jclark do you mean don't set a default value like this or don't set flags when we create the node?

compiler/modules/front.syntax/syntaxNode.bal
heshanpadmasiri: fixed

jclark (changes requested):
> @jclark do you mean don't set a default value like this or don't set flags when we create the node? 

The latter: do not include an `outputFlags` field in the node.

heshanpadmasiri:
> > @jclark do you mean don't set a default value like this or don't set flags when we create the node?
> 
> The latter: do not include an `outputFlags` field in the node.

fixed

compiler/modules/front.syntax/tests/data/testParser.json
```diff
@@ -220,10 +220,10 @@
     ["E", "expr", "[,1]", ""],
 
     { "comment" : "mapping constructor" },
-    ["V", "expr", "{ }"],
-    ["V", "expr", "{x: 1, y: 2}", "{ x: 1, y: 2 }"],
-    ["V", "expr", "{x: 1, y: 2, \"z\": 3}", "{ x: 1, y: 2, \"z\": 3 }"],
-    ["V", "expr", "{value: 1+2}", "{ value: 1 + 2 }"],
+    ["V", "expr", "{ }", "{\n}"],
```
jclark: We should be able to avoid this: have a flag saying don't add newlines after braces. Set this flag for a mapping constructor.

heshanpadmasiri: fixed

compiler/modules/front.syntax/tests/data/testParser.json
```diff
@@ -220,7 +220,7 @@
     ["E", "expr", "[,1]", ""],
 
     { "comment" : "mapping constructor" },
-    ["V", "expr", "{ }"],
+    ["V", "expr", "{ }", "{  }"],
```
jclark: Should fix this to do `{ }` with one space.

jclark (changes requested):

compiler/modules/front.syntax/tests/data/testParser.json
heshanpadmasiri: fixed

PR #884, opened 2022-02-01
Fix const/singleton cases of logical expr

### LINKED ISSUE #879: Logical operations does not produce an `unreachable code` error when lhs is not a single boolean but whole is
Following correctly gives an unreachable code error
```ballerina
public function main() {
    int|string x = 1;
    if x is int {
        if x !is int {
            return;
        }
    }
}
```

But following does not 

```ballerina
public function main() {
    int|string x = 1;
    if x is int && x !is int {
        return;
    }
}
```

The second type test don't produce a narrowing in ether case. Error is given based on the result being a singleton boolean, which is not getting correctly calculated in the latter case.

manuranga:
Fixes #879

### DIFF
--- a/compiler/modules/bir/bir.bal
+++ b/compiler/modules/bir/bir.bal
@@ -124,6 +124,10 @@ public function createBasicBlock(FunctionCode code, string? name = ()) returns B
     return bb;
 }
 
+public function discardBasicBlocksFrom(FunctionCode code, BasicBlock toDiscard) {
+    code.blocks.setLength(toDiscard.label);
+}
+
 public function lastInsnRef(BasicBlock bb) returns InsnRef {
     return { block: bb.label, index: bb.insns.length() - 1 };
 }
--- a/compiler/modules/front/expr.bal
+++ b/compiler/modules/front/expr.bal
@@ -115,6 +115,10 @@ class ExprContext {
         return bir:createBasicBlock(self.code, name);
     }
 
+    function discardBasicBlocksFrom(bir:BasicBlock toDiscard) {
+        return bir:discardBasicBlocksFrom(self.code, toDiscard);
+    }
+
     function stmtContext() returns StmtContext|CodeGenError {
         return <StmtContext>self.sc;
     }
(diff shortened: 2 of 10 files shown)


compiler/modules/bir/bir.bal
```diff
@@ -124,6 +124,16 @@ public function createBasicBlock(FunctionCode code, string? name = ()) returns B
     return bb;
 }
 
+public function discardBasicBlocksUpTo(FunctionCode code, BasicBlock upTo) {
```
jclark: Need comment or name change to make it clear whether up to is inclusive or exclusive.

I think you can do this more simply as something like `code.blocks.setLength(upTo.label)`.

jclark: Also I think UpTo is confising, because you are discarding from the end backwards to.  So I think it's `discardBasicBlocksFrom` i.e. from this block onwards.

jclark: This makes the use in expr.bal easier to understand.

compiler/modules/front/expr.bal
```diff
@@ -587,24 +591,36 @@ function codeGenLogicalBinaryExpr(ExprContext cx, bir:BasicBlock bb, s:BinaryLog
     var [leftValue, leftFlags] = booleanOperandValue(lhs);
     if leftFlags != 0 {
         if leftValue == isOr {
-            // XXX not correct errors in `right` won't be found
-            // XXX previous code used bb not block1 here
+            bir:BasicBlock fakeBlock = cx.createBasicBlock();
```
jclark: This isn't going to work in a const definition (because cx.createBasicBlock with create an error).

jclark (changes requested):

compiler/modules/front/expr.bal
```diff
@@ -115,6 +115,10 @@ class ExprContext {
         return bir:createBasicBlock(self.code, name);
     }
 
+    function discardBasicBlocksBelow(bir:BasicBlock newLast) {
```
jclark: After

compiler/modules/front/expr.bal
```diff
@@ -587,24 +591,38 @@ function codeGenLogicalBinaryExpr(ExprContext cx, bir:BasicBlock bb, s:BinaryLog
     var [leftValue, leftFlags] = booleanOperandValue(lhs);
     if leftFlags != 0 {
         if leftValue == isOr {
-            // XXX not correct errors in `right` won't be found
-            // XXX previous code used bb not block1 here
+            // Only to check errors on rhs, result is discarded.
+            bir:BasicBlock fakeBlock = { label: -1 };
+            _ = check codeGenExprForBoolean(cx, fakeBlock, right);
+            cx.discardBasicBlocksBelow(block1);
```
jclark: block1 is not necessarily the highest valid block number

compiler/modules/front/expr.bal
```diff
@@ -587,24 +591,38 @@ function codeGenLogicalBinaryExpr(ExprContext cx, bir:BasicBlock bb, s:BinaryLog
...
+    var { result: rhs, block: evalRightBlock2, ifTrue: rhsIfTrue, ifFalse: rhsIfFalse } = check codeGenExprForBoolean(rhsCx, evalRightBlock, right);
+    var [rightValue, rightFlags] = booleanOperandValue(rhs);
+    if rightFlags != 0 {
+        if rightValue == isOr {
+            bir:BranchInsn lhsBranchInsn = { dest: evalRightBlock2.label, pos };
+            block1.insns.push(lhsBranchInsn);
+            return constExprEffect(cx, evalRightBlock2, rightValue, rightFlags);
+        }
+        else {
+            cx.discardBasicBlocksBelow(block1);
```
jclark: That's not right. The RHS needs to be evaluated with side-effects.

jclark: Have a function with side-effects returning true/false

jclark (changes requested):

PR #894, opened 2022-02-03
 Fix bug with hex int literal with contextually expected type of float/decimal

### LINKED ISSUE #860: Bug with hex int literal with contextually expected type of float or decimal
The code appears to ignore the base of an integer literal when the contextually expected type is float or decimal.

ushirask:
Fixes bug with hex int literal with contextually expected type of float or decimal ignoring base
Fixes #860

### DIFF
--- a/compiler/modules/front/eval.bal
+++ b/compiler/modules/front/eval.bal
@@ -255,7 +255,7 @@ function convertToDecimalEval(err:SemanticContext cx, Position pos, float|int va
 }
 
 
-function fpLiteralValue(err:SemanticContext cx, t:SemType? expectedType, string untypedLiteral, string? typeSuffix, Position pos) returns float|decimal|err:Semantic|err:Unimplemented {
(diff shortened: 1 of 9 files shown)


compiler/modules/front/eval.bal
```diff
@@ -286,13 +284,20 @@ function intLiteralValue(err:SemanticContext cx, t:SemType? expectedType, s:IntL
...
+    }
+    return f;
 }
 
 // Even if the decimal floating point literal is parsed correctly,
 // overflows should return an error.
-function decimalFromFpLiteral(err:SemanticContext cx, string decimalStr, Position pos) returns decimal|err:Semantic {
+function decimalFromFpLiteral(err:SemanticContext cx, string decimalStr, Position pos, s:IntLiteralBase base = 10) returns decimal|err:Semantic|err:Unimplemented {
+    if base == 16 {
+        return cx.unimplementedErr("hexadecimal int cannot be converted to decimal", pos=pos);
```
jclark: Make this a semantic err, and document in subset13 that we have implemented the proposed spec change that I mentioned.

jclark (changes requested):

PR #904, opened 2022-02-08
Improve list projection to take a semtype as the key

### LINKED ISSUE #891: Support for typing list projection with int subtype
We need to be able to compute the type of `v[i]` when v is a subtype of list and i is an arbitrary subtype of int (not just `int` or a `singleton`).

Needed for #576.

rdhananjaya:
## Purpose
Fixes https://github.com/ballerina-platform/nballerina/issues/891

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/types/core.bal
```diff
@@ -945,13 +945,17 @@ function bddListAtomicType(Env env, Bdd bdd, ListAtomicType top) returns ListAto
 // This is what Castagna calls projection.
 // We will extend this to allow `key` to be a SemType, which will turn into an IntSubtype.
 // If `t` is not a list, NEVER is returned
-public function listMemberType(Context cx, SemType t, int? key = ()) returns SemType {
+public function listMemberType(Context cx, SemType t, SemType? k = INT) returns SemType {
```
jclark: `SemType` not `SemType?`

compiler/modules/types/core.bal
```diff
@@ -945,13 +945,17 @@ function bddListAtomicType(Env env, Bdd bdd, ListAtomicType top) returns ListAto
...
     if t is UniformTypeBitSet {
         return (t & LIST) != 0 ? TOP : NEVER;
     }
     else {
-        return union(bddListMemberType(cx, <Bdd>getComplexSubtypeData(t, UT_LIST_RO), key, TOP),
-                     bddListMemberType(cx, <Bdd>getComplexSubtypeData(t, UT_LIST_RW), key, TOP));
+        IntSubtype? intSubtype = ();
+        if k is ComplexSemType {
+            intSubtype = <IntSubtype>getComplexSubtypeData(k, UT_INT);
```
jclark: We should use new `intSubtype` function here (and for mappingMemberType).
It's a bit awkard that usually we mean `true` for all ints, but here we are using `()` for all ints. Can we do better?

compiler/modules/types/list.bal
```diff
@@ -217,9 +217,16 @@ function listIntersectWith(FixedLengthArray members, SemType rest, ListAtomicTyp
...
-            intersect(listMemberAt(members, rest, i), listMemberAt(lt.members, lt.rest, i)));
-    } 
+        fixedArraySet(members, i, intersect(
+                                        listMemberAt(members, rest, i),
+                                        listMemberAt(lt.members, lt.rest, i)));
+    }
+    // Intersect the last member of initial array
+    if nonRepeatedLen < members.fixedLength {
+        members.initial[nonRepeatedLen] =  intersect(
+                                                listMemberAt(members, rest, nonRepeatedLen),
```
jclark: I typically don't format calls like this: I put the start of the first arg on the same line as the `(`

compiler/modules/types/list.bal
```diff
@@ -255,7 +262,7 @@ function listInhabited(Context cx, FixedLengthArray members, SemType rest, ListC
             if listInhabited(cx, members, NEVER, neg.next) {
                 return true;
             }
-            foreach int i in len + 1 ..< negLen {
+            foreach int i in members.initial.length() + 1 ..< nt.members.initial.length() {
```
jclark: Why are we changing this?

compiler/modules/types/list.bal
```diff
@@ -334,7 +341,16 @@ function fixedArrayFill(FixedLengthArray arr, int newLen, SemType filler) {
     if newLen <= initial.length() {
         return;
     }
-    if arr.fixedLength == 0 || initial[initial.length() - 1] != filler {
+    int initLen = initial.length();
```
jclark, line 354: This was buggy before, right? Do we have a test case?

compiler/modules/types/int.bal
```diff
@@ -133,6 +133,21 @@ function intSubtypeComplement(SubtypeData d) returns SubtypeData {
     return rangeListComplement(v).cloneReadOnly();
 }
 
+// `min` and `max` are inclusive
+function intSubtypeOverlapRange(IntSubtype subtype, int min, int max) returns boolean {
+    IntSubtype range = [{ min, max }];
+    return intSubtypeIntersect(subtype, range) != false;
+}
+
+function intSubtypeContainsGreaterThan(IntSubtype subtype, int value) returns boolean {
```
jclark: Can we instead have:

```
function intSubtypeMax(IntSubtype subtype) returns int {
    return subtype[subtype.length() - 1].max;
}
```

Similarly for intSubtypeMin. Then use in intConstraints function in core.

compiler/modules/types/int.bal
```diff
@@ -133,6 +133,21 @@ function intSubtypeComplement(SubtypeData d) returns SubtypeData {
     return rangeListComplement(v).cloneReadOnly();
 }
 
+// `min` and `max` are inclusive
+function intSubtypeOverlapRange(IntSubtype subtype, int min, int max) returns boolean {
```
jclark: Slightly prefer to pass in a Range argument.

compiler/modules/types/listProj.bal
```diff
@@ -67,39 +68,44 @@ function listProjPath(Context cx, int k, Conjunction? pos, Conjunction? neg) ret
...
         int len = members.fixedLength;
-        ListAtomicType nt = cx.listAtomType(neg.atom);
+        ListAtomicType nt = neg.listType;
         int negLen = nt.members.fixedLength;
+        SemType p = NEVER;
         if len < negLen {
             if isNever(rest) {
                 return listProjExclude(cx, k, members, rest, neg.next);
             }
+            foreach int i in members.initial.length() + 1 ..< nt.members.initial.length() {
```
jclark: There seems to be a chunk of very similar code in listInhabited, which I don't understand either.

compiler/modules/types/listProj.bal
```diff
@@ -67,39 +68,44 @@ function listProjPath(Context cx, int k, Conjunction? pos, Conjunction? neg) ret
...
         }
         // now we have nt.members.length() <= len
-        SemType p = NEVER;
-        foreach int i in 0 ..< len {
-            SemType ntm = listMemberAt(nt.members, nt.rest, i);
-            SemType d = diff(fixedArrayGet(members, i), ntm);
+        foreach int i in 0 ..< int:max(members.initial.length(), neg.maxInitialLen) {
+            SemType d = diff(listMemberAt(members, rest, i), listMemberAt(nt.members, nt.rest, i));
             if !isEmpty(cx, d) {
                 FixedLengthArray s = fixedArrayShallowCopy(members);
```
jclark: I'm seeing multiple cases of fixedArrayShallowCopy then fill (or fill then set). A helper function might make things clearer.

jclark (changes requested):
I think you are going to need to talk me through the changes to listIhabited.

compiler/modules/types/listProj.bal
rdhananjaya: How about `fixedArraySetOnCopy` for the name?

compiler/modules/types/list.bal
rdhananjaya: This part of the code still had the large fixed-length array issue (https://github.com/ballerina-platform/nballerina/issues/820)

compiler/modules/types/list.bal
rdhananjaya: I observe this in the debugger, let me add a test

rdhananjaya: Test cases in `proj9-t.bal` are failing without this fix. Specifically `T4[1]`.

compiler/modules/types/listProj.bal
```diff
@@ -1,21 +1,19 @@
...
-        return NEVER;
-    }
+public function listProj(Context cx, SemType t, SemType k) returns SemType {
     if t is UniformTypeBitSet {
         return (t & LIST) != 0 ? TOP : NEVER;
     }
     else {
-        return union(listProjBdd(cx, k, <Bdd>getComplexSubtypeData(t, UT_LIST_RO), (), ()),
-                     listProjBdd(cx, k, <Bdd>getComplexSubtypeData(t, UT_LIST_RW), (), ()));
+        IntSubtype|boolean keyData = intSubtype(k);
```
jclark, line 9: If it's `false`, then the function will always return `never`, won't it?

So I would pass `IntSubtype|true` around.

jclark (commented):
Looks pretty good, but I need some help to understand it all.

compiler/modules/types/list.bal
```diff
@@ -217,9 +217,14 @@ function listIntersectWith(FixedLengthArray members, SemType rest, ListAtomicTyp
...
     int nonRepeatedLen = int:max(members.initial.length(), lt.members.initial.length());
     foreach int i in 0 ..< nonRepeatedLen {
-        fixedArraySet(members, i, 
-            intersect(listMemberAt(members, rest, i), listMemberAt(lt.members, lt.rest, i)));
-    } 
+        fixedArraySet(members, i, intersect(listMemberAt(members, rest, i),
+                                            listMemberAt(lt.members, lt.rest, i)));
+    }
+    // Intersect the last member of initial array
+    if nonRepeatedLen < members.fixedLength {
```
jclark: I don't understand this.

compiler/modules/types/list.bal
```diff
@@ -255,7 +260,7 @@ function listInhabited(Context cx, FixedLengthArray members, SemType rest, ListC
             if listInhabited(cx, members, NEVER, neg.next) {
                 return true;
             }
-            foreach int i in len + 1 ..< negLen {
+            foreach int i in members.initial.length() + 1 ..< nt.members.initial.length() {
```
jclark: For upper bound, why not max withneg.initialLen?

For lower bound, why not len?

compiler/modules/types/list.bal
```diff
@@ -366,6 +380,13 @@ function fixedArrayShallowCopy(FixedLengthArray array) returns FixedLengthArray
     return { initial: shallowCopyTypes(array.initial), fixedLength: array.fixedLength };
 }
 
+function fixedArraySetOnCopy(FixedLengthArray array, int index, SemType t, SemType rest) returns FixedLengthArray {
```
jclark: Don't like SetOnCopy. Maybe fixedArrayReplace

compiler/modules/types/list.bal
```diff
@@ -377,7 +398,7 @@ function listConjunction(Context cx, Conjunction? con) returns ListConjunction?
     return ();
 }
 
-function bddListMemberType(Context cx, Bdd b, int? key, SemType accum) returns SemType {
+function bddListMemberType(Context cx, Bdd b, IntSubtype|boolean key, SemType accum) returns SemType {
```
jclark: |true

jclark (changes requested):

compiler/modules/types/list.bal
```diff
@@ -265,7 +273,12 @@ function listInhabited(Context cx, FixedLengthArray members, SemType rest, ListC
             if listInhabited(cx, members, NEVER, neg.next) {
                 return true;
             }
-            foreach int i in len + 1 ..< negLen {
+            // If last member of neg repeats, just checking one repeating occurrence should suffice.
```
jclark, line 276: I don't understand why are not considering neg.maxInitialLen here.

Even if nt is the same after negLenLimit, the rest of the list may vary on each index up to neg.maxInitialLen.

jclark (changes requested):

PR #906, opened 2022-02-08
Take advantage of non-panicking insns for arithmetic on int subtypes

### LINKED ISSUE #899: Take advantage of non-panicking insns for arithmetic on int subtypes
Use limited range of int subtypes to determine when there is no possibility of overflow of arithmetic operations.

In the case of 32-bit integer operations other than division (we can have divide by 0 here) we can use `IntNoPanicArithmeticBinaryInsn` instead or `IntArithmeticBinaryInsn`.

 This is a front-end optimization because the panicking instructions are different.

heshanpadmasiri:
## Purpose
Take advantage of non-panicking insns for arithmetic on int subtypes
Resolves #899

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/front/expr.bal
+++ b/compiler/modules/front/expr.bal
@@ -496,30 +496,6 @@ function codeGenNegateExpr(ExprContext cx, bir:BasicBlock nextBlock, Position po
     return { result, block: nextBlock };
 }
 
-function isNonPanicking(bir:Operand lhs, bir:Operand rhs, bir:ArithmeticBinaryOp op) returns boolean {
-    t:IntSubtypeConstraints? lhsConstraints = t:intSubtypeConstraints(lhs.semType);
-    t:IntSubtypeConstraints? rhsConstraints = t:intSubtypeConstraints(rhs.semType);
-    if lhsConstraints == () || rhsConstraints == () {
-        return false;
-    }
-    match op {
-        "+"|"-" => {
-            // largest type that can't overflow is unsigned32 (positive) and signed32 (negative)
-            return lhsConstraints.max <= 0xffffffff && rhsConstraints.max <= 0xffffffff &&
-                   lhsConstraints.min >= -0x80000000 && rhsConstraints.min >= -0x80000000;
-        }
-        "*" => {
-            // largest type that can't overflow is unsigned16 (positive) and signed16 (negative)
-            return lhsConstraints.max <= 0xffff && rhsConstraints.max <= 0xffff &&
-                   lhsConstraints.min >= -0x8000 && rhsConstraints.min >= -0x8000;
-        }
-        "/" => {
-            return false;
-        }
-    }
-    return false;
-}
-
 function codeGenArithmeticBinaryExpr(ExprContext cx, bir:BasicBlock bb, bir:ArithmeticBinaryOp op, Position pos, bir:Operand lhs, bir:Operand rhs) returns CodeGenError|ExprEffect {
     ArithmeticOperandPair? pair = arithmeticOperandPair(lhs, rhs);
     bir:Register result;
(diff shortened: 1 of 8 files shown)


compiler/modules/front/expr.bal
```diff
@@ -574,6 +580,29 @@ function codeGenArithmeticBinaryExpr(ExprContext cx, bir:BasicBlock bb, bir:Arit
...
+function isNonPanickingIntOp(bir:Operand lhs, bir:Operand rhs, bir:ArithmeticBinaryOp op) returns boolean {
+    t:IntSubtypeConstraints? lhsConstraints = t:intSubtypeConstraints(lhs.semType);
+    t:IntSubtypeConstraints? rhsConstraints = t:intSubtypeConstraints(rhs.semType);
+    if lhsConstraints == () || rhsConstraints == () {
+        return false;
+    }
+    match op {
+        "+"|"-" => {
+            // largest type that can't overflow is unsigned32 (positive) and signed32 (negative)
+            return lhsConstraints.max <= 0xffffffff && rhsConstraints.max <= 0xffffffff &&
```
jclark: Use predefined constants in lang.int e.g. `int:UNSIGNED32_MAX_VALUE`.

In the match expression, set two variables noPanicMin and noPanicMax. Then check them against each constraint in a loop at the end.

```
foreach var c in constraints {
    if c.min < noPanicMin || c.max > noPanicMax {
        return false
    }
}
return true;
```

compiler/modules/front/expr.bal
```diff
@@ -574,6 +580,29 @@ function codeGenArithmeticBinaryExpr(ExprContext cx, bir:BasicBlock bb, bir:Arit
...
+                   lhsConstraints.min >= -0x80000000 && rhsConstraints.min >= -0x80000000;
+        }
+        "*" => {
+            // largest type that can't overflow is unsigned16 (positive) and signed16 (negative)
+            return lhsConstraints.max <= 0xffff && rhsConstraints.max <= 0xffff &&
```
jclark: Really? Can two int:Signed32 values overflow? Which?

jclark (changes requested):

compiler/modules/front/expr.bal
heshanpadmasiri: Sorry I didn't notice that there is a typo in the [spec](https://ballerina.io/spec/lang/2021R1/#built-in_subtypes) (positive side have one digit more than the negative side).

heshanpadmasiri: created issue #1038

compiler/modules/front/expr.bal
heshanpadmasiri: fixed

compiler/modules/front/expr.bal
```diff
@@ -510,7 +510,13 @@ function codeGenArithmeticBinaryExpr(ExprContext cx, bir:BasicBlock bb, bir:Arit
             return { result: { value, semType: resultType }, block: bb };
         }
         result = cx.createTmpRegister(t:INT, pos);
-        bir:IntArithmeticBinaryInsn insn = { op, pos, operands, result };
+        bir:Insn insn;
+        if isNonPanickingIntOp(operands[0], operands[1], op) {
```
jclark: I wrote a comment but it disappeared. Write like this

```
var name = intArithmeticOpNeverPanics(op, operands);
insn = { op, pos, name, operand, result };
```

jclark (changes requested):

compiler/modules/front/expr.bal
heshanpadmasiri: Fixed. Not sure if this is a bug but if I use name directly as `insn = { op, pos, name, operand, result };` (without if) compiler complains saying `a type compatible with mapping constructor expressions not found in type 'wso2/nballerina.bir:0.1.0:Insn'`

jclark: Yes, I think that is a JBUG: I seem to remember coming across something similar.

compiler/modules/front/expr.bal
```diff
@@ -574,6 +581,38 @@ function codeGenArithmeticBinaryExpr(ExprContext cx, bir:BasicBlock bb, bir:Arit
     return { result, block: bb };
 }
 
+function intArithmeticOpNeverPanics(bir:ArithmeticBinaryOp op, bir:Operand[] operands) returns bir:INSN_INT_NO_PANIC_ARITHMETIC_BINARY|bir:INSN_INT_ARITHMETIC_BINARY {
```
jclark: A function with this name needs to return boolean.

compiler/modules/front/expr.bal
```diff
@@ -510,7 +510,14 @@ function codeGenArithmeticBinaryExpr(ExprContext cx, bir:BasicBlock bb, bir:Arit
             return { result: { value, semType: resultType }, block: bb };
         }
         result = cx.createTmpRegister(t:INT, pos);
-        bir:IntArithmeticBinaryInsn insn = { op, pos, operands, result };
+        var name = intArithmeticOpNeverPanics(op, operands);
```
jclark: Add JBUG comment

jclark (changes requested):

compiler/modules/front/expr.bal
heshanpadmasiri: fixed

compiler/modules/front/expr.bal
heshanpadmasiri: fixed

Issue #900, opened 2022-02-07
Generate better code for boxing/unboxing int subtypes


heshanpadmasiri:
When we have an int subtype that fits in 56 bits, we know the representation is immediate. Converting the value to a tagged pointer is just a truncate/shift/increment. Similarly converting from a tagged pointer is just a shift/mask/sign extend.

jclark:
Be sure to test with negative numbers.

manuranga:
I added max/min intimidate int consts in my pr https://github.com/ballerina-platform/nballerina/pull/902/files#diff-e816d0935faff2498db272a1ac22b7168bbfe690b42a7a0ca8908c123d03daffR30

PR #912, opened 2022-02-10
Capture immediate info in Repr and optimize equality check

### LINKED ISSUE #893: Use integer/string subtypes to optimize more equality cases
Some equality operations can be optimized into pointer comparisons.

For some subtypes of int or string, we can determine that the tagged representation will always be immediate.

When we have an `==` comparison between two types T1 and T2, if the intersection will always use a tagged representation, we can optimize it into a pointer comparison.

Part of this is to write a function for each of int and string that determines with the subtype is always immediate.

We can do something similar for some cases of `===`.

manuranga:
Fixes #893, #900

### DIFF
--- a/compiler/modules/nback/equal.bal
+++ b/compiler/modules/nback/equal.bal
@@ -59,20 +59,18 @@ final RuntimeFunction decimalExactEqFunction = {
 type CmpEqOp "ne"|"eq";
 
 function buildEquality(llvm:Builder builder, Scaffold scaffold, bir:EqualityInsn insn) returns BuildError? {
-    var [lhsRepr, lhsValue] = check buildReprValue(builder, scaffold, insn.operands[0]);
-    var [rhsRepr, rhsValue] = check buildReprValue(builder, scaffold, insn.operands[1]);
+    var [lhs, rhs] = insn.operands;
+    var [lhsRepr, lhsValue] = check buildReprValue(builder, scaffold, lhs);
+    var [rhsRepr, rhsValue] = check buildReprValue(builder, scaffold, rhs);
     CmpEqOp op = insn.op[0] == "!" ?  "ne" : "eq"; 
     boolean exact = insn.op.length() == 3; // either "===" or "!=="
     bir:Register result = insn.result;
     match [lhsRepr.base, rhsRepr.base] {
         [BASE_REPR_TAGGED, BASE_REPR_TAGGED] => {
-            if reprIsImmediate(lhsRepr) || reprIsImmediate(rhsRepr) {
+            if reprIsImmediate(lhsRepr) || reprIsImmediate(rhsRepr) || reprIsImmediate(semTypeRepr(t:intersect(lhs.semType, rhs.semType))) {
                 return buildStoreBoolean(builder, scaffold, builder.iCmp(op, lhsValue, rhsValue), result);
             }
             else if reprIsString(lhsRepr) && reprIsString(rhsRepr) {
-                if isAnyOperandSmallString(insn.operands) {
-                    return buildStoreBoolean(builder, scaffold, builder.iCmp(op, lhsValue, rhsValue), result);
-                }
                 return buildEqualStringString(builder, scaffold, op, <llvm:PointerValue>lhsValue, <llvm:PointerValue>rhsValue, result);
             }
             else if reprIsDecimal(lhsRepr) && reprIsDecimal(rhsRepr) {
@@ -89,10 +87,10 @@ function buildEquality(llvm:Builder builder, Scaffold scaffold, bir:EqualityInsn
             return buildEqualTaggedBoolean(builder, scaffold, op, <llvm:PointerValue>rhsValue, lhsValue, result);
         }
         [BASE_REPR_TAGGED, BASE_REPR_INT] => {
-            return buildEqualTaggedInt(builder, scaffold, op, <llvm:PointerValue>lhsValue, rhsValue, result);
+            return buildEqualTaggedInt(builder, scaffold, op, <llvm:PointerValue>lhsValue, rhsValue, <IntRepr>rhsRepr, result);
         }
         [BASE_REPR_INT, BASE_REPR_TAGGED] => {
-            return buildEqualTaggedInt(builder, scaffold, op, <llvm:PointerValue>rhsValue, lhsValue, result);
+            return buildEqualTaggedInt(builder, scaffold, op, <llvm:PointerValue>rhsValue, lhsValue, <IntRepr>lhsRepr, result);
         }
         [BASE_REPR_BOOLEAN, BASE_REPR_BOOLEAN]
         | [BASE_REPR_INT, BASE_REPR_INT] => {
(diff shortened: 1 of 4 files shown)


compiler/modules/nback/scaffold.bal
```diff
@@ -37,20 +37,37 @@ const BASE_REPR_TAGGED = "BASE_REPR_TAGGED";
...
 |};
 
 // Maps any Ballerina value to a tagged pointer
 type TaggedRepr readonly & record {|
-    BASE_REPR_TAGGED base;
+    BASE_REPR_TAGGED base = BASE_REPR_TAGGED;
     t:UniformTypeBitSet subtype;
-    llvm:IntegralType llvm;
+    llvm:IntegralType llvm = LLVM_TAGGED_PTR;
+    boolean isImmediate;
```
jclark: I think `alwaysImmediate` is better: since when this is false it means there can be both immediate and non-immediate values.

compiler/modules/nback/scaffold.bal
```diff
@@ -37,20 +37,37 @@ const BASE_REPR_TAGGED = "BASE_REPR_TAGGED";
 type BaseRepr UniformBaseRepr|BASE_REPR_TAGGED;
 type RetBaseRepr BaseRepr|BASE_REPR_VOID;
 
-type UniformRepr readonly & record {|
-    UniformBaseRepr base;
-    llvm:SingleValueType llvm;
-    t:UniformTypeBitSet subtype?;
+// Maps int to i1
+type BooleanRepr readonly & record {|
```
jclark, line 47: I think it might be helpful to have a ReprFields record with base/llvm/isImmediate fields and thenn do `*ReprFields;` in each *Repr to make subtyping clearer.

compiler/modules/nback/equal.bal
```diff
@@ -59,20 +59,19 @@ final RuntimeFunction decimalExactEqFunction = {
...
+    var [lhsRepr, lhsValue] = check buildReprValue(builder, scaffold, lhs);
+    var [rhsRepr, rhsValue] = check buildReprValue(builder, scaffold, rhs);
     CmpEqOp op = insn.op[0] == "!" ?  "ne" : "eq"; 
     boolean exact = insn.op.length() == 3; // either "===" or "!=="
     bir:Register result = insn.result;
     match [lhsRepr.base, rhsRepr.base] {
         [BASE_REPR_TAGGED, BASE_REPR_TAGGED] => {
-            if reprIsImmediate(lhsRepr) || reprIsImmediate(rhsRepr) {
+            // JBUG cast to boolean
+            if <boolean>lhsRepr.isImmediate || <boolean>rhsRepr.isImmediate || <boolean>semTypeRepr(t:intersect(lhs.semType, rhs.semType)).isImmediate {
```
jclark: I suggest we work around this bug by adding a reprIsImmediate function that does `<boolean>repr.isImmediate`

compiler/modules/nback/scaffold.bal
```diff
@@ -502,8 +516,32 @@ function semTypeRepr(t:SemType ty) returns Repr {
...
+        boolean isImmediate = (w & ~(t:NIL|t:BOOLEAN|t:INT|t:STRING)) == 0 &&
+                              ((w & t:INT) == 0 || isIntConstrainedToImmediate(t:intSubtypeConstraints(ty))) &&
+                              ((w & t:STRING) == 0 || isStringSubtypeImmediate(t:stringSubtype(ty)));
+        TaggedRepr repr = { subtype: w, isImmediate };
         return repr;
     }
     panic error("unimplemented type (" + w.toHexString() + ")");
 }
+
+function isIntConstrainedToImmediate(t:IntSubtypeConstraints? c) returns boolean {
```
jclark: I would make this parallel to isStringSubtypeImmediate (i.e. put call to intSubtypeConstraints here).

I would call these intSubtypeAlwaysImmediate/stringSubtypeAlwaysImmediate.

compiler/modules/nback/scaffold.bal
```diff
@@ -502,8 +516,32 @@ function semTypeRepr(t:SemType ty) returns Repr {
     int supported = t:NIL|t:BOOLEAN|t:INT|t:FLOAT|t:DECIMAL|t:STRING|t:LIST|t:MAPPING|t:ERROR;
     int maximized = w | supported;
     if maximized == t:TOP || maximized == (t:NON_BEHAVIOURAL|t:ERROR) || (w & supported) == w {
-        TaggedRepr repr = { base: BASE_REPR_TAGGED, llvm: LLVM_TAGGED_PTR, subtype: w };
+        boolean isImmediate = (w & ~(t:NIL|t:BOOLEAN|t:INT|t:STRING)) == 0 &&
```
jclark: This would be much more understandable separated into function using if/else.

jclark (commented):
This looks nice. I particularly like using semTypeRepr in equal.bal.

compiler/modules/nback/scaffold.bal
manuranga: I did make it parallel, in the process I change the type of `intSubtypeConstraints`, I think this way makes more scene, this communicates that `intSubtypeConstraints` should only ever be called on a type that is has a subset that belongs to int. ie Tᴵᴺᵀ ≠ ∅

manuranga: I think it'll work even better if `intSubtype` returns `IntSubtype|true`

compiler/modules/nback/scaffold.bal
```diff
@@ -502,8 +528,46 @@ function semTypeRepr(t:SemType ty) returns Repr {
...
+        }
+    }
+    return true;
+}
+
+function isIntSubtypeAlwaysImmediate(t:IntSubtype|true  intSubtype) returns boolean {
+    return isIntConstrainedToImmediate(t:intSubtypeConstraints(intSubtype));
+}
+
+function isIntConstrainedToImmediate(t:IntSubtypeConstraints? c) returns boolean {
```
jclark, line 571: Don't see the point of making this a separate function.

compiler/modules/types/core.bal
```diff
@@ -874,12 +874,9 @@ public type IntSubtypeConstraints readonly & record {|
...
 |};
 
-// Returns `()` if `t` is not a proper, non-empty subtype of `int`.
-// i.e. returns `()` if `t` contains all or non of `int`.
-public function intSubtypeConstraints(SemType t) returns IntSubtypeConstraints? {
-    var intSubtype = intSubtype(t);
-    // JBUG can't flatten inner if-else
-    if intSubtype is boolean {
+// Returns `()` if `t` is an unconstrained int.
+public function intSubtypeConstraints(IntSubtype|true intSubtype) returns IntSubtypeConstraints? {
```
jclark: I don't think this is an improvement.

The point of IntSubtypeConstraints is to *hide* IntSubtype for the majority of cases that don't care about it.

Also your code using this has lots of awkward casts

jclark (changes requested):

compiler/modules/nback/scaffold.bal
manuranga: I am using it in equal.bal.
Alternative : add something like `boolean alwaysInImmediateRange;` to `IntRepr`. Even then I'll need both functions.

manuranga: I think this will help with #900, so let me do it that way.

PR #916, opened 2022-02-10
Optimize testing for int subtype that is a single range

### LINKED ISSUE #890: Optimize testing for int subtype that is a single range
Special case testing an int representation against an int subtype that is a single range.

ushirask:
Fixes #890

### DIFF
--- a/compiler/modules/nback/typeTest.bal
+++ b/compiler/modules/nback/typeTest.bal
@@ -95,7 +95,7 @@ function buildTypeTestedValue(llvm:Builder builder, Scaffold scaffold, bir:Regis
     }
     else if baseRepr == BASE_REPR_INT {
         t:IntSubtypeConstraints? intConstraints = t:intSubtypeConstraints(semType);
-        if intConstraints != () {
+        if intConstraints != () && intConstraints.all {
             llvm:Value hasType1 = builder.iCmp("sle", llvm:constInt(LLVM_INT, intConstraints.min), value);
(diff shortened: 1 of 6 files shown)


jclark (changes requested):
Looks good but we need some more tests.

- more ranges e.g. int:Signed32
- a range with gaps in it (i.e. `all` not true)
- type cast

compiler/modules/nback/typeTest.bal
```diff
@@ -94,7 +94,15 @@ function buildTypeTestedValue(llvm:Builder builder, Scaffold scaffold, bir:Regis
         }
     }
     else if baseRepr == BASE_REPR_INT {
-        hasType = <llvm:Value>builder.call(scaffold.getRuntimeFunctionDecl(typeContainsIntFunction), [scaffold.getTypeTest(<t:ComplexSemType>semType), value]);
+        t:IntSubtypeConstraints? intConstraints = t:intSubtypeConstraints(semType);
+        if intConstraints != () && intConstraints.all {
+            llvm:Value hasType1 = builder.iCmp("sle", llvm:constInt(LLVM_INT, intConstraints.min), value);
+            llvm:Value hasType2 = builder.iCmp("sge", llvm:constInt(LLVM_INT, intConstraints.max), value);
```
manuranga: instead hasType1 and hasType2, how about `lowerBoundMet` and `upperBoundMet`.

jclark: Do we really need the temp variables?

ushirask: I'll remove the temp variables

PR #925, opened 2022-02-14
Introduce mappingAtomicTypeMemberAt for mapping 

### LINKED ISSUE #914: Introduce function like listAtomicTypeMemberAt for mapping
Use here

https://github.com/ballerina-platform/nballerina/blob/06e1da907865142409bedf9e13dcb06ef9449830/compiler/modules/bir/verify.bal#L202

When we have an atomic type (e.g. in a mapping constructor) we should find the member type directly from the atomic type rather than using the complex projection algorithm.

ushirask:
Fixes #914

### DIFF
--- a/compiler/modules/front/expr.bal
+++ b/compiler/modules/front/expr.bal
@@ -847,6 +847,7 @@ function codeGenMappingConstructor(ExprContext cx, bir:BasicBlock bb, t:SemType?
     bir:Operand[] operands = [];
     string[] fieldNames = [];
     map<Position> fieldPos = {};
+    int i = 0;
     foreach s:Field f in expr.fields {
         string name = f.name;
         Position? prevPos = fieldPos[name];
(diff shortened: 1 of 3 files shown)


compiler/modules/front/expr.bal
```diff
@@ -847,6 +847,7 @@ function codeGenMappingConstructor(ExprContext cx, bir:BasicBlock bb, t:SemType?
     bir:Operand[] operands = [];
     string[] fieldNames = [];
     map<Position> fieldPos = {};
+    int i = 0;
```
jclark: Update to current main and then do this like I did for codeGenListConstructor/

compiler/modules/types/mapping.bal
```diff
@@ -9,6 +9,16 @@ public type MappingAtomicType readonly & record {|
     SemType rest;
 |};
 
+public function mappingAtomicTypeMemberAt(MappingAtomicType mat, string key, int i) returns SemType {
```
jclark: Get rid of `i` argument (?!)

Don't call the argument `key` since that's a keyword (use `k`).

compiler/modules/types/mapping.bal
```diff
@@ -9,6 +9,16 @@ public type MappingAtomicType readonly & record {|
...
+    int? n = mat.names.indexOf(key, 0);
```
jclark: Call this variable `i` of `index`

compiler/modules/types/mapping.bal
```diff
@@ -9,6 +9,16 @@ public type MappingAtomicType readonly & record {|
...
+    if n is int {
```
jclark: Use a conditional expressions

jclark (changes requested):

Issue #924, opened 2022-02-14
Add function to types module to compute projection for every index in one call


jclark:
This does a projection for all indices at once 

```
// For each [r, s] in the return value, for an index i in r, the projection for t[i] is r 
function listAllMemberTypes(Context cx, SemType t) returns [Range,SemType][] {}
```

Just iterating over every index is too inefficient for fixed length arrays (consider comparing `int[]` with `int[10000000]`).

jclark:
Then use this to finish comparableNillableList

heshanpadmasiri:
Fallowing tests that has been disabled with #927 must be re-enabled with this:
- 11-nested/order1-v.bal
- 11-nested/order2-v.bal
- 11-nested/order3-e.bal
- 11-nested/order3-v.bal
- 11-nested/order4-v.bal
- 11-nested/order5-v.bal

rdhananjaya:
I see two ways to approach this.
1. Identify all individual ranges in list type and then invoke `listProj` for each of those ranges.
2. Add an alternative version (duplicate) of `listProjExclude` that would return all range-semtype pairs for each path.

@jclark I prefer the first approach, any thoughts?

jclark:
My only thought is that 1 is only going to work if you generate a list of disjoint ranges from the ranges mentioned in the list type.

jclark:
I reworked #927.

In particular, I factored out the difficult logic into a combineRanges function.   You can probably reuse this for 1 in conjunction with listAtomicTypeAllMemberTypes.

You might want to change it to return two parallel arrays i.e. `[Range[], [int?,int?][]]`.

PR #927, opened 2022-02-15
Allow relational operators between comparable tuple types

### LINKED ISSUE #910: Allow relational operators between comparable tuple types
We need to allow comparison between e.g.

* `[int, string]` and `[int, string]`, or
* `[string, int]` and `[string, int...]`

heshanpadmasiri:
## Purpose
Allow relational operators between comparable tuple types
Resolves #910

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/types/core.bal
```diff
@@ -815,9 +815,59 @@ function comparableNillableList(Context cx, SemType t1, SemType t2) returns bool
...
+        }
+    }
+    else {
+        result = comparable(cx, listMemberType(cx, t1, INT), listMemberType(cx, t2, INT));
+    }
+    memo.comparable = result;
+    return result;
+}
+
+function updateComparableMemo(Context cx, ComparableMemo memo, SemType t1, SemType t2) returns boolean {
```
jclark: I don't understand what you are trying to achieve with this function. `comparable` memoizes.

compiler/modules/types/core.bal
```diff
@@ -815,9 +815,59 @@ function comparableNillableList(Context cx, SemType t1, SemType t2) returns bool
...
-    // SUBSET need to iterate members when tuples are supported
-    // following relies on the fact `listMemberType(cx, NIL, ()) = NEVER`
-    boolean result = comparable(cx, listMemberType(cx, t1, INT), listMemberType(cx, t2, INT));
+    boolean result;
+    var members1 = listAllMemberTypes(cx, t1);
+    var members2 = listAllMemberTypes(cx, t2);
+    // listAllMemberTypes return () if list type is not atomic
+    if members1 != () && members2 != () {
+        result = false;
+        int currentIndex = 0;
```
jclark: Use shorter names for currentIndex/currentMemberRange/currentMemberTy.

Also be consistent about abbreviating. It doesn't make sense to have a name where `current` is unabbreviated more `Type` is abbreviated to `Ty`, since `cur` is the better abbreviation.

compiler/modules/types/core.bal
```diff
@@ -815,9 +815,59 @@ function comparableNillableList(Context cx, SemType t1, SemType t2) returns bool
...
     ComparableMemo memo = { semType1: t1, semType2: t2 };
     cx.comparableMemo.add(memo);
-    // SUBSET need to iterate members when tuples are supported
-    // following relies on the fact `listMemberType(cx, NIL, ()) = NEVER`
-    boolean result = comparable(cx, listMemberType(cx, t1, INT), listMemberType(cx, t2, INT));
+    boolean result;
+    var members1 = listAllMemberTypes(cx, t1);
+    var members2 = listAllMemberTypes(cx, t2);
+    // listAllMemberTypes return () if list type is not atomic
+    if members1 != () && members2 != () {
```
jclark: This will be easier to understand if you factor out a function that merges the lists from the two calls to listAllMemberTypes into a `[Range,Semtype,Semtype][]`, then iterate over that.  Write a unit test for this merging function. Call something like mergeListMemberTypes.

jclark (changes requested):

compiler/modules/types/core.bal

compiler/modules/types/core.bal
heshanpadmasiri: function removed (no longer needed)

compiler/modules/types/core.bal
```diff
@@ -815,13 +815,67 @@ function comparableNillableList(Context cx, SemType t1, SemType t2) returns bool
...
 }
 
+function mergeListMemberTypes([Range,SemType][] list1, [Range,SemType][] list2) returns [Range, SemType, SemType][] {
+    int index1 = 0;
+    int index2 = 0;
+    int currentStart = 0;
+    var [rng1, ty1] = list1[index1];
+    var [rng2, ty2] = list2[index2];
+    [Range, SemType, SemType][] mergedMembers = [];
+    while index1 < list1.length() && index2 < list2.length() {
```
jclark: Here and below you are calling the `length` function on each iteration. Instead create local variables `len1` and `len2`.

compiler/modules/types/core.bal
```diff
@@ -815,13 +815,67 @@ function comparableNillableList(Context cx, SemType t1, SemType t2) returns bool
...
+        }
+    }
+    else {
+        result = comparable(cx, listMemberType(cx, t1, INT), listMemberType(cx, t2, INT));
+    }
     memo.comparable = result;
     return result;
 }
 
+function mergeListMemberTypes([Range,SemType][] list1, [Range,SemType][] list2) returns [Range, SemType, SemType][] {
```
jclark: Space after `Range,`

But perhaps we should define a type for `[Range,SemType][]` e.g. `MemberTypes`

compiler/modules/types/core.bal
```diff
@@ -815,13 +815,67 @@ function comparableNillableList(Context cx, SemType t1, SemType t2) returns bool
     }
     ComparableMemo memo = { semType1: t1, semType2: t2 };
     cx.comparableMemo.add(memo);
-    // SUBSET need to iterate members when tuples are supported
-    // following relies on the fact `listMemberType(cx, NIL, ()) = NEVER`
-    boolean result = comparable(cx, listMemberType(cx, t1, INT), listMemberType(cx, t2, INT));
+    boolean result;
+    var members1 = listAllMemberTypes(cx, t1);
+    var members2 = listAllMemberTypes(cx, t2);
+    // listAllMemberTypes return () if list type is not atomic
```
jclark: I suggest you panic for now if either list is non-atomic.

jclark (changes requested):
This looks much better.

jclark:
Do you have tests for recursive tuple types?

compiler/modules/types/core.bal

compiler/modules/types/core.bal

compiler/modules/types/core.bal

heshanpadmasiri:
> Do you have tests for recursive tuple types?

We have `14-tuple/comp14-e.bal`  and `14-tuple/comp6-e.bal`

compiler/modules/types/core.bal
```diff
@@ -915,6 +969,28 @@ public function listAtomicSimpleArrayMemberType(ListAtomicType? atomic) returns
...
+    int fixedLength = members.fixedLength;
+    int currentLength = 0;
+    foreach SemType initialType in members.initial {
+        memberTypes.push([{ min: currentLength, max: currentLength + 1 }, initialType]);
+        currentLength += 1;
+    }
+    if currentLength < fixedLength {
+        memberTypes.push([{ min: currentLength, max: fixedLength }, members.initial[members.initial.length() - 1]]);
+    }
+    memberTypes.push([{ min: fixedLength, max: int:MAX_VALUE }, atomicType.rest]);
```
jclark: Don't do this if rest is never.

jclark (changes requested):

compiler/modules/types/core.bal

compiler/modules/types/core.bal
```diff
@@ -815,13 +815,66 @@ function comparableNillableList(Context cx, SemType t1, SemType t2) returns bool
     }
     ComparableMemo memo = { semType1: t1, semType2: t2 };
     cx.comparableMemo.add(memo);
-    // SUBSET need to iterate members when tuples are supported
-    // following relies on the fact `listMemberType(cx, NIL, ()) = NEVER`
-    boolean result = comparable(cx, listMemberType(cx, t1, INT), listMemberType(cx, t2, INT));
+    boolean result;
+    if listAtomicTypeRw(cx, t1) == () || listAtomicTypeRw(cx, t2) == () {
```
jclark: Why are we using listAtomicTypeRw here? That is tested in listAllMemberTypes.

jclark (changes requested):

compiler/modules/types/core.bal
heshanpadmasiri: I am using it as a way to detect if the list type is not atomic (If this is not nil current placeholder function should work). When the proper version of `listAllMemberTypes` is implemented this whole if else part can be replaced with the body of else

jclark:
> I am using it as a way to detect if the list type is not atomic (If this is not nil current placeholder function should work). When the proper version of `listAllMemberTypes` is implemented this whole if else part can be replaced with the body of else  

We are supposed to be panicking if it is not atomic....

heshanpadmasiri:
> > I am using it as a way to detect if the list type is not atomic (If this is not nil current placeholder function should work). When the proper version of `listAllMemberTypes` is implemented this whole if else part can be replaced with the body of else
> 
> We are supposed to be panicking if it is not atomic....

Yes the `listAllMemberTypes` will panic but we have some tests (such as [`11-nested/order5-v.bal`](https://github.com/ballerina-platform/nballerina/blob/main/compiler/testSuite/11-nested/order5-v.bal) that have non atomic lists (not tuples) that will cause listAllMemberTypes to panic. May I disable those tests until #924 is done?

jclark:
Yes, they should be -fv tests, which #924 will change to -v tests.  Mention them in #924.

jclark (changes requested):
See comments.

heshanpadmasiri:
> Yes, they should be -fv tests, which #924 will change to -v tests. Mention them in #924.

fixed

compiler/modules/types/tests/coretest.bal
```diff
@@ -331,3 +331,23 @@ function testIntSubtypeConstraints() {
     test:assertEquals(c13.max, 15);
     test:assertEquals(c13.all, false);
 }
+
+@test:Config{}
+function testComparableNillableList() {
+    [Range, SemType][] members1 = [[{ min: 0, max: 10 }, INT], [{ min: 10, max: int:MAX_VALUE }, STRING]];
```
jclark, line 337: There's a confusion here: max in a Range in **inclusive** (maximum always implies inclusive).

PR #930, opened 2022-02-17
Parse inline fixed length array type descriptors

### LINKED ISSUE #913: Parse inline fixed length array type descriptors
Need to be able to parse:

```
type Int int;

function foo() {
  Int[2] n = [0, 1];
}
```

ushirask:
Fixes #913


compiler/modules/front.syntax/preparse.bal
```diff
@@ -89,8 +89,16 @@ function preparseArrayTypeDesc(Tokenizer tok) returns boolean|err:Syntax {
         check tok.advance();
         _ = check tok.expectIdentifier();
     }
-    check tok.expect("[");
-
-    // SUBSET fixed length array types / tuples are not supported, is a td only if no token between `[` `]`
-    return tok.current() == "]";
+    while tok.current() == "[" {
```
jclark: This doesn't look like it's on the right track. I would expect something more like preparseBracketed.

compiler/modules/front.syntax/preparse.bal
```diff
@@ -89,8 +89,16 @@ function preparseArrayTypeDesc(Tokenizer tok) returns boolean|err:Syntax {
...
+        check tok.advance();
+        if tok.current() == "]" {
+            return true;
+        }
+        SimpleConstExpr|err:Syntax expr = parseSimpleConstExpr(tok);
```
jclark: We shouldn't be using parseXXX in a preparse.

jclark (changes requested):

compiler/modules/front.syntax/preparse.bal
```diff
@@ -90,7 +90,40 @@ function preparseArrayTypeDesc(Tokenizer tok) returns boolean|err:Syntax {
         _ = check tok.expectIdentifier();
     }
     check tok.expect("[");
-
-    // SUBSET fixed length array types / tuples are not supported, is a td only if no token between `[` `]`
-    return tok.current() == "]";
+    while true {
```
jclark: I don't understand the logic here. Please explain in detail.

Why is there a `while` here?

jclark (changes requested):

compiler/modules/front.syntax/preparse.bal
ushirask: When there are multiple array dimensions such as `T[2][2]` we would have to iterate through expression right? Here after the first `[2]` since the next token is `[` we will continue the loop.

compiler/modules/front.syntax/preparse.bal
```diff
@@ -16,15 +16,16 @@ final readonly & map<CLOSE_BRACKET> closeBracketMap = {
...
     return preparseBracketedTypeDesc(tok, ")");
 }
 
 function preparseTupleTypeDesc(Tokenizer tok) returns boolean|err:Syntax {
+    check tok.expect("[");
     return preparseBracketedTypeDesc(tok, "]");
 }
 
 function preparseBracketedTypeDesc(Tokenizer tok, CLOSE_BRACKET close) returns boolean|err:Syntax {
-    check tok.advance();
```
jclark: I am not convinced we need to change this.

compiler/modules/front.syntax/preparse.bal
```diff
@@ -41,7 +42,7 @@ function preparseBracketedTypeDesc(Tokenizer tok, CLOSE_BRACKET close) returns b
     if t == () {
         return tok.err("incomplete statement");
     }
-    return t != ".";
+    return t is [IDENTIFIER, string] || t == "_" || t == "|" || t == "&" || t =="?";
```
jclark: Why can't we keep this as a negative test?

compiler/modules/front.syntax/preparse.bal
jclark: Note precondition for preparseArrayTypeDesc

jclark (changes requested):

compiler/modules/front.syntax/preparse.bal
ushirask: For a negative test we'll also have to consider `=` and compound assignment operators.

jclark (changes requested):
> For a negative test we'll also have to consider `=` and compound assignment operators. 

Yes, we would.  Is there some reason you cannot do this? We have a CompoundAssignDelim defined....

ushirask:
> > For a negative test we'll also have to consider `=` and compound assignment operators.
> 
> Yes, we would. Is there some reason you cannot do this? We have a CompoundAssignDelim defined....

Changed it.

compiler/modules/front.syntax/tests/data/testParser.json
```diff
@@ -390,12 +390,34 @@
     ["V", "stmt", "decimal d = 1E+12d;"],
     ["V", "stmt", "byte i = 10;"],
     ["V", "stmt", "byte[] i = [10];"],
+    ["V", "stmt", "T[2] n = [2, 1.5];"],
```
jclark, line 393: These are all good. But we should have some more tests for things that should *not* be parsed as type descriptors.

```
v[x] = 1;
v[x] += 1;
v[x].foo();
v[2][3].foo();
v[1][n] = bar();
int[x].foo();
int[1] += foo();
```

jclark (changes requested):
Also add some end to end tests for fixed length tuple types: v, e and p types. (You can make a separate issue and do a follow-on PR if you prefer, but this needs doing in this milestone).

I don't think there is anything else needed but we need to test.

compiler/modules/front.syntax/tests/data/testParser.json
ushirask: Shall we modify our `testPreparse` for testing `preparseArrayTypeDesc` or just add these as v tests?

jclark: I don't remember why we need testPreparse. A couple of these cannot be v tests.

jclark: I will merge this now.  You can follow up with more tests.

PR #931, opened 2022-02-17
Implement byte[] as array of bytes

### LINKED ISSUE #169: Implement `byte[]` as array of bytes
This  implements `byte[]` using a specialized representation of arrays that represents the array as an array of bytes.

We have to deal with the fact that an `byte[] <: any[]`, so code dealing with `any[]` will have to deal with the (unlikely possibility) that it gets an array of bytes rather than a tagged pointer. (This is now handled using exactness.)

This depends on #168.

manuranga:
Fixes #169

### DIFF
--- a/compiler/modules/nback/structure.bal
+++ b/compiler/modules/nback/structure.bal
@@ -30,8 +30,17 @@ final RuntimeFunction mappingIndexedSetFunction = {
     attrs: []
 };
 
-final RuntimeFunction listConstructFunction = {
-    name: "list_construct",
+final RuntimeFunction listConstruct8Function = {
+    name: "list_construct_8",
+    ty: {
+        returnType: heapPointerType(llListType),
+        paramTypes: [llvm:pointerType(llStructureDescType), LLVM_INT]
+    },
+    attrs: []
+};
+
+final RuntimeFunction listConstruct1Function = {
+    name: "list_construct_1",
     ty: {
         returnType: heapPointerType(llListType),
         paramTypes: [llvm:pointerType(llStructureDescType), LLVM_INT]
@@ -115,10 +124,12 @@ const LLVM_INDEX = "i32";
 
 type ListRepr readonly & object {
     llvm:Type memberType;
+    llvm:Type memberHeapType;
     int listDescGetIndex;
     int listDescSetIndex;
     int listDescInexactSetIndex;
     boolean isSpecialized;
+    RuntimeFunction construct;
     function buildMember(llvm:Builder builder, Scaffold scaffold, bir:Operand member, t:SemType memberType) returns llvm:Value|BuildError;
     function buildMemberStore(llvm:Builder builder, Scaffold scaffold, llvm:Value value, bir:Register reg);
 };
@@ -130,6 +141,7 @@ const LIST_DESC_FIRST_FUNCTION_INDEX = 3;
 final readonly & map<ListRepr> listReprs = {
     generic: object {
         llvm:Type memberType = LLVM_TAGGED_PTR;
+        llvm:Type memberHeapType = LLVM_TAGGED_PTR;
         int listDescGetIndex = LIST_DESC_FIRST_FUNCTION_INDEX;
         int listDescSetIndex = LIST_DESC_FIRST_FUNCTION_INDEX + 1;
         int listDescInexactSetIndex = LIST_DESC_FIRST_FUNCTION_INDEX + 2;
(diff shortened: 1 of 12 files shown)


runtime/list.c
```diff
@@ -21,13 +21,20 @@ static inline bool listDescAllowsTaggedAt(ListDescPtr ldp, int64_t index, Tagged
     return memberTypeContainsTagged(mt, val);
 }
 
-ListPtr _bal_list_construct(ListDescPtr desc, int64_t capacity) {
+ListPtr _bal_list_generic_construct(ListDescPtr desc, int64_t capacity) {
```
jclark: `_generic_` isn't really the right word in this. You use it for int arrays also, right? It's for anything that has members of size 8.

There's a bit of a mismatch between `generic` as used in the ListDesc (which means any list) and `generic` in these `list_generic` functions, where it means tagged ptr.

runtime/list.c
```diff
@@ -2,7 +2,7 @@
 
 #include <string.h>
 
-#define ARRAY_LENGTH_MAX ((int64_t)(INT64_MAX/sizeof(TaggedPtr)))
+#define GENERIC_ARRAY_LENGTH_MAX ((int64_t)(INT64_MAX/sizeof(TaggedPtr)))
```
jclark: Leave this as is. We are enforcing it for all arrays.

compiler/modules/nback/structure.bal
```diff
@@ -314,14 +345,27 @@ function buildSpecializedListSet(llvm:Builder builder, Scaffold scaffold, llvm:V
...
-    builder.store(val, builder.getElementPtr(array, [llvm:constInt(LLVM_INT, 0), index], "inbounds"));
+    array = builder.bitCast(array, heapPointerType(llvm:arrayType(repr.memberHeapType, 0)));
+    llvm:Value heapVal = convertValue(builder, val, repr.memberType, repr.memberHeapType);
+    builder.store(heapVal, builder.getElementPtr(array, [llvm:constInt(LLVM_INT, 0), index], "inbounds"));
     builder.br(bbJoin);
     builder.positionAtEnd(bbTaggedSet);
     return bbJoin;
 }
 
+function convertValue(llvm:Builder builder, llvm:Value value, llvm:Type fromTy, llvm:Type toTy) returns llvm:Value {
```
jclark, line 380: I think it's better to have two functions listReprConvertTo/FromHeapType(ListRepr, Builder, Value)

jclark (changes requested):

runtime/list.c
manuranga: Ya, I was considering `_bal_list_construct_64` and `_bal_list_construct_8`.
`generic` in `GenericArray` means "not specific", and `generic` in `list_generic` means tagged.

runtime/list.c
manuranga: I meant to increase the max size just for bytes up to INT64_MAX. Now I see I haven't done it correctly (should have changed `_bal_array_grow`), I'll revert this for now. 

But shouldn't we increase the max size for bytes?

runtime/list.c
jclark: I think on the C side, it will make more sense to do this in terms of bytes (i.e. sizeof) rather than bits, so _8 and _1.

I think we should probably change list_generic to list_tagged.  But let's do that in a followup PR.

runtime/list.c
jclark: I see no point in increasing it. Just useless extra complexity.

We cannot go up to INT64_MAX anyway since we are using top 8 bits in pointers as a tag.

runtime/list.c
```diff
@@ -55,13 +64,23 @@ TaggedPtr _bal_list_int_array_get_tagged(TaggedPtr p, int64_t index) {
 }
 
 int64_t _bal_list_int_array_get_int(TaggedPtr p, int64_t index) {
-    // SUBSET should only be called when subset supports int subtype arrays
-    fprintf(stderr, "warning: access int array via int type desc func pointer\n");
```
jclark: Why is this needed for this PR?

compiler/modules/nback/structure.bal
```diff
@@ -31,7 +31,16 @@ final RuntimeFunction mappingIndexedSetFunction = {
 };
 
 final RuntimeFunction listConstructFunction = {
```
jclark: This needs to be listConstruct8Function to be consistent with existing naming patterns

compiler/modules/nback/structure.bal
```diff
@@ -31,7 +31,16 @@ final RuntimeFunction mappingIndexedSetFunction = {
...
+    name: "list_construct_8",
+    ty: {
+        returnType: heapPointerType(llListType),
+        paramTypes: [llvm:pointerType(llStructureDescType), LLVM_INT]
+    },
+    attrs: []
+};
+
+final RuntimeFunction byteListConstructFunction = {
+    name: "list_construct_1",
```
jclark, line 43: listConstruct1Function or change the C function name...

jclark (changes requested):
Looks good.

runtime/list.c
```diff
@@ -189,8 +210,7 @@ PanicCode _bal_list_generic_set_int(TaggedPtr p, int64_t index, int64_t val) {
 }
 
 PanicCode _bal_list_int_array_set_tagged(TaggedPtr p, int64_t index, TaggedPtr val) {
-    ListPtr lp = taggedToPtr(p);
-    if (!listDescAllowsTaggedAt(lp->desc, index, val)) {
+    if ((getTag(val) & UT_MASK) != TAG_INT) {
```
jclark, line 213: Shouldn't we have aa similar change to float_array_set_tagged?

PR #942, opened 2022-02-22
Implement `listAllMemberTypes` for any list

### LINKED ISSUE #924: Add function to types module to compute projection for every index in one call

rdhananjaya:
## Purpose
Fixes https://github.com/ballerina-platform/nballerina/issues/924

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/types/listProj.bal
```diff
@@ -27,8 +27,38 @@ function listProjBdd(Context cx, IntSubtype|true k, Bdd b, Conjunction? pos, Con
     }
 }
 
+public function listAllMemberTypes(Context cx, SemType t) returns [Range[], SemType[]] {
+    if t is UniformTypeBitSet {
+        return (t & LIST) != 0 ? [[{ min: 0, max: int:MAX_VALUE }], [TOP]] : [[{ min: 0, max: int:MAX_VALUE }], [NEVER]];
```
rdhananjaya, line 32: I would like to extract variables positive int range with TOP and positive int range with NEVER, I just can't think of a name for them.

compiler/modules/types/list.bal
```diff
@@ -34,12 +34,25 @@ public function listAtomicTypeMemberAt(ListAtomicType atomic, int i) returns Sem
...
-    SemType[] initial = atomicType.members.initial;
+    foreach var [r, i1, i2] in combineRanges(r1, r2) {
+        ranges.push(r);
+        types.push(union(i1 == () ? NEVER : t1[i1],
+                         i2 == () ? NEVER : t2[i2]));
+    }
+    return [ranges, types];
+}
+
+public function listAtomicTypeAllMemberTypes(FixedLengthArray fixedArray, SemType rest) returns ListMemberTypes {
```
jclark: This signature change is not OK: the function name starts with `listAtomicType` because it operates on a ListAtomicType. No public functions take separated FixedLengthArray/SemType arguments.

compiler/modules/types/listProj.bal
```diff
@@ -27,8 +27,41 @@ function listProjBdd(Context cx, IntSubtype|true k, Bdd b, Conjunction? pos, Con
     }
 }
 
+final ListMemberTypes LIST_MEMBER_TYPES_ALL_TOP = [[{ min: 0, max: int:MAX_VALUE }], [TOP]];
+final ListMemberTypes LIST_MEMBER_TYPES_ALL_NEVER = [[{ min: 0, max: int:MAX_VALUE }], [NEVER]];
+
+public function listAllMemberTypes(Context cx, SemType t) returns ListMemberTypes {
```
jclark: What is this doing in listProj.bal? Please put it back where it was in core.bal.

jclark (changes requested):

compiler/modules/types/listProj.bal
rdhananjaya: Isn't this a counterpart of `listProj` in `listProj.bal:4`?

compiler/modules/types/core.bal
```diff
@@ -905,14 +905,13 @@ public function listAtomicSimpleArrayMemberType(ListAtomicType? atomic) returns
 
 final ListAtomicType LIST_ATOMIC_TOP = { members: { initial: [], fixedLength: 0 }, rest: TOP };
 
-// placeholder for #924
 public function listAllMemberTypes(Context cx, SemType t) returns ListMemberTypes {
-    ListAtomicType? atomicType = listAtomicTypeRw(cx, t);
-    if atomicType == () {
-        panic error("expected atomic list type");
+    if t is UniformTypeBitSet {
+        return (t & LIST) != 0 ? LIST_MEMBER_TYPES_ALL_TOP : LIST_MEMBER_TYPES_ALL_NEVER;
```
jclark: This isn't right for LIST_RO: rest needs to be just readonly for that case.

compiler/modules/types/list.bal
```diff
@@ -22,6 +22,21 @@ type ListConjunction record {|
     ListConjunction? next;
 |};
 
+public type ListMemberTypes [Range[], SemType[]];
+
+function listMemberTypesUnion(ListMemberTypes mt1, ListMemberTypes mt2) returns ListMemberTypes {
```
jclark: Since this is non-public, it should be further down the file (just before combineRanges).

compiler/modules/types/listProj.bal
```diff
@@ -27,8 +27,31 @@ function listProjBdd(Context cx, IntSubtype|true k, Bdd b, Conjunction? pos, Con
     }
 }
 
+final ListMemberTypes LIST_MEMBER_TYPES_ALL_TOP = [[{ min: 0, max: int:MAX_VALUE }], [TOP]];
```
jclark: These need to be readonly

compiler/modules/types/list.bal
```diff
@@ -22,6 +22,21 @@ type ListConjunction record {|
     ListConjunction? next;
 |};
 
+public type ListMemberTypes [Range[], SemType[]];
```
jclark, line 26: We should maintain the invariant that the SemType array here does not have entries == NEVER.  Please add a comment to that effect.

compiler/modules/types/listProj.bal
```diff
@@ -27,8 +27,31 @@ function listProjBdd(Context cx, IntSubtype|true k, Bdd b, Conjunction? pos, Con
...
+final ListMemberTypes LIST_MEMBER_TYPES_ALL_NEVER = [[{ min: 0, max: int:MAX_VALUE }], [NEVER]];
```
jclark: This should just be an empty list.

compiler/modules/types/listProj.bal
```diff
@@ -110,3 +133,48 @@ function listProjExclude(Context cx, IntSubtype|true k, FixedLengthArray members
         return p;
     }
 }
+
+function listProjAll(Context cx, Conjunction? pos, Conjunction? neg) returns ListMemberTypes {
+    SemType|[FixedLengthArray, SemType] intersection = listProjIntersectConjunctions(cx, pos, neg);
+    if intersection is SemType {
+        return [[{ min: 0, max: int:MAX_VALUE }], [NEVER]];
```
jclark: Isn't this LIST_MEMBER_TYPES_ALL_NEVER

compiler/modules/types/listProj.bal
jclark: I would call these LIST_MEMBER_TYPES_ALL and LIST_MEMBER_TYPES_NONE

compiler/modules/types/listProj.bal
```diff
@@ -27,8 +27,31 @@ function listProjBdd(Context cx, IntSubtype|true k, Bdd b, Conjunction? pos, Con
     }
 }
 
+final ListMemberTypes LIST_MEMBER_TYPES_ALL_TOP = [[{ min: 0, max: int:MAX_VALUE }], [TOP]];
+final ListMemberTypes LIST_MEMBER_TYPES_ALL_NEVER = [[{ min: 0, max: int:MAX_VALUE }], [NEVER]];
+
+function listProjBddAllKeys(Context cx, Bdd b, Conjunction? pos, Conjunction? neg) returns ListMemberTypes {
```
jclark: The `Keys` in the name isn't consistent with other names.

compiler/modules/types/listProj.bal
```diff
@@ -27,8 +27,31 @@ function listProjBdd(Context cx, IntSubtype|true k, Bdd b, Conjunction? pos, Con
...
 function listProjPath(Context cx, IntSubtype|true k, Conjunction? pos, Conjunction? neg) returns SemType {
+    SemType|[FixedLengthArray, SemType] intersection = listProjIntersectConjunctions(cx, pos, neg);
+    if intersection is SemType {
+        return intersection;
+    }
+    var [members, rest] = intersection;
+    return listProjExclude(cx, k, members, rest, listConjunction(cx, neg));
+}
+
+function listProjIntersectConjunctions(Context cx, Conjunction? pos, Conjunction? neg) returns SemType|[FixedLengthArray, SemType] {
```
jclark: Why are we returning a union here?

compiler/modules/types/listProj.bal
```diff
@@ -110,3 +133,48 @@ function listProjExclude(Context cx, IntSubtype|true k, FixedLengthArray members
...
+function listProjAll(Context cx, Conjunction? pos, Conjunction? neg) returns ListMemberTypes {
+    SemType|[FixedLengthArray, SemType] intersection = listProjIntersectConjunctions(cx, pos, neg);
+    if intersection is SemType {
+        return [[{ min: 0, max: int:MAX_VALUE }], [NEVER]];
+    }
+    var [members, rest] = intersection;
+    return listProjPathAll(cx, members, rest, listConjunction(cx, neg));
+}
+
+function listProjPathAll(Context cx, FixedLengthArray members, SemType rest, ListConjunction? neg) returns ListMemberTypes {
```
jclark: I don't understand why we need so much new code.

jclark:
This all seems too complicated to me.  I would approach it by generating a list of ranges from the Bdd, such that if any two integers are both in the same range (or both not in any range), the projection will be the same for both integers. I think you can do this easily with combineRanges. Then use listProj on one integer in each range.

compiler/modules/types/list.bal
```diff
@@ -562,6 +574,11 @@ function combineRanges(Range[] ranges1, Range[] ranges2) returns [Range, int?, i
     return combined;
 }
 
+function combineRanges2(Range[] range1, Range[] range2) returns Range[] {
```
rdhananjaya: I'm having trouble finding a name for this function.

rdhananjaya: What if we define this function as an anonymous function where we use this?

compiler/modules/types/list.bal
```diff
@@ -22,6 +22,10 @@ type ListConjunction record {|
     ListConjunction? next;
 |};
 
+// Pair of empty Range, SemType arrays represent NEVER at every member index.
```
jclark: I do not understand this comment.

PR #943, opened 2022-02-23
Clean up anomalous uses of tmp registers

### LINKED ISSUE #941: Clean up anomalous uses of tmp registers
Currently we assign to tempRegisters "internally" in the following places

1. https://github.com/ballerina-platform/nballerina/blob/0005efda65f580233f63a8485cd5ad774e5feb4d/compiler/modules/front/stmt.bal#L1161
2. https://github.com/ballerina-platform/nballerina/blob/0005efda65f580233f63a8485cd5ad774e5feb4d/compiler/modules/front/expr.bal#L312
3. https://github.com/ballerina-platform/nballerina/blob/0005efda65f580233f63a8485cd5ad774e5feb4d/compiler/modules/front/expr.bal#L317
4. https://github.com/ballerina-platform/nballerina/blob/0005efda65f580233f63a8485cd5ad774e5feb4d/compiler/modules/front/expr.bal#L696
5. https://github.com/ballerina-platform/nballerina/blob/0005efda65f580233f63a8485cd5ad774e5feb4d/compiler/modules/front/expr.bal#L698

_Originally posted by @heshanpadmasiri in https://github.com/ballerina-platform/nballerina/pull/939#discussion_r811546704_

heshanpadmasiri:
## Purpose
Clean up anomalous uses of tmp registers
Resolves to #941

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/bir/bir.bal
+++ b/compiler/modules/bir/bir.bal
@@ -152,6 +152,7 @@ public type RegisterBase record {|
 |};
 
 public type Register TmpRegister|AssignTmpRegister|ParamRegister|VarRegister|FinalRegister|NarrowRegister;
+public type BindableRegister ParamRegister|FinalRegister|VarRegister|NarrowRegister;
 public type DeclRegisterKind PARAM_REGISTER_KIND|VAR_REGISTER_KIND|FINAL_REGISTER_KIND;
 
 public type DeclRegister record {|
@@ -173,6 +174,8 @@ public type AssignTmpRegister readonly & record {|
 
 public type NarrowRegister readonly & record {|
     *RegisterBase;
+    // position of the register that was narrowed
+    Position prevPos;
     NARRROW_REGISTER_KIND kind;
 |};
 
@@ -203,8 +206,10 @@ public function createFinalRegister(FunctionCode code, SemType semType, string n
     return r;
 }
 
-public function createNarrrowRegister(FunctionCode code, SemType semType, string? name = (), Position? pos = ()) returns NarrowRegister {
-    NarrowRegister r = { number: code.registers.length(), semType, name, pos, kind: NARRROW_REGISTER_KIND };
+public function createNarrrowRegister(FunctionCode code, SemType semType, Register prev, string? name = (), Position? pos = ()) returns NarrowRegister {
+    // for debug information we at least need the position
+    Position prevPos = prev is NarrowRegister ? prev.prevPos : <Position> prev.pos;
+    NarrowRegister r = { number: code.registers.length(), semType, name, pos, kind: NARRROW_REGISTER_KIND, prevPos };
     code.registers.push(r);
     return r;
 }
--- a/compiler/modules/bir/bir.bal
+++ b/compiler/modules/bir/bir.bal
@@ -137,7 +137,8 @@ public enum RegisterKind {
     VAR_REGISTER_KIND,
     FINAL_REGISTER_KIND,
     NARRROW_REGISTER_KIND,
-    TMP_REGISTER_KIND
+    TMP_REGISTER_KIND,
+    ASSIGN_TMP_REGISTER_KIND
 }
 
 public type RegisterBase record {|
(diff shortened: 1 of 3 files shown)


compiler/modules/front/stmt.bal
```diff
@@ -1158,7 +1158,17 @@ function codeGenExprForCond(StmtContext cx, bir:BasicBlock bb, Environment env,
     }
     else if flags != 0 {
         bir:Register reg = cx.createTmpRegister(t:BOOLEAN);
-        bir:AssignInsn insn = { result: reg, operand: { value, semType: t:BOOLEAN }, pos: expr.startPos };
+        bir:EqualityOp op;
+        bir:Operand equalityOperand;
+        if value {
+            equalityOperand = { value: false, semType: t:BOOLEAN };
```
jclark: I don't understand what you are doing here. Why does it need to be false here and true in the other branch?

Shouldn't this use t:singleton?

jclark (changes requested):

compiler/modules/front/stmt.bal
heshanpadmasiri: I think I misunderstood what you said in the [comment](https://github.com/ballerina-platform/nballerina/issues/941#issuecomment-1048373596). Fixed

compiler/modules/front/stmt.bal
```diff
@@ -1158,7 +1158,12 @@ function codeGenExprForCond(StmtContext cx, bir:BasicBlock bb, Environment env,
     }
     else if flags != 0 {
         bir:Register reg = cx.createTmpRegister(t:BOOLEAN);
-        bir:AssignInsn insn = { result: reg, operand: { value, semType: t:BOOLEAN }, pos: expr.startPos };
+        bir:EqualityInsn insn = {
+            op: "===",
```
jclark: This is violating the rule that the intersection of the types of the operands of === must not be disjoint.

You need to use `===` and `!==` ops with the equal operands, as I suggested.

jclark (changes requested):

compiler/modules/front/stmt.bal

heshanpadmasiri:
Replaced `TmpRegister` with `AssignTmpRegister` in `AssignInsn`

compiler/modules/front/stmt.bal
```diff
@@ -911,7 +911,7 @@ function lookupVarRefForAssign(StmtContext cx, Environment env, string varName,
 
 function codeGenAssign(StmtContext cx, Environment env, bir:BasicBlock block, bir:Register result, s:Expr expr, t:SemType semType, Position pos) returns CodeGenError|bir:BasicBlock {
     var { result: operand, block: nextBlock } = check cx.codeGenExpr(block, env, semType, expr);
-    if result !is bir:TmpRegister|bir:VarRegister|bir:FinalRegister {
+    if result !is bir:AssignTmpRegister|bir:VarRegister|bir:FinalRegister {
```
jclark: We should be passing in the right kind of register here not casting. This means more changes are still needed. I guess Binding needs changing to use the right kind of register.

jclark (changes requested):

compiler/modules/front/stmt.bal
heshanpadmasiri: We currently use bindings with register to which you can't assign values at 2 places

1. https://github.com/ballerina-platform/nballerina/blob/1db49a0f695ac02032f4e8556e80f212dd6ae89e/compiler/modules/front/stmt.bal#L226
2. https://github.com/ballerina-platform/nballerina/blob/1db49a0f695ac02032f4e8556e80f212dd6ae89e/compiler/modules/front/stmt.bal#L801

If we are to keep allowing this we can also check for register type in [`lookupVarRefForAssign`](https://github.com/ballerina-platform/nballerina/blob/1db49a0f695ac02032f4e8556e80f212dd6ae89e/compiler/modules/front/stmt.bal#L890) and return the correct type (or panic if it is not the correct type)

jclark: The binding will need to allow for
- params
- final variables
- narrowed variables
- normal variables

Only the last one can be assigned to.

The first two correspond to errors in the source: so these cases should be rejected during the lookup as a semantic error. In the third case, the assignment will cause the variable not be narrowed.  There will be a different type for the original binding, which won't allowed for narrowed registers.

There's no need to explicitly check and panic: just do the cast and put a comment explaining why the cast must succeed.

compiler/modules/bir/bir.bal
```diff
@@ -165,8 +167,15 @@ public type TmpRegister readonly & record {|
...
 
+public type AssignTmpRegister readonly & record {|
+    *RegisterBase;
+    ASSIGN_TMP_REGISTER_KIND kind;
+|};
+
 public type NarrowRegister readonly & record {|
     *RegisterBase;
+    // position of the register that was narrowed
+    Position prevPos;
```
jclark, line 178: We need the number of the register not its position.

PR #945, opened 2022-02-25
Implement fixed length arrays

### LINKED ISSUE #944: Implement fixed length array types
Only know thing needed to make fixed length array types work is #913.  But we need to check that there is nothing else.

Need -p, -v and -e tests.

ushirask:
Fixes #944 and #913


compiler/testSuite/14-array/fixedlength7-fe.bal
```diff
@@ -0,0 +1,3 @@
+public function main() {
+    int[true] v = [1, 2]; // @error
```
ushirask: @jclark An error is thrown at line 1 instead of 2, with `error: initializer of 'main' is not a subtype of the declared type`

jclark: Create a bug report then.

jclark (changes requested):
If there are things that need fixing to make this work properly, then this PR needs to do them.

There's one bug, plus one test that is renamed to -fv.

compiler/testSuite/14-array/fixedlength7-fe.bal
jclark: If the problem is related to fixed length tuples, then fix it here.

docs/subset14.md
```diff
@@ -112,7 +112,7 @@ postfix-type-desc =
 
 optional-type-desc = postfix-type-desc "?"
 
-array-type-desc = postfix-type-desc "[" "]"
+array-type-desc = postfix-type-desc "[" [simple-const-expr] "]"
```
jclark: This isn't enough: look at 2022R2 spec.

compiler/testSuite/14-array/fixedlength1-e.bal
```diff
@@ -0,0 +1,6 @@
+import ballerina/io;
+
+public function main() {
+    int[4] v = [1, 2, 3, 4];
+    io:println(v[4]); // @error
```
ushirask, line 5: Error message is `error: type of member access is never`

compiler/testSuite/14-array/fixedlength2-e.bal
```diff
@@ -0,0 +1,4 @@
+public function main() {
+    int[4] v = [1, 2, 3, 4];
+    v[4] = 4; // @error
```
ushirask, line 3: Error message is `error: value assigned to member of list is not a subtype of array member type`

ushirask (commented):
Some errors are giving unclear messages

compiler/testSuite/14-array/fixedlength1-p.bal
```diff
@@ -0,0 +1,4 @@
+public function main() {
+    int[4] v = [1, 2];
+    v[3] = 4; // @panic no filler value
```
ushirask, line 3: Should add filler values

jclark: Can you please open a separate issue for this?

ushirask: Added in #946

compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -536,6 +519,31 @@ function parseSimpleConstExpr(Tokenizer tok) returns SimpleConstExpr|err:Syntax
             LiteralExpr expr = {  startPos, endPos, value: t == "true" };
             return expr;
         }
+    }
+    return parseArrayLenExpr(tok);
+}
+
+function parseArrayLenExpr(Tokenizer tok) returns SimpleConstExpr|err:Syntax {
```
jclark: parseArrayLengthExpr

compiler/modules/bir/verify.bal
```diff
@@ -184,7 +184,11 @@ function verifyListConstruct(VerifyContext vc, ListConstructInsn insn) returns E
     }
     Operand[] operands = insn.operands;
     foreach int i in 0 ..< operands.length() {
-        check verifyOperandType(vc, operands[i], t:listAtomicTypeMemberAt(lat, i), "type of list constructor member is not allowed by the list type", insn.pos);
+        SemType st = t:listAtomicTypeMemberAt(lat, i);
```
jclark: Let's deal with the unhelpful error message in a separate PR.

It's better to leave this as is and deal with this expr.bal: we can give a proper location as we did with function arguments.

compiler/modules/front/const.bal
```diff
@@ -26,11 +26,14 @@ function resolveConstDefn(ModuleSymbols mod, s:ConstDefn defn) returns s:Resolve
...
 }
 
-function resolveConstExpr(ModuleSymbols mod, s:ModuleLevelDefn defn, s:Expr expr, t:SemType? expectedType) returns s:ResolvedConst|ResolveTypeError {
+function resolveConstExpr(ModuleSymbols mod, s:ModuleLevelDefn defn, s:Expr expr, t:SemType? expectedType, boolean td = false) returns s:ResolvedConst|ResolveTypeError {
     ExprContext cx = new ExprContext(mod, defn, constCode, constEnvironment, ());
     var { result } = check codeGenExpr(cx, constBasicBlock, expectedType, expr);
     bir:ConstOperand operand = <bir:ConstOperand>result;
     if expectedType != () && !t:isSubtype(mod.tc, operand.semType, expectedType) {
+        if td && expr is s:VarRefExpr {
+            return err:semantic(`initializer ${expr.name} is not an int`, s:locationInDefn(defn, expr.startPos));
```
jclark: Don't understand this at all. How do you know that `int` is expected here?

jclark (changes requested):

compiler/modules/front/const.bal
ushirask: `td` is true only with `resolveConstIntExpr` so it will be an int, maybe we could propagate the type/ error message and use that here instead?

compiler/modules/front/const.bal
```diff
@@ -26,11 +26,14 @@ function resolveConstDefn(ModuleSymbols mod, s:ConstDefn defn) returns s:Resolve
...
 }
 
-function resolveConstExpr(ModuleSymbols mod, s:ModuleLevelDefn defn, s:Expr expr, t:SemType? expectedType) returns s:ResolvedConst|ResolveTypeError {
+function resolveConstExpr(ModuleSymbols mod, s:ModuleLevelDefn defn, s:Expr expr, t:SemType? expectedType, string? td = ()) returns s:ResolvedConst|ResolveTypeError {
     ExprContext cx = new ExprContext(mod, defn, constCode, constEnvironment, ());
     var { result } = check codeGenExpr(cx, constBasicBlock, expectedType, expr);
     bir:ConstOperand operand = <bir:ConstOperand>result;
     if expectedType != () && !t:isSubtype(mod.tc, operand.semType, expectedType) {
+        if td != () && expr is s:VarRefExpr {
+            return err:semantic(`initializer ${expr.name} is not ${td}`, s:locationInDefn(defn, expr.startPos));
```
jclark: This still doesn't seem good to me: 

Why is the test for VarRefExpr right?  I guess you are using that so that you can expr.name. But that's not really the right thing to test.

We are calling resolveConstExpr in the following cases

1. RHS of const definition
2. In a match pattern in a match stmt
3. In a type desc for a singleton type
4. In an array length in an array type desc

The current error message is really only right for case (1).

What has VarRefExpr got to do with any of these?

Please propose a better design. We can discuss in today's call.

jclark (changes requested):

PR #950, opened 2022-03-02
Add filler values to tuples

### LINKED ISSUE #946: Add filler values to tuples
When initializing a tuple or fixed length array with fewer members than the minimum required length, remaining members should be filled with filler values, if the rest type has a filler; otherwise this should be a compile-time error.

i.e. `int[3] a = [1, 2]` should be filled as `[1, 2, 0]`

ushirask:
Fixes #946


manuranga:
This should idealy be done in C.

ushirask:
> This should idealy be done in C.

Since we need a compile-time error. wouldn't it be needed to be done here?

jclark:
It's way, way more complicated than this. I will need to do some work on this first.

Issue #883, opened 2022-02-01
Improve verifiability of BIR's representation of conditional type narrowing


jclark:
The BIR is supposed to be _verifiable_. This means that

- there is a definition of when the BIR is valid (we can distinguish well-formed and well-typed)
- the validity of the BIR is the contract between the front-end and back-end: the front-end is required to produce valid BIR and the backend should handle any valid BIR
- there should be a way to verify that when BIR is valid

Probably the most challenging thing from a verification perspective is conditional type narrowing. Currently, the BIR has a conditional narrowing instruction, which includes the "basis" for the narrowing. The basis is the logical combination of instruction references referring back to the instructions that tested the condition; the results of executing the referenced instructions are what justify the narrowing. The basis part of the conditional narrowing instruction is rather unsatisfying, because it has a much more complex internal structure than other instructions.

Using the basis to verify the correctness of the narrowing has not been implemented yet, and I suspect it will be hard, particularly when the basis is a complex and/or combination.

This issue is to develop an idea to improve on this situation. There are a number of elements:

- different representation of conditional type narrowing in the BIR
- revised strategy for codeGen of logical negation
- explicitly distinguishing forward/backward edges

which I will describe below.

These would need to be tested by an implementation.

jclark:
Alternative idea for representation of conditional type narrowing:

- Represent variable narrowing by having a optional 3-tuple of registers associated with a conditional branch insn
   1. unnarrowed register
   2. narrowed register when true
   3. narrowed register when false
- Add a combine narrowings instruction: this would take 2 (or perhaps N >= 2) existing narrowed registers and construct a new narrowed register with a type that was the union of the existing registers

The narrowed registers would be initialized by these two instructions and would not final (i.e. no instructions would assign to them).

The combine narrowings instruction would be allowed at the start of a block, and would be verified based on the forward edges that come into the block (rather like a phi node).

Edited: alternative to first insn is combined BranchOnTypeTest insn.

jclark:
Generating the alternative representation described above would require a new strategy for handling negation.  The idea is

- codeGenExprForBoolean gets an additional negated argument
- it checks for each kind of expression that can produce a narrowing and calls it with a negated argument; other kinds of expression are handled as now
- codeGenTypeTest already has a negated argument: just need to flip it
- codeGenEqualityExpr needs to be refactored to split the op into negated/exact
- codeGen for `&&` and `||` can handle negated by switching `&&` and `||` and passing negated to subexpressions
[...]

PR #952, opened 2022-03-03
Introduce TypeBranch and TypeMerge instructions

manuranga:
Part of #883

### DIFF
--- a/compiler/modules/bir/bir.bal
+++ b/compiler/modules/bir/bir.bal
@@ -195,9 +195,8 @@ public type AssignTmpRegister readonly & record {|
 
 public type NarrowRegister readonly & record {|
     *RegisterBase;
-    // number of the register that was narrowed
-    // Could be another NarrowRegister.
-    int underlying;
+    // Register that was narrowed. Could be another NarrowRegister.
+    Register underlying;
     // It's name comes from the underlying register.
     () name = ();
     NARROW_REGISTER_KIND kind = NARROW_REGISTER_KIND;
@@ -231,7 +230,7 @@ public function createFinalRegister(FunctionCode code, SemType semType, Position
 }
 
 public function createNarrowRegister(FunctionCode code, SemType semType, Register underlying, Position? pos = ()) returns NarrowRegister {
-    NarrowRegister r = { number: code.registers.length(), underlying: underlying.number, semType, pos };
+    NarrowRegister r = { number: code.registers.length(), underlying, semType, pos };
     code.registers.push(r);
     return r;
 }
@@ -293,11 +292,12 @@ public enum InsnName {
     INSN_CALL,
     INSN_INVOKE,
     INSN_ASSIGN,
-    INSN_COND_NARROW,
     INSN_TYPE_CAST,
     INSN_TYPE_TEST,
     INSN_BRANCH,
     INSN_COND_BRANCH,
+    INSN_TYPE_BRANCH,
+    INSN_TYPE_MERGE,
     INSN_CATCH,
     INSN_PANIC
 }
@@ -327,8 +327,8 @@ public type Insn
     |ListConstructInsn|ListGetInsn|ListSetInsn
     |MappingConstructInsn|MappingGetInsn|MappingSetInsn
     |StringConcatInsn|RetInsn|AbnormalRetInsn|CallInsn
-    |AssignInsn|CondNarrowInsn|TypeCastInsn|TypeTestInsn
-    |BranchInsn|CondBranchInsn|CatchInsn|PanicInsn|ErrorConstructInsn;
+    |AssignInsn|TypeCastInsn|TypeTestInsn|TypeMergeInsn
+    |BranchInsn|TypeBranchInsn|CondBranchInsn|CatchInsn|PanicInsn|ErrorConstructInsn;
 
 public type Operand ConstOperand|Register;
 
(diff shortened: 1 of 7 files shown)


manuranga:
I ran into a bit of an issue.
let's consider the following case:
```ballerina
public function main() {
    int|string x = 1;
    boolean|float y = true;
    int|float z = 0.2;
    if x is int && y is float && z is int {
        int _ = x;
        float _ = y;
        int _ = z;
    }
}
``` 
In the current draft PR I am generating code as follows:
before

I tried to get rid of the extra BBs to get following code:
after

I am able to generate the above but it has backward branches. Looks like I need to create the `type_branches` in the opposite order I create the BBs.

manuranga:
Just for the record above was a misunderstanding, the backwards-ness is determined by the semantics (what needs to be removed to form DAG?) not by the BB label order.

manuranga:
I found another issue. Consider
```ballerina
public function main() {
    int|string|boolean|float|decimal x = 1;
    if x is int|string|boolean && x is int|string && x is int {
        int _ = x;
        return;
    }
    string|boolean|float|decimal _ = x;
}
```
I currently create:



In order to get the correct result, I ought to generate a single merge per `if`, instead of generating pair-wise. Functionality is not affected.

jclark:
What is the graph you should produce?

manuranga:
Same graph, but with a single type merge.
`x.5 = type_merge [bb0 %x.2] [bb1 %x.4] [bb3 %x.7]`

compiler/modules/bir/bir.bal
```diff
@@ -298,6 +297,8 @@ public enum InsnName {
     INSN_TYPE_TEST,
```
jclark, line 296: Need to remove COND_NARROW

compiler/modules/bir/bir.bal
```diff
@@ -628,47 +629,24 @@ public type TypeTestInsn readonly & record {|
...
-    InsnRef insn;
-    // Result of the insn
-    boolean result;
+public type TypeMergeInsn readonly & record {|
+    *InsnBase;
+    INSN_TYPE_MERGE name = INSN_TYPE_MERGE;
+    NarrowRegister result;
+    Register[] operands;
+    // Must be predecessors.length() > 1
+    Label[] predecessors;
```
jclark, line 648: Need to say predecessors.length() == operands.length()

compiler/modules/front/stmt.bal
```diff
@@ -24,15 +24,27 @@ type Assignment readonly & record {|
...
+
+type CondTarget record {|
     bir:BasicBlock block;
-    bir:Register|boolean result;
-    Narrowing[] ifTrue = [];
-    Narrowing[] ifFalse = [];
+    TypeMergeSrc? src = ();
+|};
+
+type TypeMergeSrc record {|
```
jclark: TypeMergeOrigin

compiler/modules/front/stmt.bal
```diff
@@ -272,40 +287,39 @@ function codeGenOnPanic(StmtContext cx, Position pos) {
...
-// If the scope doesn't complete normally, will return empty assignments and non-empty narrowings.
-// Caller should ignore those narrowings and assume it narrows all local variables to NEVER.
-function codeGenScope(StmtContext cx, bir:BasicBlock bb, Environment initialEnv, s:StmtBlock|s:IfElseStmt scope, Narrowing[] initialNarrowing = []) returns CodeGenError|StmtEffect {
+function codeGenScopeForCondTarget(StmtContext cx, CondTarget target, Environment initialEnv, s:StmtBlock|s:IfElseStmt scope) returns CodeGenError|StmtEffect {
+    var { bindings } = initTarget(cx.exprContext(initialEnv), target, scope.startPos);
+    return codeGenScope(cx, target.block, initialEnv, scope, bindings);
+}
+
+// If the scope doesn't complete normally, will return empty assignments and bindings.
+function codeGenScope(StmtContext cx, bir:BasicBlock bb, Environment initialEnv, s:StmtBlock|s:IfElseStmt scope, Binding? initialBinding = ()) returns CodeGenError|StmtEffect {
```
jclark: Say what a scope is

compiler/modules/front/stmt.bal
```diff
@@ -272,40 +287,39 @@ function codeGenOnPanic(StmtContext cx, Position pos) {
...
     Environment env = environmentCopy(initialEnv);
+    if initialBinding != () {
```
jclark: This looks weird

compiler/modules/front/stmt.bal
```diff
@@ -349,26 +363,66 @@ function codeGenStmt(StmtContext cx, bir:BasicBlock? curBlock, Environment env,
 
 function applyEffect(Environment env, StmtEffect effect) {
```
jclark: applyStmtEffect

jclark (changes requested):

compiler/modules/front/stmt.bal
```diff
@@ -24,15 +24,27 @@ type Assignment readonly & record {|
...
 type CondExprEffect record {|
+    // Expression is unconditionally true or false
+    boolean uncond;
+    CondTarget target;
+|}|record {|
+    CondTarget trueTarget;
+    CondTarget falseTarget;
+|};
+
+type CondTarget record {|
```
jclark: TypeMerger

compiler/modules/front/expr.bal
```diff
@@ -302,6 +294,27 @@ function codeGenExpr(ExprContext cx, bir:BasicBlock bb, t:SemType? expected, s:E
...
+    else {
+        bir:AssignTmpRegister result = cx.createAssignTmpRegister(t:BOOLEAN, pos);
+        bir:BasicBlock contBlock = cx.createBasicBlock();
+        assignAndJoin(cx, true, trueMerger.dest, result, contBlock.label, pos);
+        assignAndJoin(cx, false, falseMerger.dest, result, contBlock.label, pos);
+        return { block: contBlock, result };
+    }
+}
+
+function assignAndJoin(ExprContext cx, boolean value, bir:BasicBlock block, bir:AssignTmpRegister result, bir:Label contBlock, Position pos) {
```
jclark: Name of this is not OK. Name needs to make it clear that it is mutating `block`.

`block` should be before `value`.

It's creating an assign and a branch insn. The fact that you are using it for joins is not part of the semantic of the function. So something like addAssignAndBranch.

jclark (changes requested):
I will look again after this is rebased with BindingChain change.

compiler/modules/nback/scaffold.bal
```diff
@@ -190,7 +190,14 @@ class Scaffold {
...
         foreach int i in 0 ..< reprs.length() {
             bir:Register register = code.registers[i];
-            self.addresses.push(builder.alloca(reprs[i].llvm, (), register.name));
+            string? name;
+            if register is bir:NarrowRegister {
+                name = register.underlying.name;
+            }
+            else {
+                name = register.name;
+            }
```
manuranga, line 199: This should be `unnarrow(register).name`

PR #959, opened 2022-03-08
Fix unclear error messages

### LINKED ISSUE #958: Unclear error messages in fixed length arrays
Some errors given for fixed length arrays have unhelpful error messages.
```
public function main() {
    int[4] v = [1, 2, 3, 4];
    io:println(v[4]); // @error
}
```
`error: type of member access is never`

```
public function main() {
    int[4] v = [1, 2, 3, 4];
    v[4] = 4; // @error
```
`error: value assigned to member of list is not a subtype of array member type`

ushirask:
Fixes #958

### DIFF
--- a/compiler/modules/front/expr.bal
+++ b/compiler/modules/front/expr.bal
@@ -442,8 +442,8 @@ function codeGenMemberAccessExpr(ExprContext cx, bir:BasicBlock block1, Position
             var { result: r, block: nextBlock } = check codeGenExprForInt(cx, block1, index);
             t:SemType memberType = t:listMemberType(cx.mod.tc, l.semType, r.semType);
             if t:isEmpty(cx.mod.tc, memberType) {
-                return t:listAtomicTypeRw(cx.mod.tc, l.semType) is () ? 
-                    cx.semanticErr("type of member access is never", pos) : cx.semanticErr("index out of range", s:range(index));
+                string msg = t:listAtomicTypeRw(cx.mod.tc, l.semType) == () ? "type of member access is never" : "index out of range" ;
+                return cx.semanticErr(msg, s:range(index));
             }
             // XXX this isn't correct for singletons
             bir:TmpRegister result = cx.createTmpRegister(memberType, pos);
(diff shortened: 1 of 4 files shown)


compiler/modules/front/expr.bal
```diff
@@ -442,7 +442,8 @@ function codeGenMemberAccessExpr(ExprContext cx, bir:BasicBlock block1, Position
             var { result: r, block: nextBlock } = check codeGenExprForInt(cx, block1, index);
             t:SemType memberType = t:listMemberType(cx.mod.tc, l.semType, r.semType);
             if t:isEmpty(cx.mod.tc, memberType) {
-                return cx.semanticErr("type of member access is never", pos);
+                return t:listAtomicTypeRw(cx.mod.tc, l.semType) is () ?
```
jclark: Don't understand why you are checking whether l.semType is atomic.

If we are going to give two error messages here, I would distinguish the case where r.semType is a singleton.

Also use a conditional expression to compute the message, and then pass that to cx.semanticErr

compiler/modules/front/stmt.bal
```diff
@@ -964,6 +964,9 @@ function codeGenAssignToMember(StmtContext cx, bir:BasicBlock startBlock, Enviro
         else {
             var { result: index, block: nextBlock } = check cx.codeGenExprForInt(block1, env, lValue.index);
             t:SemType memberType = t:listMemberType(cx.mod.tc, reg.semType, index.semType);
+            if t:isEmpty(cx.mod.tc, memberType) && t:listAtomicTypeRw(cx.mod.tc, reg.semType) !is () {
```
jclark: Same comment as for expr.bal

jclark (changes requested):

compiler/modules/front/expr.bal
ushirask: > The original error is thrown for [09-bug/array3-e.bal](https://github.com/ballerina-platform/nballerina/blob/main/compiler/testSuite/09-bug/array3-e.bal), which can be distinguished from fixed length array by `t:listAtomicTypeRw`. Is there a better way to do this?

That case is when the list is equivalent to an empty tuple.  So it is still index out of range. The use of listAtomicTypeRw makes no sense here.

jclark (changes requested):
Using listAtomicTypeRw here is not correct

compiler/modules/front/stmt.bal
```diff
@@ -964,6 +964,9 @@ function codeGenAssignToMember(StmtContext cx, bir:BasicBlock startBlock, Enviro
         else {
             var { result: index, block: nextBlock } = check cx.codeGenExprForInt(block1, env, lValue.index);
             t:SemType memberType = t:listMemberType(cx.mod.tc, reg.semType, index.semType);
+            if t:isEmpty(cx.mod.tc, memberType) {
+                    return cx.semanticErr("index out of range", s:range(lValue.index));
```
jclark: Indentation

jclark (changes requested):

PR #979, opened 2022-03-23
Remove runtime functions from the stacktrace when runtime built with -g

### LINKED ISSUE #970: Remove runtime functions from the stacktrace when runtime built with -g
With #964 if the runtime is compiled with `-g` stack trace in the case of a panic will include the runtime functions (since runtime function calls now have locations). For example in the case of `01-int/div1-p.bal` we will get the stack trace as
```
panic: arithmetic overflow
    _bal_error_construct /home/heshan/work/nballerina/runtime/error.c:10
    _bal_panic_construct /home/heshan/work/nballerina/runtime/panic.c:67
    div ../../../compiler/testSuite/01-int/div1-p.bal
    main (root) ../../../compiler/testSuite/01-int/div1-p.bal:6
    main /home/heshan/work/nballerina/runtime/main.c:11
```
instead of 
```
panic: arithmetic overflow
    div ../../../compiler/testSuite/01-int/div1-p.bal:10
    main (root) ../../../compiler/testSuite/01-int/div1-p.bal:6
```

heshanpadmasiri:
## Purpose
Remove runtime functions from the stacktrace when runtime built with -g
Resolves #970

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/runtime/error.c
+++ b/runtime/error.c
@@ -61,6 +61,7 @@ static void processInitialPC(PC pc, BacktraceStartLine *backtraceStartLine);
 static enum DemangleResult demangle(const char *mangledName, const char **localName, FILE *fp);
 static bool demangleModules(const char **mangledModules, FILE *fp);
 static bool demangleCountedName(const char **pp, const char **name);
+static bool isRuntimeFunction(const char* mangledFunctionName);
 
 TaggedPtr _bal_error_construct(TaggedPtr message, int64_t lineNumber) {
     SimpleBacktrace simpleBacktrace;
@@ -176,7 +177,6 @@ static void processPCs(GC PC* pcs, uint32_t nPCs, uint32_t start, int64_t lineNu
     // Handle the starting pc seperately
     BacktraceStartLine backtraceStartLine = { { error->code, error->message, error->fp }, lineNumber };
     processInitialPC(pcs[start], &backtraceStartLine);
-    backtraceStartLine.lineNumber = 0;
     if (backtraceStartLine.error.code == NO_DEBUG_INFO) {
         for (uint32_t i = start + 1; i < nPCs; i++) {
             // We do not need to pass error_callback because, already we have an error.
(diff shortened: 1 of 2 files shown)


runtime/error.c
```diff
@@ -61,6 +61,7 @@ static void processInitialPC(PC pc, BacktraceStartLine *backtraceStartLine);
 static enum DemangleResult demangle(const char *mangledName, const char **localName, FILE *fp);
 static bool demangleModules(const char **mangledModules, FILE *fp);
 static bool demangleCountedName(const char **pp, const char **name);
+static bool isPrefix(const char* prefix, const char* str);
```
jclark: Space before `*` not after.

runtime/error.c
```diff
@@ -235,6 +240,18 @@ static void printBacktraceLine(const char *filename, int64_t lineNumber, const c
     fflush(fp);
 }
 
+static bool isPrefix(const char* prefix, const char* str) {
```
jclark: You can write this so that it looks at each character of prefix and str at most once. (Don't have calls to strlen.)

jclark (changes requested):

runtime/error.c
heshanpadmasiri: fixed

runtime/error.c
heshanpadmasiri: fixed

runtime/error.c
```diff
@@ -235,6 +240,15 @@ static void printBacktraceLine(const char *filename, int64_t lineNumber, const c
     fflush(fp);
 }
 
+static bool isPrefix(const char *prefix, const char *str) {
+    while (*++prefix == *++str) {
```
jclark: Don't think it's safe to ignore the first character.

```
for (; *prefix == *str && *prefix != '\0'; prefix++, str++) { }
```

jclark (changes requested):

runtime/error.c
heshanpadmasiri: fixed

PR #947, opened 2022-02-28
Add `Region` for `BIR`

poorna2152:

PR #983, opened 2022-03-28
Fix loop backward branches

manuranga:
Follow up on #947.
Backwards flag is missing on `foreach` loops and is in the the wrong place for `continue` on `foreach` loops.
Example:
```ballerina
import ballerina/io;

public function main() {
    foreach int i in 1 ..< 10 {
        io:println(i);
        if i == 6 {
            continue;
        }
        io:println(true);
    }
    io:println(false);
}
```
Current:

In this PR:

### DIFF
--- a/compiler/modules/front/stmt.bal
+++ b/compiler/modules/front/stmt.bal
@@ -50,6 +50,7 @@ type LoopContext record {|
     // following block is reachable
     Assignment[] onBreakAssignments = [];
     Assignment[] onContinueAssignments = [];
+    boolean continueIsBackward;
 |};
 
 class StmtContext {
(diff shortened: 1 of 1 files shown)


compiler/modules/front/stmt.bal
```diff
@@ -50,6 +55,7 @@ type LoopContext record {|
     // following block is reachable
     Assignment[] onBreakAssignments = [];
     Assignment[] onContinueAssignments = [];
+    LoopKind kind;
```
jclark: I think it's easier to have `boolean continueIsBackward;`. Then you don't need the LoopKind enum.

jclark:
I love your diagrams!

manuranga:
Moved to  #990

PR #971, opened 2022-03-17
Verify regions

ushirask:
Partially fixes #995

PR #985, opened 2022-03-31
Verify region in BIR

ushirask:
Verify regions in BIR

### DIFF
--- a/compiler/modules/bir/verify.bal
+++ b/compiler/modules/bir/verify.bal
@@ -70,11 +70,27 @@ public function verifyFunctionCode(Module mod, FunctionDefn defn, FunctionCode c
     foreach BasicBlock b in code.blocks {
         check verifyBasicBlock(cx, b);
     }
+    foreach Region r in code.regions {
+        check verifyRegion(cx, r, code.blocks);
+    }
 }
 
 type IntBinaryInsn IntArithmeticBinaryInsn|IntBitwiseBinaryInsn;
 
 type Error err:Semantic|err:Internal;
+
+function verifyRegion(VerifyContext vc, Region region, BasicBlock[] blocks) returns Error? {
+    BasicBlock entry = blocks[region.entry];
+    Insn insn = entry.insns[entry.insns.length() - 1];
+    if region.kind == REGION_COND && insn !is CondBranchInsn {
+        return vc.invalidErr("condional region is not a conditional insn", pos=insn.pos);
+    }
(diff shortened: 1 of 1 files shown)


compiler/modules/bir/verify.bal
```diff
@@ -70,12 +70,80 @@ public function verifyFunctionCode(Module mod, FunctionDefn defn, FunctionCode c
...
+            insn2 = getCondInsn(blocks, insn.ifFalse);
+            if insn2 is BranchInsn && insn2.dest == region.exit {
+                return;
+            }
+        }
+        return vc.invalidErr("loop region exit is invalid", pos=insn.pos);
+    }
+}
+
+function getCont(Insn insn, BasicBlock[] blocks) returns Insn? {
```
jclark, line 132: Don't know what `Cont` means.

compiler/modules/bir/verify.bal
```diff
@@ -70,12 +70,80 @@ public function verifyFunctionCode(Module mod, FunctionDefn defn, FunctionCode c
...
+    foreach Region r in code.regions {
+        check verifyRegion(cx, r, code.blocks);
+    }
 }
 
 type IntBinaryInsn IntArithmeticBinaryInsn|IntBitwiseBinaryInsn;
 
 type Error err:Semantic|err:Internal;
 
+function verifyRegion(VerifyContext vc, Region region, BasicBlock[] blocks) returns Error? {
```
jclark, line 82: Write a comment describing precisely what this is verifying.

jclark: Why is `blocks` an argument? It's part of `vc`.

compiler/modules/bir/verify.bal
```diff
@@ -70,12 +70,80 @@ public function verifyFunctionCode(Module mod, FunctionDefn defn, FunctionCode c
...
+    if insn is BranchInsn {
+        return getCondInsn(blocks, insn.dest);
+    }
+    else if insn is CondBranchInsn {
+        return insn;
+    }
+    return;
+}
+
+function getCondInsn(BasicBlock[] blocks, Label label) returns Insn {
```
jclark, line 142: I don't know what this does.  I suspect it should be a method on VerifyContext.

compiler/modules/bir/verify.bal
jclark: I suspect it should be a method on VerifyContext.

ushirask:
Moved to #971

PR #987, opened 2022-04-05
Integrate conformance tests

### LINKED ISSUE #980: Integrate conformance tests to the test suite
Integrate [conformance tests](https://github.com/ballerina-platform/ballerina-spec/tree/conformance-tests/conformance/lang) from `ballerina-spec` to the test suite.

heshanpadmasiri:
## Purpose
Integrate conformance tests
Resolves #980 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


jclark:
We don't want to add the conformance .balt files to our repo.  We should use the ones in the spec repo.

heshanpadmasiri:
> We don't want to add the conformance .balt files to our repo. We should use the ones in the spec repo.

fixed

test/balt.mk
```diff
@@ -6,6 +6,6 @@ $(TARGETS): $(TDIRS)
 
 $(TDIRS):
 	mkdir -p out/$@
-	$(MAKE) -C out/$@ -f ../../balt-sub.mk tdir=$@ $(MAKECMDGOALS)
+	$(MAKE) -C out/$@ -f ../../balt-sub.mk balt_files="$(abspath $(wildcard ../compiler/testSuite/*.balt))" tdir=$@ $(MAKECMDGOALS)
```
jclark: Please avoid `abspath`. Everything else in the build system uses relative paths.

conformance/findJava.sh
```diff
@@ -0,0 +1 @@
+../test/findJava.sh
```
jclark: Why do we need this?

conformance/runcheck.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
jclark, line 1: Why do we need this? Isn't it just a copy?

jclark (changes requested):

test/balt.mk
heshanpadmasiri: fixed

conformance/findJava.sh
heshanpadmasiri: fixed. (changed the relative path in `test/balt-sub.mk` to work in both `test` and `conformance` directories)

conformance/runcheck.sh
heshanpadmasiri: fixed. Only difference was how we treat panics. (In normal tests we validate both if a panic happened and if the panic message is correct. But in conformance we check only if panic happened because our panic messages are not the same as jBallerina (Same is true for errors as well). Refactored this part out as `checkPanic.sh`

test/balt.mk
```diff
@@ -6,6 +6,6 @@ $(TARGETS): $(TDIRS)
 
 $(TDIRS):
 	mkdir -p out/$@
-	$(MAKE) -C out/$@ -f ../../balt-sub.mk tdir=$@ $(MAKECMDGOALS)
+	$(MAKE) -C out/$@ -f ../../balt-sub.mk balt_files="$(addprefix "../../" $(wildcard ../compiler/testSuite/*.balt))" tdir=$@ $(MAKECMDGOALS)
```
jclark: Can't we just do pass along TDIRS e.g. `balt_dirs=$(TDIRS)`

test/runcheck.sh
```diff
@@ -16,7 +16,7 @@ mkdir -p "$parent/actual"
 if test $kind == p; then
     balFile=`echo "$b" | sed -e 's/.exe/.bal/'`
     $("./$1" >/dev/null 2>"$out")
-    head -n 2 $out | sed -e 's/.*bal:/'"$balFile"':/' | diff -u "$2" -
+    ../../checkPanic.sh "$out" "$balFile" "$2"
```
jclark: I don't think this is a good approach, because it means checkpanic.sh now has to be run in a particular directory.

Better to have two separate runcheck.sh for now, but put a comment in each saying what the difference is.

heshanpadmasiri: fixed

test/balt.mk
heshanpadmasiri: fixed. `TDIRS` is only the name of directories with balt files.  We need a path to those directories since balt-sub.mk don't know whether we called it from test directory (in this case balt files are in testSuite directory) or conformance directory (in this case tests are in conformance/tests directory).

PR #990, opened 2022-04-06
Build basic blocks in topological order

manuranga:
* Build basic blocks in topological order
* Fix loop backward branches

Prerequisite for #883

### DIFF
--- a/compiler/modules/nback/module.bal
+++ b/compiler/modules/nback/module.bal
@@ -1,4 +1,5 @@
 import wso2/nballerina.bir;
+import wso2/nballerina.comm.err;
 import wso2/nballerina.print.llvm;
 import wso2/nballerina.types as t;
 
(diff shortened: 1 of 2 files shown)


compiler/modules/nback/module.bal
```diff
@@ -107,8 +138,10 @@ function createFunctionDI(ModuleDI mod, bir:File[] files, bir:FunctionDefn birFu
     });
 }
 
-function buildFunctionBody(llvm:Builder builder, Scaffold scaffold, bir:FunctionCode code) returns BuildError? {
-    foreach var b in code.blocks {
-        check buildBasicBlock(builder, scaffold, b);
+function buildFunctionBody(llvm:Builder builder, Scaffold scaffold, bir:BasicBlock[] blocks, bir:Label[] orderedLabels) returns BuildError? {
```
jclark: I suggest you reverse the list before passing it in, so the labels give the order to build in. Then call the param `buildOrder`.

compiler/modules/nback/module.bal
```diff
@@ -56,12 +57,42 @@ public function buildModule(bir:Module birMod, *Options options) returns [llvm:M
...
+            panic err:impossible(orphaned.toString() + " orphan basic block" + (orphaned == 1 ? "" : "s"));
+        }
+        check buildFunctionBody(builder, scaffold, code.blocks, sorted);
     }
     check birMod.finish();
     return [llMod, createTypeUsage(mod.usedSemTypes)];
 }
 
+function topoSort(bir:BasicBlock[] bbs, bir:Label l, bir:Label[] result) {
+    if result.indexOf(l) != () {
```
jclark: Better to have an additional param that is a `boolean[]` saying whether a label is in result.

compiler/modules/nback/module.bal
```diff
@@ -56,12 +57,42 @@ public function buildModule(bir:Module birMod, *Options options) returns [llvm:M
...
+        if orphaned != 0 {
+            panic err:impossible(orphaned.toString() + " orphan basic block" + (orphaned == 1 ? "" : "s"));
+        }
+        check buildFunctionBody(builder, scaffold, code.blocks, sorted);
     }
     check birMod.finish();
     return [llMod, createTypeUsage(mod.usedSemTypes)];
 }
 
+function topoSort(bir:BasicBlock[] bbs, bir:Label l, bir:Label[] result) {
```
jclark: What this is doing IIUC is adding to `result` the labels reachable in forward order from `l`. So I would have a name that reflects that. I would name variables a bit differently

* `blocks` rather than `bbs`
* `label` rather than `l`
* `reachableLabels` rather than `result`

compiler/modules/nback/module.bal
```diff
@@ -56,12 +57,42 @@ public function buildModule(bir:Module birMod, *Options options) returns [llvm:M
...
+        var insn = insns[insnsLen - 1];
+        if insn is bir:BranchInsn && !insn.backward {
+            topoSort(bbs, insn.dest, result);
+        }
+        else if insn is bir:CondBranchInsn {
+            topoSort(bbs, insn.ifTrue, result);
+            topoSort(bbs, insn.ifFalse, result);
+        }
+    }
+    result.push(bb.label);
```
jclark: Isn't this `l`?

Is this code really right? Doesn't the test for `result.indexOf(l)` need to be just before the `push`, since the recursive calls to `topoSort` can add members to the result.

jclark (changes requested):

compiler/modules/nback/module.bal
manuranga: > reachableLabels rather than result


It calculates more than reachability. It is ordered such that each conditional appears before it's jump target. So I named it `ordered`.

compiler/modules/nback/module.bal
manuranga: yes it's `l`, fixed.

Checking in ether place is correct, checking it in the beginning is more efficient. Having a boolean array makes it clearer.

PR #993, opened 2022-04-14
Redo subtyping and projection for fixed length array types and variable length tuples

jclark:
I cannot convince myself of the correctness of the current approach.

The key problem is `listInhabited` needs to explore possibilities that cannot be represented compactly as a FixedLengthArray.  Specifically, with rest types and fixed length arrays we need to explore possibilities that look like `[T1 * N, T2, T3...]`, where N is a large number, and T*N  means T repeated N times.


compiler/testSuite/semtype/proj9-t.bal
```diff
@@ -12,7 +12,8 @@ type T1 int[100000];
 type T2 [C01, C01, C01, (int|float)...];
 
 
-// @type T3[0] = NOTC01
+// T3[0] is int because e.g. [0, 17, 0, 0,...,0] (length 100000) belongs to T1 and !T2
```
jclark, line 15: @rdhananjaya Please check this change.

compiler/testSuite/semtype/proj10-t.bal
```diff
@@ -8,7 +8,7 @@ type FFFloat FirstFive|float;
 type T1 [int...];
 type T2 [int, int, int...];
 
-// @type T3[0] = NEVER
```
jclark, line 11: @rdhananjaya Please check this change.

PR #998, opened 2022-04-28
Refactor bindings

manuranga:
Refactor bindings by extracting linked list to a separate record.

### DIFF
--- a/compiler/modules/front/expr.bal
+++ b/compiler/modules/front/expr.bal
@@ -34,7 +34,6 @@ type Binding record {|
     bir:DeclRegister|bir:NarrowRegister reg;
     boolean isFinal;
     boolean used = false;
-    Binding? prev;
     // When this binding represents a narrowing, this refers to the
     // original binding that was not narrowed.
     // In the case of the nested narrowing, this points all the way
--- a/compiler/modules/front/stmt.bal
+++ b/compiler/modules/front/stmt.bal
@@ -8,8 +8,13 @@ type Position d:Position;
 
 type Range d:Range;
 
+type BindingChain record {|
+    Binding head;
+    BindingChain? prev;
+|};
+
 type Environment record {|
-    Binding? bindings;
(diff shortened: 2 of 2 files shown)


compiler/modules/front/expr.bal
```diff
@@ -34,14 +34,18 @@ type Binding record {|
...
     boolean used = false;
-    Binding? prev;
     // When this binding represents a narrowing, this refers to the
     // original binding that was not narrowed.
     // In the case of the nested narrowing, this points all the way
     // back to the original explicit binding, not to the previous narrowing.
     Binding? unnarrowed = ();
 |};
 
+type Bindings record {|
```
jclark: This should be in stmt.bal.

BindingChain

jclark (changes requested):

compiler/modules/front/stmt.bal
```diff
@@ -375,14 +380,13 @@ function invalidateAssignments(Assignment[] assignments, Narrowing[] narrowings)
 }
 
 function unusedLocalVariables(StmtContext cx, Environment env, Binding? bindingLimit) returns CodeGenError? {
```
jclark, line 382: I wonder about using a Binding to represent bindingLimit. You are relying on each Binding having a distinct (!==) identity. Would it be better to use a BindingChain here?

manuranga: Do you mean to use the distinct identity of `BindingChain` instead of `Binding`? I'll try that.

Another thing that makes Bindings confusing is `()` is used with two different meanings. In `Environment`, bindings being nil indicates the end of the BindingChain. In `StmtEffect`, bindings being nil indicates statement didn't produce any Bindings.

I'll try below

```ballerina
const EMPTY_CHAIN = false;

type BindingChain EMPTY_CHAIN|record {|
    Binding head;
    BindingChain prev;
|};
```

Issue #576, opened 2021-10-18
Implement tuple types


jclark:
Implementing tuples is similar to records except that

- tuples support ordering using relational operators
- we have optimized array implementations

Things to do:

- [x] #891 
- [x] Rework array comparison in backend
- [x] Change ListDesc in runtime and init.bal to support tuples
- [x] Allow tuples in resolve types
- [x] Support in init.bal for type testing tuple types
- [x] Use correct contextually expected type for list constructor members
- [x] #910
- [x] Exactness handling in nback for list get on tuples
- [x] Exactness handling in nback for list set on tuples
- [x] #915
- [x] Make langlib list functions work on tuples (then get rid of arrayMemberType)
- [x] #924
- [x] #946 
- [x] #956

manuranga:
Need to figure out if recursive tuple types can be comparable.

PR #999, opened 2022-04-29
Add failing test cases in conformance tests

heshanpadmasiri:
Error messages I recieved for these tests are as fallows:
listconstruct1-fv.bal:
```
../../../compiler/testSuite/14-bug/listconstruct1-v.bal:3:28: error: ambiguous inherent type for list constructor
    anydata|anydata[] a5 = [1, 2, 3];
```
listconstruct2-fv.bal:
```
error: {ballerina}IteratorMutabilityError {"message":"Table was mutated after the iterator was created"}
        at ballerina.lang.table.0:externNext(internal.bal:42)
           ballerina.lang.table.0.TableIterator:next(internal.bal:39)
           wso2.nballerina.nback.0:createTypeUsage(module.bal:108)
           wso2.nballerina.nback.0:buildModule(module.bal:63)
           wso2.nballerina.0.CompileContext:buildModule(compile.bal:41)
           wso2.nballerina.0:processModule(compile.bal:91)
           wso2.nballerina.0:compileBalFile(compile.bal:76)
           wso2.nballerina.0:main(main.bal:62)

```
listconstruct3-fv.bal
```
../../../compiler/testSuite/14-bug/listconstruct3-v.bal:3:23: error: parse error at '===': invalid expression list
    io:println([1, 2] === [1, 2]); // @output false
```

record-fv.bal
```
../../../compiler/testSuite/14-bug/record-v.bal:14:12: error: ambiguous inherent type for mapping constructor
    C? b = {a: 1};
```

`listfill1-v.bal` actual output:
```
[1,2,3]
[1,"foo"]
```


compiler/testSuite/14-bug/listconstruct3-fv.bal
```diff
@@ -0,0 +1,4 @@
+import ballerina/io;
```
jclark: This is an unimplemented feature (see #1003) not a bug.

compiler/testSuite/14-bug/listfill1-fv.bal
```diff
@@ -0,0 +1,9 @@
+import ballerina/io;
+
+public function main() {
+    int[5] a1 = [1, 2, 3];
```
jclark: This is handled by #576. Please do this as a separate pull request connected to that issue.

compiler/testSuite/14-bug/record-fv.bal
```diff
@@ -0,0 +1,16 @@
...
+|};
+
+type B record {|
+    int a;
+|};
+
+type C A|B;
+
+public function main() {
+    C? b = {a: 1};
```
jclark, line 14: Why do you think this is correct? Please quote relevant part of spec.

compiler/testSuite/14-bug/listconstruct1-fv.bal
```diff
@@ -0,0 +1,6 @@
+import ballerina/io;
+public function main() {
+    anydata|anydata[] a5 = [1, 2, 3];
```
jclark, line 3: Please quote relevant part of spec that shows this is an error.

jclark (changes requested):

compiler/testSuite/14-bug/listconstruct3-fv.bal
heshanpadmasiri: fixed

compiler/testSuite/14-bug/listfill1-fv.bal
heshanpadmasiri: fixed

compiler/testSuite/14-bug/record-fv.bal
heshanpadmasiri: For this and the [fallowing](https://github.com/ballerina-platform/nballerina/pull/999#discussion_r862556737),  issue I had in the spec was with the phrase "it is a compile-time error if this does not leave a single ... type descriptor" (relevant sections [mapping-constructor-expr](https://ballerina.io/spec/lang/2022R1/#mapping-constructor-expr), [list-constructor-expr](https://ballerina.io/spec/lang/2022R1/#list-constructor-expr). If this means 

1. If the number of type descriptors left is not equal to 1 (then it is a compile time error). Then we don't have a bug. JBallerina has one.
2. If the number of type descriptors left is less than 1 ie. 0 (then it is a compile time error). Then we have a bug and JBallerina is correct.

jclark: It certainly means (1) not (2).  You need exactly one type descriptor to use as the inherent type. That's not the issue. The issue is whether anydata|anydata counts as a single type descriptor or two type descriptors.  I think the answer has to be two, because the spec does not define any equality relationship for type descriptors. So I don't think this is a bug in nballerina.

jclark:
listconstruct2-fv looks jballerina compiler bug.  This should be reported.  Try to find a workaround. If not, then the test case should  go in future with a reference to the jballerina issue.

jclark (commented):
Is there anything left apart from listconstruct2?  That should be handled as in https://github.com/ballerina-platform/nballerina/pull/999#issuecomment-1118102271

heshanpadmasiri:
Closed in favor of #1008

PR #1008, opened 2022-05-06
Add workaround for table mutation issue

heshanpadmasiri:
## Purpose
Add workaround for table mutation [issue](https://github.com/ballerina-platform/ballerina-lang/issues/36056)

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/nback/module.bal
+++ b/compiler/modules/nback/module.bal
@@ -102,10 +102,17 @@ function orderByFwdTargets(bir:BasicBlock[] blocks, bir:Label label, boolean[] v
     ordered.push(label);
 }
 
+isolated function index(UsedSemType ty) returns int {
+    return ty.index;
+}
+
 function createTypeUsage(table<UsedSemType> usedSemTypes) returns TypeUsage {
     byte[] uses = [];
     t:SemType[] types = [];
-    foreach var used in usedSemTypes {
+    // JBUG can't directly iterate over usedSemTypes
+    UsedSemType[] usedSemTy = usedSemTypes.toArray();
+    usedSemTy = usedSemTy.sort("ascending", index);
+    foreach var used in usedSemTy {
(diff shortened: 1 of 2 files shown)


compiler/modules/nback/module.bal
```diff
@@ -102,10 +102,17 @@ function orderByFwdTargets(bir:BasicBlock[] blocks, bir:Label label, boolean[] v
...
+    return ty.index;
+}
+
 function createTypeUsage(table<UsedSemType> usedSemTypes) returns TypeUsage {
     byte[] uses = [];
     t:SemType[] types = [];
-    foreach var used in usedSemTypes {
+    // JBUG 36056 can't directly iterate over usedSemTypes
+    UsedSemType[] usedSemTy = usedSemTypes.toArray();
+    usedSemTy = usedSemTy.sort("ascending", index);
```
jclark: I don't think you need this.  They are added to the table in order of increasing index. So the workaround can just change the next line to `usedSemTypes.toArray()`.

jclark (changes requested):

compiler/modules/nback/module.bal
heshanpadmasiri: I also thought the same but just for the failing testcase what I am getting from `usedSemTypes.toArray()` is not sorted.

jclark: That should be reported as another bug (maybe the spec is not as clear as it should be here).

jclark: How about `from var u in usedSemType select u`? That should give them in the right order.

heshanpadmasiri: fixed (used `toArray()` function since directly using the query expression without it causes the same JBUG)

PR #1010, opened 2022-05-09
Verify tmp registers

### LINKED ISSUE #995: Verify register kinds and regions in BIR
**Description:**

Verify register kinds and regions in the BIR verfier

- [x] #963

Check if
- Temp registers are assigned once and only once as a result of an instruction.
- List/map operand in ListSetInsn / MappingSetInsn is not a final register.
- Temp registered are assigned in a previous instruction in the block before being used as a operand in another instruction.
- Narrow register is a subtype of underlying unnarrowed register.

- [ ] Verify regions

Check if
- All blocks in the function code can be reached from block 0 via forward edges
- Backward branches must point to a block that is a loop head
- Region records are created for each single-entry single-exit region in the function code
- Graphs formed by forward edges must be acyclic

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**Steps to reproduce:**

**Related Issues:**

ushirask:
Part of #995 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/bir/verify.bal
+++ b/compiler/modules/bir/verify.bal
@@ -10,8 +10,8 @@ class VerifyContext {
     private final Module mod;
     private final t:Context tc;
     private final FunctionDefn defn;
-    final FunctionCode code;
-    boolean[] tmpRegisterUsed = [];
+    private final FunctionCode code;
+    private final boolean[] tmpRegisterUsed = [];
 
     function init(Module mod, FunctionDefn defn, FunctionCode code) {
         self.mod = mod;
@@ -66,6 +66,22 @@ class VerifyContext {
     function symbolToString(Symbol sym) returns string {
         return self.mod.symbolToString(self.defn.partIndex, sym);
     }
+
+    function getRegister(int i) returns Register {
+        return self.code.registers[i];
+    }
+
+    function getUsedTmpRegisters(int i) returns boolean {
+        return self.tmpRegisterUsed[i];
+    }
+
+    function addTmpRegisters(int i) {
+        self.tmpRegisterUsed[i] = true;
+    }
+
+    function getNumTmpRegisters() returns int {
+        return self.tmpRegisterUsed.length();
+    }
 }
 
 public function verifyFunctionCode(Module mod, FunctionDefn defn, FunctionCode code) returns Error? {
@@ -98,11 +114,11 @@ type ResultInsn IntArithmeticBinaryInsn|IntNoPanicArithmeticBinaryInsn|IntBitwis
 
 function verifyInsnRegisterKinds(VerifyContext vc, Insn insn) returns Error? {
     if insn is ResultInsn {
-        if (vc.tmpRegisterUsed.length() > insn.result.number) && vc.tmpRegisterUsed[insn.result.number] {
+        if (vc.getNumTmpRegisters() > insn.result.number) && vc.getUsedTmpRegisters(insn.result.number) {
             return vc.invalidErr("tmp register defined in multiple places", <Position>insn.result.pos);
         }
         else {
-            vc.tmpRegisterUsed[insn.result.number] = true;
+            vc.addTmpRegisters(insn.result.number);
         }
     }
 
@@ -142,13 +158,13 @@ function verifyRegisterKind(VerifyContext vc, Operand r) returns Error? {
     if r !is Register {
         return;
     }
-    if r is TmpRegister && (vc.tmpRegisterUsed.length() <= r.number || !vc.tmpRegisterUsed[r.number]) {
+    if r is TmpRegister && (vc.getNumTmpRegisters() <= r.number || !vc.getUsedTmpRegisters(r.number)) {
         return vc.invalidErr("tmp register not initialized", <Position>r.pos);
     }
     if r is NarrowRegister {
-        Register narrowed = vc.code.registers[r.underlying];
+        Register narrowed = vc.getRegister(r.underlying);
         while narrowed is NarrowRegister {
-            narrowed = vc.code.registers[narrowed.underlying];
+            narrowed = vc.getRegister(narrowed.underlying);
         }
         if !t:isSubtype(vc.typeContext(), r.semType, narrowed.semType) {
             return vc.invalidErr("narrow register is not a subtype of narrowed register", <Position>r.pos);
--- a/compiler/modules/bir/verify.bal
+++ b/compiler/modules/bir/verify.bal
@@ -100,11 +100,7 @@ type MultipleOperandInsn IntBinaryInsn|IntNoPanicArithmeticBinaryInsn|FloatArith
     |DecimalArithmeticBinaryInsn|CompareInsn|ListConstructInsn|ListGetInsn|MappingConstructInsn
     |MappingGetInsn|StringConcatInsn|EqualityInsn|TypeMergeInsn;
 
-type ResultInsn IntArithmeticBinaryInsn|IntNoPanicArithmeticBinaryInsn|IntBitwiseBinaryInsn
-    |FloatArithmeticBinaryInsn|FloatNegateInsn|DecimalArithmeticBinaryInsn|DecimalNegateInsn
-    |ConvertToIntInsn|ConvertToFloatInsn|ConvertToDecimalInsn|BooleanNotInsn|CompareInsn
-    |EqualityInsn|ListConstructInsn|ListGetInsn|MappingConstructInsn|MappingGetInsn
-    |StringConcatInsn|CallInsn|TypeCastInsn|TypeTestInsn|CatchInsn|ErrorConstructInsn;
+type ResultInsn readonly & record { *ResultInsnBase; };
 
 function verifyInsnRegisterKinds(VerifyContext vc, Insn insn) returns Error? {
     if insn is ResultInsn {
--- a/compiler/modules/bir/verify.bal
+++ b/compiler/modules/bir/verify.bal
@@ -10,15 +10,13 @@ class VerifyContext {
     private final Module mod;
     private final t:Context tc;
     private final FunctionDefn defn;
-    private final FunctionCode code;
     private final boolean[] tmpRegisterUsed = [];
 
-    function init(Module mod, FunctionDefn defn, FunctionCode code) {
+    function init(Module mod, FunctionDefn defn) {
         self.mod = mod;
         t:Context tc  = mod.getTypeContext();
         self.tc = tc;
         self.defn = defn;
-        self.code = code;
     }
 
     function isSubtype(t:SemType s, t:SemType t) returns boolean {
@@ -67,10 +65,6 @@ class VerifyContext {
         return self.mod.symbolToString(self.defn.partIndex, sym);
     }
 
-    function getRegister(int i) returns Register {
-        return self.code.registers[i];
-    }
-
     function getUsedTmpRegisters(int i) returns boolean {
         return self.tmpRegisterUsed[i];
     }
@@ -85,7 +79,7 @@ class VerifyContext {
 }
 
 public function verifyFunctionCode(Module mod, FunctionDefn defn, FunctionCode code) returns Error? {
-    VerifyContext cx = new(mod, defn, code);
+    VerifyContext cx = new(mod, defn);
     foreach BasicBlock b in code.blocks {
         check verifyBasicBlock(cx, b);
     }
@@ -104,7 +98,7 @@ function verifyBasicBlock(VerifyContext vc, BasicBlock bb) returns Error? {
 
 type MultipleOperandInsn IntBinaryInsn|IntNoPanicArithmeticBinaryInsn|FloatArithmeticBinaryInsn
     |DecimalArithmeticBinaryInsn|CompareInsn|ListConstructInsn|ListGetInsn|MappingConstructInsn
-    |MappingGetInsn|StringConcatInsn|EqualityInsn;
+    |MappingGetInsn|StringConcatInsn|EqualityInsn|TypeMergeInsn;
 
 type ResultInsn IntArithmeticBinaryInsn|IntNoPanicArithmeticBinaryInsn|IntBitwiseBinaryInsn
     |FloatArithmeticBinaryInsn|FloatNegateInsn|DecimalArithmeticBinaryInsn|DecimalNegateInsn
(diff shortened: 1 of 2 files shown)


compiler/modules/bir/verify.bal
```diff
@@ -10,12 +10,15 @@ class VerifyContext {
     private final Module mod;
     private final t:Context tc;
     private final FunctionDefn defn;
+    final FunctionCode code;
+    boolean[] tmpRegisterUsed = [];
```
jclark: tmpRegisterUsed looks like it's final too.

I suggest you make both of them private, and then add methods to VerifyContext to encapsulate how you use them.

compiler/modules/bir/verify.bal
```diff
@@ -79,6 +82,77 @@ type Error err:Semantic|err:Internal;
...
         check verifyInsn(vc, insn);
+        check verifyInsnRegisterKinds(vc, insn);
+    }
+}
+
+type MultipleOperandInsn IntBinaryInsn|IntNoPanicArithmeticBinaryInsn|FloatArithmeticBinaryInsn
+    |DecimalArithmeticBinaryInsn|CompareInsn|ListConstructInsn|ListGetInsn|MappingConstructInsn
+    |MappingGetInsn|StringConcatInsn|EqualityInsn;
+
+type ResultInsn IntArithmeticBinaryInsn|IntNoPanicArithmeticBinaryInsn|IntBitwiseBinaryInsn
```
jclark: Might be simpler to do something like `readonly & record { *ResultInsnBase; readonly ... }` (if jBallerina accepts it)

jclark (changes requested):

compiler/modules/bir/verify.bal
ushirask: can we make `tmpRegisterUsed` final since we modify the boolean values?

compiler/modules/bir/verify.bal
```diff
@@ -10,6 +10,7 @@ class VerifyContext {
     private final Module mod;
     private final t:Context tc;
     private final FunctionDefn defn;
+    private final boolean[] tmpRegisterUsed = [];
```
manuranga: I would just pass it as a param and not make VC stateful.

compiler/modules/bir/verify.bal
```diff
@@ -79,6 +92,73 @@ type Error err:Semantic|err:Internal;
...
+        return vc.invalidErr("tmp register not initialized", <Position>r.pos);
+    }
+    if r is NarrowRegister {
+        Register narrowed = r.underlying;
+        while narrowed is NarrowRegister {
+            narrowed = narrowed.underlying;
+        }
+        if !t:isSubtype(vc.typeContext(), r.semType, narrowed.semType) {
+            return vc.invalidErr("narrow register is not a subtype of narrowed register", <Position>r.pos);
+        }
```
manuranga: I don't think we need to check the every time we read form a NarrowRegister, just checking when we first narrow (ie: TypeMerge/TypeBr) is enough. That overlaps with my work.

compiler/modules/bir/verify.bal
```diff
@@ -79,6 +92,73 @@ type Error err:Semantic|err:Internal;
...
+    if insn is ResultInsn {
+        if (vc.getNumTmpRegisters() > insn.result.number) && vc.getUsedTmpRegisters(insn.result.number) {
+            return vc.invalidErr("tmp register defined in multiple places", <Position>insn.result.pos);
+        }
+        else {
+            vc.addTmpRegisters(insn.result.number);
+        }
+    }
+
+    if insn is BranchInsn|CatchInsn {
```
manuranga: Can we use a match stmt instead?
```ballerina
match expr {
      var { operand } => {
          ...
      }
      var { operands } => {
          ...
      }
}

```

compiler/modules/bir/verify.bal
```diff
@@ -79,6 +92,73 @@ type Error err:Semantic|err:Internal;
...
+}
+
+type MultipleOperandInsn IntBinaryInsn|IntNoPanicArithmeticBinaryInsn|FloatArithmeticBinaryInsn
+    |DecimalArithmeticBinaryInsn|CompareInsn|ListConstructInsn|ListGetInsn|MappingConstructInsn
+    |MappingGetInsn|StringConcatInsn|EqualityInsn|TypeMergeInsn;
+
+type ResultInsn readonly & record { *ResultInsnBase; };
+
+function verifyInsnRegisterKinds(VerifyContext vc, Insn insn) returns Error? {
+    if insn is ResultInsn {
```
manuranga: Can we use `if insn is ResultInsnBase {` instead?

compiler/modules/bir/verify.bal
```diff
@@ -79,6 +92,73 @@ type Error err:Semantic|err:Internal;
...
+        if (vc.getNumTmpRegisters() > insn.result.number) && vc.getUsedTmpRegisters(insn.result.number) {
```
manuranga: `if tmpRegisterUsed[insn.result.number] {`

Pre-populate the tmpRegisterUsed array with `false` upto `registers.length()`. Then you don't have to check the length.

compiler/modules/bir/verify.bal
```diff
@@ -67,18 +67,85 @@ class VerifyContext {
...
+                }
+            }
+        }
+    }
+}
+
+function verifyNonFinalRegisterKind(VerifyContext vc, Operand r, boolean[] tmpRegisterUsed) returns Error? {
+    if r is FinalRegister {
+        return vc.invalidErr("invalid register kind final for insn: " + r.kind, <Position>r.pos);
+    }
```
manuranga: Can we do this check at verifyListSet/verifyMappingSet?

compiler/modules/bir/verify.bal
```diff
@@ -67,18 +67,69 @@ class VerifyContext {
...
+
+    if insn is BranchInsn|CatchInsn {
+        return;
+    }
+    else if insn is CallInsn {
+        foreach Operand arg in insn.args {
+            check verifyRegisterKind(vc, arg, tmpRegisterUsed, insn.pos);
+        }
+    }
```
manuranga: Move this to match and get rid of the `if`

compiler/modules/bir/verify.bal
```diff
@@ -67,18 +67,65 @@ class VerifyContext {
...
+        else {
+            tmpRegisterUsed[insn.result.number] = true;
+        }
+    }
+
+    match insn {
+        var { operand } => {
+            check verifyRegisterKind(vc, operand, tmpRegisterUsed, insn.pos);
+        }
+        var { operands } => {
```
manuranga: Can we do `var { operands } | { args: operands } => {` not sure if it works in jBallerina, if not I think a issue needs to be created.

Anyway I think `CallInsn` needs to be changed  to have `operands` instead of `args` for consistency. Please also create an issue for that.

ushirask: We can do `var { operands } | var { args: operands } => {`

compiler/modules/bir/verify.bal
```diff
@@ -67,18 +67,59 @@ class VerifyContext {
...
     foreach Insn insn in bb.insns {
         check verifyInsn(vc, insn);
+        check verifyInsnRegisterKinds(vc, insn, tmpRegisterUsed);
+    }
+}
+
+function verifyInsnRegisterKinds(VerifyContext vc, Insn insn, boolean[] tmpRegisterUsed) returns Error? {
+    if insn is ResultInsnBase {
+        if tmpRegisterUsed[insn.result.number] {
+            return vc.invalidErr("tmp register defined in multiple places", insn.pos);
```
manuranga: I think we should call this `tmpRegisterInit`. `defined` doesn't sound right. Can we say `multiple assignments to a tmp register` ?

ushirask: Fixed.

compiler/modules/bir/verify.bal
```diff
@@ -67,18 +67,59 @@ class VerifyContext {
...
+            // JBUG #35557 can't iterate operands
+            Operand[] ops = operands;
+            foreach Operand op in ops {
+                check verifyRegisterKind(vc, op, tmpRegisterInit, insn.pos);
+            }
+        }
+    }
+}
+
+function verifyRegisterKind(VerifyContext vc, Operand r, boolean[] tmpRegisterInit, Position pos) returns Error? {
```
manuranga: Current name doesn't explain what it does. How about `verifyTmpOprandInitialized`

compiler/modules/bir/verify.bal
```diff
@@ -67,18 +67,59 @@ class VerifyContext {
...
+            // JBUG #35557 can't iterate operands
+            Operand[] ops = operands;
+            foreach Operand op in ops {
+                check verifyTmpOprandInitialized(vc, op, tmpRegisterInit, insn.pos);
+            }
+        }
+    }
+}
+
+function verifyTmpOprandInitialized(VerifyContext vc, Operand r, boolean[] tmpRegisterInit, Position pos) returns Error? {
```
manuranga: `verifyTmpOperandInitialized`

compiler/modules/bir/verify.bal
```diff
@@ -67,18 +67,59 @@ class VerifyContext {
 
 public function verifyFunctionCode(Module mod, FunctionDefn defn, FunctionCode code) returns Error? {
     VerifyContext cx = new(mod, defn);
+    boolean[] tmpRegisterInit = [];
```
manuranga: On a second though, let's call this `tmpRegisterInitialized` to match with the func name.

compiler/modules/bir/verify.bal
```diff
@@ -67,18 +67,59 @@ class VerifyContext {
...
 
-function verifyBasicBlock(VerifyContext vc, BasicBlock bb) returns Error? {
+function verifyBasicBlock(VerifyContext vc, BasicBlock bb, boolean[] tmpRegisterInit) returns Error? {
     foreach Insn insn in bb.insns {
         check verifyInsn(vc, insn);
+        check verifyInsnRegisterKinds(vc, insn, tmpRegisterInit);
+    }
+}
+
+function verifyInsnRegisterKinds(VerifyContext vc, Insn insn, boolean[] tmpRegisterInit) returns Error? {
```
manuranga: `verifyTmpRegisters`

compiler/modules/bir/verify.bal
```diff
@@ -226,6 +274,9 @@ function verifyListSet(VerifyContext vc, ListSetInsn insn) returns Error? {
     if !vc.isSubtype(insn.operands[0].semType, t:LIST) {
         return vc.semanticErr("list set applied to non-list", insn.pos);
     }
+    if insn.operands[0] is FinalRegister {
+        return vc.invalidErr("invalid register kind final for ListSetInsn", insn.pos);
+    }
```
manuranga: This is wrong.

```ballerina
import ballerina/io;
public function main() {
    final int[] i = [1, 2, 3];
    i[0] = 4;
    io:println(i);
}
```
Above is a valid Ballerina program.

compiler/modules/bir/verify.bal
```diff
@@ -251,6 +302,9 @@ function verifyMappingSet(VerifyContext vc, MappingSetInsn insn) returns Error?
     if !vc.isSubtype(insn.operands[0].semType, t:MAPPING) {
         return vc.semanticErr("mapping set applied to non-mapping", insn.pos);
     }
+    if insn.operands[0] is FinalRegister {
+        return vc.invalidErr("invalid register kind final for MappingSetInsn", insn.pos);
+    }
```
manuranga: Same as above

compiler/modules/front/stmt.bal
```diff
@@ -699,6 +703,12 @@ function codeGenMatchStmt(StmtContext cx, bir:BasicBlock startBlock, Environment
     else {
         // Match expression is not a variable: we do not get type narrowing
         int patternIndex = 0;
+        if matched is bir:TmpRegister {
+            bir:AssignTmpRegister result = cx.createAssignTmpRegister(matched.semType, matched.pos);
+            bir:AssignInsn insn = { result, operand: matched, pos : stmt.startPos };
+            matched = result;
```
manuranga: In a match we are only reading `matched`, aren't we ? why do we need `AssignTmpRegister`?

ushirask: In a case like 
```

Comment (ushirask) [reply compiler/modules/front/stmt.bal]: In a case like 
```
match x + 1 {
    2 => { 
        return "1";
    }
    4 => {
        return "2";
    }
}
```
even though matched is a `TmpRegister` it will be used in multiple blocks for Equality instructions, so to avoid that I think we can either do something like this or change all expr code gens to create `AssignTmpRegister` instead of Tmp depending on the situation.

Comment (manuranga) [reply compiler/modules/front/stmt.bal]: I think we'll have to change the definition of `TmpRegister` to, assigned once, but can be used form multiple blocks.
@jclark what do you think?

Comment (ushirask) [reply compiler/modules/front/stmt.bal]: I believe this is already the case, the problem is when the `TmpRegister` is used in a different block than the block it was initially assigned.

Comment (manuranga) [reply compiler/modules/front/stmt.bal]: > but can be used multiple times

I meant
>but can be used from multiple block

let me edit.

**Comment by manuranga** [comment]
@jclark this is ready for you to review.

File: compiler/modules/bir/verify.bal [hunk 19]
```diff
@@ -67,18 +67,34 @@ class VerifyContext {
...
 
-function verifyBasicBlock(VerifyContext vc, BasicBlock bb) returns Error? {
+function verifyBasicBlock(VerifyContext vc, BasicBlock bb, boolean[] tmpRegisterInitialized) returns Error? {
     foreach Insn insn in bb.insns {
         check verifyInsn(vc, insn);
+        check verifyTmpRegisters(vc, insn, tmpRegisterInitialized);
+    }
+}
+
+function verifyTmpRegisters(VerifyContext vc, Insn insn, boolean[] tmpRegisterInitialized) returns Error? {
```
manuranga: let's call this `verifyTmpRegisterAssignment`

compiler/modules/bir/verify.bal
```diff
@@ -67,18 +67,34 @@ class VerifyContext {
 
 public function verifyFunctionCode(Module mod, FunctionDefn defn, FunctionCode code) returns Error? {
     VerifyContext cx = new(mod, defn);
+    boolean[] tmpRegisterInitialized = [];
+    if code.registers.length() > 0 {
```
jclark, line 71: Cleaner to do `tmpRegisterIntialized.setLength(code.registers.length());`

compiler/modules/bir/verify.bal
```diff
@@ -67,18 +67,34 @@ class VerifyContext {
...
+function verifyBasicBlock(VerifyContext vc, BasicBlock bb, boolean[] tmpRegisterInitialized) returns Error? {
     foreach Insn insn in bb.insns {
         check verifyInsn(vc, insn);
+        check verifyTmpRegisterAssignment(vc, insn, tmpRegisterInitialized);
+    }
+}
+
+function verifyTmpRegisterAssignment(VerifyContext vc, Insn insn, boolean[] tmpRegisterInitialized) returns Error? {
+    if insn is ResultInsnBase {
+        if tmpRegisterInitialized[insn.result.number] {
```
jclark, line 92: Add `int num = insn.result.number;`

jclark (changes requested):

anupama-pathirage:
Closing this PR due to inactivity. Please re-open if this is still needed.

PR #1011, opened 2022-05-09
Verify function code graph is acyclic

### LINKED ISSUE #995: Verify register kinds and regions in BIR

ushirask:
Partially fixes #995

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/bir/verify.bal
+++ b/compiler/modules/bir/verify.bal
@@ -104,7 +104,7 @@ function traveseGraph(VerifyContext vc, BasicBlock[] blocks, Label label, boolea
         if insn.backward {
             if visited[insn.dest] {
                 insn = check terminator(vc, blocks[insn.dest], pos);
-                if insn is BranchInsn|CondBranchInsn {
+                if insn is BranchInsn|CondBranchInsn|TypeBranchInsn {
                     return;
                 }
             }
@@ -114,7 +114,7 @@ function traveseGraph(VerifyContext vc, BasicBlock[] blocks, Label label, boolea
             check traveseGraph(vc, blocks, insn.dest, visited, currentGraph, pos); 
         }
     }
-    else if insn is CondBranchInsn {
+    else if insn is CondBranchInsn|TypeBranchInsn {
         boolean[] curGraph = currentGraph.clone();
         check traveseGraph(vc, blocks, insn.ifFalse, visited, currentGraph, pos);
         check traveseGraph(vc, blocks, insn.ifTrue, visited, curGraph, pos);
--- a/compiler/modules/bir/verify.bal
+++ b/compiler/modules/bir/verify.bal
@@ -70,11 +70,55 @@ public function verifyFunctionCode(Module mod, FunctionDefn defn, FunctionCode c
     foreach BasicBlock b in code.blocks {
         check verifyBasicBlock(cx, b);
     }
+    
 }
 
 type IntBinaryInsn IntArithmeticBinaryInsn|IntBitwiseBinaryInsn;
 
 type Error err:Semantic|err:Internal;
+
+function verifyGraph(VerifyContext vc, BasicBlock[] blocks, Position pos) returns Error? {
+    boolean[] visited = from int b in 0 ..< blocks.length() select false;
+    check traveseGraph(vc, blocks, 0, visited, visited, pos);
+    if visited.indexOf(false) != () {
+        return vc.invalidErr("unreachable blocks in function code", pos);
+    }
+}
+
+function traveseGraph(VerifyContext vc, BasicBlock[] blocks, Label label, boolean[] visited, boolean[] currentGraph, Position pos) returns Error? {
+    if currentGraph[label] {
+        return vc.invalidErr("forward edge graph is not acyclic", pos);
+    }
+    if visited[label] {
+        return;
+    }
+    visited[label] = true;
+    currentGraph[label] = true;
+    Insn[] insns = blocks[label].insns;
+    var insnsLen = insns.length();
+    if insnsLen > 0 {
+        Insn insn = insns[insnsLen - 1];
+        if insn is BranchInsn {
+            if insn.backward {
+                if visited[insn.dest] {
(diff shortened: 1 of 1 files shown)


compiler/modules/bir/verify.bal
```diff
@@ -70,12 +70,60 @@ public function verifyFunctionCode(Module mod, FunctionDefn defn, FunctionCode c
...
+    visited[label] = true;
+    currentGraph[label] = true;
+    Insn[] insns = blocks[label].insns;
+    var insnsLen = insns.length();
+    if insnsLen > 0 {
+        Insn insn = insns[insnsLen - 1];
+        if insn is BranchInsn {
+            if insn.backward {
+                if visited[insn.dest] {
+                    insn = blocks[insn.dest].insns[blocks[insn.dest].insns.length() - 1];
```
manuranga: Extract a function called `terminator`, param should be the block.

compiler/modules/bir/verify.bal
```diff
@@ -70,12 +70,60 @@ public function verifyFunctionCode(Module mod, FunctionDefn defn, FunctionCode c
...
+        return;
+    }
+    Label? onPanic = blocks[label].onPanic;
+    if onPanic is Label {
+        visited[onPanic] = true;
+    }
+    visited[label] = true;
+    currentGraph[label] = true;
+    Insn[] insns = blocks[label].insns;
+    var insnsLen = insns.length();
```
manuranga: int

compiler/modules/bir/verify.bal
```diff
@@ -70,12 +70,60 @@ public function verifyFunctionCode(Module mod, FunctionDefn defn, FunctionCode c
...
+    if insnsLen > 0 {
```
manuranga: Is having a block without any insn valid? I think not. We should make this a validation failure.

compiler/modules/bir/verify.bal
manuranga: If we got an illegal BIR, this line may panic. We shouldn't.

compiler/modules/bir/verify.bal
```diff
@@ -70,12 +70,60 @@ public function verifyFunctionCode(Module mod, FunctionDefn defn, FunctionCode c
...
     }
+    check verifyGraph(cx, code.blocks, defn.position);
 }
 
 type IntBinaryInsn IntArithmeticBinaryInsn|IntBitwiseBinaryInsn;
 
 type Error err:Semantic|err:Internal;
 
+function verifyGraph(VerifyContext vc, BasicBlock[] blocks, Position pos) returns Error? {
+    boolean[] visited = from int b in 0 ..< blocks.length() select false;
```
manuranga: No need to use query , `visited[blocks.length() - 1] = false` will fill.

compiler/modules/bir/verify.bal
```diff
@@ -70,12 +70,63 @@ public function verifyFunctionCode(Module mod, FunctionDefn defn, FunctionCode c
...
+    }
+    Label? onPanic = blocks[label].onPanic;
+    if onPanic is Label {
+        visited[onPanic] = true;
+    }
+    visited[label] = true;
+    currentGraph[label] = true;
+    Insn insn = check terminator(vc, blocks[label], pos);
+    if insn is BranchInsn {
+        if insn.backward {
```
manuranga, line 120: We can't just relay on backwards branch, you have to validate it as well.

manuranga: Ok you are just cheeking it to give  error, that is ok. But still this check needs to happen for each path otherwise the it is not strong enough.

compiler/modules/bir/verify.bal
```diff
@@ -151,6 +216,12 @@ function verifyInsn(VerifyContext vc, Insn insn) returns Error? {
     else if insn is ErrorConstructInsn {
         check validOperandString(vc, name, insn.operand, insn.pos);
     }
+    else if insn is BranchInsn && insn.backward {
+        if check terminator(vc, vc.getBlock(insn.dest), insn.pos) is BranchInsn|CondBranchInsn|TypeBranchInsn
```
manuranga, line 220: This is checking that your can't `return` form the LoopHead. I don't think this is a useful thing to check. What we need is a much stronger guarantee. That is, LoopHead is to a dominator. This will automatically guarantee that LoopHead can't `return`.

compiler/modules/bir/verify.bal
```diff
@@ -63,19 +66,81 @@ class VerifyContext {
...
+    check traveseGraph(vc, blocks, 0, visited, pos);
+    if visited.indexOf(false) != () {
+        return vc.invalidErr("unreachable blocks in function code", pos);
+    }
+}
+
+function traveseGraph(VerifyContext vc, BasicBlock[] blocks, Label label, boolean[] visited, Position pos, Label[] currentPath = []) returns Error? {
+    if currentPath.indexOf(label) != () {
+        return vc.invalidErr("forward edge graph is not acyclic", pos);
+    }
```
manuranga, line 107: This check is not strong enough since we don't travel every path due to not revisiting visited blocks.
If we travel all paths, then this is valid, but then it is slower than it needs to be.

manuranga:
I coped most of what this covers to #1026. Rest needs to be addressed under loop region validation. Hence closing.

PR #1012, opened 2022-05-12
Add debug support for variables

### LINKED ISSUE #1006: Implement type system in DWARF
﻿Fallowing #935 in order for us to properly view variables in a debugger we need to
1. Port our type system to DWARF
        DWARF has its own type system described under section 5 [here](https://dwarfstd.org/doc/DWARF4.pdf). We have the necessary functions to build those types using the DIBuilder ([ref](https://llvm.org/doxygen/group__LLVMCCoreDebugInfo.html))
2. (most probably) Add a pretty printer to show values nicely
        We will need something similar to rust-gdb / rust-lldb (relevant part of the code lives [here](https://github.com/rust-lang/rust/tree/master/src/etc ) and is shipped with the rust compiler) so our values are easier to read. Found an easier to follow guide on how to make a simple pretty priter [here](https://undo.io/resources/gdb-watchpoint/here-quick-way-pretty-print-structures-gdb/ ).
3. Integrate VS code 
        We can directly plug GDB/LLDB to the VS code debugger using the launcher.json similar to the following. After that as long as break points and values properly work in the gdb side they should work in VS code side as well. (Confirmed to work with base types)
```
{
        // Use IntelliSense to learn about possible attributes.
        // Hover to view descriptions of existing attributes.
        // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
        // ref: https://code.visualstudio.com/docs/cpp/launch-json-reference
        "version": "0.2.0",
        "configurations": [
            {
                "name": "GDB test",
                "type": "cppdbg", // GDB
                "request": "launch",
                "program": "${workspaceFolder}/1-v.exe", // exe name here
                "cwd": "${workspaceFolder}",
                "stopAtEntry": false,
                "linux": {
                    "MIMode": "gdb",
                    "miDebuggerPath": "/usr/bin/gdb"
                }
            }
        ]
}
```

heshanpadmasiri:
## Purpose
Add debug support for variables
Resolved to #1006

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/front/stmt.bal
```diff
@@ -369,43 +359,53 @@ function addBindings(BindingChain? bindingLimit, BindingChain? bodyBindings, int
     return newBindings;
 }
 
+type ScopedStmt s:MatchStmt|s:WhileStmt|s:ForeachStmt|s:IfElseStmt;
```
manuranga: previous place is better since `codeGenScope` also use this.

runtime/py_gdb/prettyprinter.py
```diff
@@ -0,0 +1,316 @@
...
+    value = (Decimal(10) ** (exponent - 6176)) * significant
+    if sign:
+        value *= -1
+    return value
+
+def dpd_to_int(bits):
+    # original bits are assumed to be Xabc Ydef Zghi
+    # encoding table https://en.wikipedia.org/wiki/Densely_packed_decimal
+    i = bits & 1;
+    f = (bits & 1 << 4) >> 4;
```
jclark: Make precedence explicit between shift and `&`.

compiler/modules/front/stmt.bal
```diff
@@ -370,42 +361,51 @@ function addBindings(BindingChain? bindingLimit, BindingChain? bodyBindings, int
 }
 
 function codeGenStmt(StmtContext cx, bir:BasicBlock? curBlock, Environment env, s:Stmt stmt) returns CodeGenError|StmtEffect {
+    boolean scopedStmt = stmt is ScopedStmt;
```
jclark: I think we can do this more cleanly by adding a function

```
function  codeGenScopedStmt(StmtContext cx, bir:BasicBlock curBlock, Environment env, ScopedStmt stmt) returns CodeGenError|StmtEffect {
    cx.pushScope(start.startPos, stmt.endPos);
    StmtEffect effect;
    if stmt is s:MatchStmt {
        effect = check codeGenMatchStmt(cx, curBlock, env, stmt);
    }
    // etc
    cx.popScope();
    return effect;
}
```

compiler/modules/nback/debug.bal
```diff
@@ -0,0 +1,225 @@
...
+    boolean debugFull;
+|};
+
+// Debug location will always be added
+public const int DEBUG_USAGE_ERROR_CONSTRUCT = 0;
+public const int DEBUG_USAGE_CALL = 1;
+
+public type DebugLocationUsage DEBUG_USAGE_ERROR_CONSTRUCT|DEBUG_USAGE_CALL;
+
+class DIScaffold {
```
jclark, line 37: This is done in rather OO, Java-ish style. Can you make stylistically consistent with Scaffold? The class should just mininimally encapsulate the state.  Most of the real work is done by module level functions not methods.

jclark (changes requested):

compiler/modules/nback/debug.bal
heshanpadmasiri: fixed

compiler/modules/front/stmt.bal

runtime/py_gdb/prettyprinter.py
heshanpadmasiri: fixed

compiler/modules/nback/debug.bal
```diff
@@ -0,0 +1,230 @@
...
+function diScope(bir:Position pos, Scope parent) returns DIScope {
+    foreach var child in parent.childScopes {
+        if child.startPos <= pos && child.endPos >= pos {
+            return diScope(pos, child);
+        }
+    }
+    return parent.diScope;
+}
+
+function semTypeToDWARF(t:SemType semType) returns DWARFType {
```
jclark: Ballerina style is to capitalize just first letter of acronym (e.g. Xml, Http), so Dwarf not DWARF.

jclark (changes requested):

compiler/modules/nback/debug.bal
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    function createScope(bir:Position startPos, bir:Position endPos, Scope parent, Scope[] childScopes) returns Scope {
+        var [line, column] = self.scaffold.lineColumn(startPos);
+        DIScope diScope = self.diBuilder.createLexicalBlock(parent.diScope, self.diFile, line, column);
+        return { diScope, startPos, endPos, childScopes };
+    }
+
+    // we use a seperate declPosition instead of startPos of registerScope since register scope starts at the begining of the Stmt/Function
+    function addRegisterScope(bir:RegisterScope registerScope, bir:Position declPos, Scope parent) returns Scope {
```
jclark: Why does this need to be a method?

compiler/modules/nback/debug.bal
heshanpadmasiri: fixed

compiler/modules/nback/debug.bal
heshanpadmasiri: Made in to a module-level function and renamed `addToParentScope`. This is used to  add `RegisterScope` to `Scope` which acts as nodes in a "scope tree". When creating a debug position (via `debugLocation` function), we need to find a leaf node in that tree covering that position which will then act as the scope for that position.

Issue #1013, opened 2022-05-12
Refactor `front` module following up the TypeBranch changes


manuranga:
Follow up on #952.

Potential refactorings

1. Get rid of Assignment, Environment
Model Assignment in a similar fashion to a narrowing Binding. That is, Assignment will be another Binding where `reg` happens to be the same as unnarrowed Register.

Challenges:

- Currently binding only points all the way back to original binding. But for loop validation we need the immediately previous one.
- Need to add pos to binding

2. Make BindingChain non nillable
```ballerina
const EMPTY_CHAIN = false;

type NonEmptyBindingChain record {|
    Binding head;
    BindingChain prev;
|};

type BindingChain NonEmptyBindingChain|EMPTY_CHAIN;
```
Currently there are two sources of BindingChain being nil, ie 1. No bindings 2. Given statement doesn't produce bindings. This clears it up.

3. Make StmtEffect nillable, and block in it non-nillable

```ballerina
type StmtEffect record {|
   bir:BasicBlock block;
   BindingChain? bindings = ();
|}
```

manuranga:
4. Refactor `groupOriginsByUnnarrowed` such that it only return `MergeOriginGroup`s that narrows **each** path. Then don't need to return `numOrigins`.

PR #1016, opened 2022-05-13
Refactor away Environment and Assignment

### LINKED ISSUE #1014: Assignments in statement blocks ending with `break` does not invalidate narrowings
This is because foreach doesn't return assignments.

https://github.com/ballerina-platform/nballerina/blob/accf778002716d72bc17d3ee961c68b23004e942/compiler/modules/front/stmt.bal#L500-L501

### LINKED ISSUE #1017: Invalid assignment to narrowed variable in loop error
**Description:**
Fallowing code will produce an error `error: assignment to narrowed variable 'a' in loop`
```
import ballerina/io;
public function main() {
    float|int a = 5;
    if a is int {
        a = 8;
        int c = 12;
        while c > 10 {
            c -= 1;
        }
        io:println(c); // @output 10
    }
}
```
Please note that `a` is not assigned inside the loop and we don't get an error if we don't have the while loop

manuranga:
Part of #1013
Fixes #1014
Fixes #1017


compiler/modules/front/expr.bal
```diff
@@ -36,6 +36,9 @@ type Binding record {|
     // In the case of the nested narrowing, this points all the way
     // back to the original explicit binding, not to the previous narrowing.
     Binding? unnarrowed = ();
+    Position pos;
+    // If this binding is an assignment, reg number of the invalidated NarrowRegister
+    int? invalidated = ();
```
jclark: I think we need a longer name for this.

What distinguishes a Binding that represents an assignment? Is it invalidated being non-nil? If so, we should put `assignment` in the name.

Why is this a register number rather than a register?

compiler/modules/front/stmt.bal
```diff
@@ -62,14 +48,13 @@ type LExprEffect record {|
...
     bir:BasicBlock? onBreak;
     bir:BasicBlock? onContinue;
     LoopContext? enclosing;
-    // will use this with while true to determine whether
-    // following block is reachable
-    Assignment[] onBreakAssignments = [];
-    Assignment[] onContinueAssignments = [];
+    // Will use this to validate assignments in loops and
+    // determine assignments to be passed up
+    BindingChain?[] onBreakBindings = [];
```
jclark: I find this a bit weird.  Why do we need a BindingChain here?  It will always be a BindingChain starting with a Binding that represents an assignment, right? It is possible to have a subtype to capture this.

compiler/modules/front/stmt.bal
```diff
@@ -334,134 +305,114 @@ function codeGenScope(StmtContext cx, bir:BasicBlock bb, Environment initialEnv,
...
+}
+
+// List Bindings from `bindings` to `bindingLimit` in reverse
+function reverseBindingChain(BindingChain? bindingLimit, BindingChain? bindings) returns Binding[] {
+    Binding[] result = [];
+    BindingChain? b = bindings;
+    while b !== bindingLimit {
+        // Since `bindingLimit` is a sub-chain of `bindings`, we never reach nil on `b`.
+        var { head, prev } = <BindingChain>b;
+        result.unshift(head);
```
jclark: Not a good idea to do this in a loop.

jclark (changes requested):

compiler/modules/front/expr.bal
manuranga: I changed to `NarrowRegister? assignmentInvalidateReg`. Now with the type being `NarrowRegister` I think even `invalidatedReg` makes sense. wdut?

manuranga:
Still feels like following will be clearer (and I'll be able to get rid of `<bir:NarrowRegister>binding.reg` casting).

```ballerina
type BaseBinding record {|
    string name;
    bir:Register reg;
|};

type DeclBinding record {|
    *BaseBinding;
    bir:DeclRegister reg;
    boolean isFinal;
    boolean used = false;
|};

type NarrowBinding record {|
    *BaseBinding;
    bir:NarrowRegister reg;
    DeclBinding unnarrowed;
|};

type AssignmentBinding record {|
    *BaseBinding;
    bir:DeclRegister reg; // same as unnarrowed.reg and invalidates.underlying.underling...
    DeclBinding unnarrowed;
    bir:NarrowRegister invalidates;
    Position pos;
|};

type Binding DeclBinding|NarrowBinding|AssignmentBinding;
```

Should I try this? Maybe as a separate PR?

jclark:
I think `assignmentInvalidatedReg` is better than `assignmentInvalidateReg`.

I note that expr.bal doesn't use `assignmentInvalidatedReg`. I wonder if this field should be in BindingChain rather than Binding.

compiler/modules/front/expr.bal
```diff
@@ -26,18 +26,35 @@ type ExprEffect record {|
...
 |};
 
+type NarrowBinding record {|
+    *BaseBinding;
+    bir:NarrowRegister reg;
+    DeclBinding unnarrowed;
+|};
+
+type AssignmentBinding record {|
+    *BaseBinding;
```
jclark: This can be *NarrowBinding;

compiler/modules/front/expr.bal
```diff
@@ -26,18 +26,35 @@ type ExprEffect record {|
     Binding? binding = ();
 |};
 
-type Binding record {|
+type BaseBinding record {|
```
jclark: Get rid of this.

compiler/modules/front/expr.bal
```diff
@@ -63,13 +80,13 @@ class ExprContext {
     final StmtContext? sc;
     final ModuleSymbols mod;
     final s:ModuleLevelDefn defn;
-    final Environment env;
+    private final BindingChain? bindings;
```
jclark: BindingChain should be declared in this file.

compiler/modules/front/expr.bal
```diff
@@ -137,9 +154,27 @@ class ExprContext {
         }
     }
 
-    function exprContextWithBindings(BindingChain bindings) returns ExprContext {
-        return new(self.mod, self.defn, self.code, { bindings, assignments: self.env.assignments }, self.sc);
+    function exprContext(BindingChain? bindings) returns ExprContext {
+        return new(self.mod, self.defn, self.code, bindings, self.sc);
+    }
+
+    function createMerger(bir:Label originLabel, TypeMerger? merger) returns TypeMerger {
```
jclark: TypeMerger should be declared in this file

jclark (changes requested):

Issue #1022, opened 2022-05-22
Verify TypeMerge and TypeBranch instructions


manuranga:
Following needs to be verified

Flow related:
* Non-backwards branches form a DAG
~~* Backward edges points to an entry node of the current node's [strongly connected component](https://en.wikipedia.org/wiki/Strongly_connected_component).
(Is this too strong? Will we ever have optimizations that violate this? This is a prerequisite for LOOP_REGION and reducible CFG) 
I plan to do this via [Tarjan's algorithm](https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm).~~ This will be move to Region validation.
* Each Register is assigned before it is used.
* Check type flow :
`TypeBranch`'s `ifTrueRegister` can only be used in successors (inclusive) of `ifTrue`
`TypeBranch`'s `ifFalseRegister` can only be used in successors (inclusive)  of `ifFalse`
* Branches into a TypeMerge matches with its predecessors array

Local:
* TypeBranch: `operand  == rootUnderlying(ifTrueRegister) == rootUnderlying(ifFalseRegister)`
`operand.semType ∩ semType <: ifTrueRegister.semType` 
`operand.semType ∖ semType <: ifFalseRegister.semType` 
* `rootUnderlying(result) == rootUnderlying(operands[n])`
TypeMerge: `∪ operands[n].semType <: result.semType`

Impl choices:
* When to do the local checks (ie call `verifyBasicBlock`)
  - In a separate loop
  - During flow checking

* How to represent Initialized variables. This needs to pass down with the flow, bifurcating at `TypeBranch`.
  - Linked list similar to Bindings.
  - boolean array of size code.registers.length() 
    - Clone the array between sibling branches
    - Mark, unmark and reused for sibling branch

jclark:
I think the easiest is a boolean array and clone between branches.

How are you planning to deal with something like:

```
if foo() {
   int x = 17;
   bar(x);
}
else {
   int y = 20;
   bar(y);
}
```

When you merge these branches, you have x initialized on one branch, and y initialized on the other branch.  They are  both out of scope, but you don't know that from the BIR (unless you use the debug scope info).

I *think* for out current subset we can treat them as both uninitialized.  But this obviously won't be enough for supporting local variable declarations without initializers. For this, I suggest you look at how the Java spec (and maybe C# spec) specify definite assignment.

manuranga:
[...]

PR #1026, opened 2022-05-29
Verify register initialization

manuranga:
Address the flow verification part of the #1022

### DIFF
--- a/compiler/modules/bir/verify.bal
+++ b/compiler/modules/bir/verify.bal
@@ -63,6 +63,38 @@ class VerifyContext {
     function symbolToString(Symbol sym) returns string {
         return self.mod.symbolToString(self.defn.partIndex, sym);
     }
+
+    function verifyCodeContext(FunctionCode code) returns VerifyCodeContext|Error {
+        return new(self, code, self.defn.position);
+    }
+}
+
+class VerifyCodeContext {
+    private final VerifyContext vc;
+    private final Position funcPos;
+    final BasicBlock[] blocks;
+    final int[] fwdInDegrees = [];
+    final RegFlow[][] blocksFlows = []; // RegFlow per block, per each incoming forward edge
+    // predecessors of the currently visiting block. Unlike other fields, which only accumulates data, this grows and shrinks.
+    final BlockSet preds = [];
+
+    function init(VerifyContext vc, FunctionCode code, Position funcPos) returns Error? {
+        // JBUG: error if we moved this assignments below error return
+        self.vc = vc;
+        self.funcPos = funcPos;
+        self.blocks = code.blocks;
+        int numBlocks = self.blocks.length();
+        if numBlocks == 0 {
+            return self.invalidErr("no basic blocks in function code", funcPos);
+        }
+        self.fwdInDegrees.setLength(numBlocks);
+        self.blocksFlows.setLength(numBlocks);
+        self.preds.setLength(numBlocks);
+    }
+
+    function invalidErr(d:Message m, Position|Range? pos = ()) returns err:Internal {
+        return self.vc.invalidErr(m, pos ?: self.funcPos);
+    }
 }
 
 type RegSet boolean[];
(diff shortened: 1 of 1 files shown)


compiler/modules/bir/verify.bal
```diff
@@ -65,13 +65,223 @@ class VerifyContext {
...
+    foreach var insn in block.insns {
+        if insn is TypeMergeInsn {
+            check verifyTypeMergeFlow(vc, afterMerge, flows, insn);
+        }
+        else {
+            afterMerge = true;
+        }
+        match insn {
+            var { result } => {
+                regs[result.number] = true;
```
manuranga, line 221: I need to special case ListSetInsn.
1. Why isn't this failing ?
2. May I go back to how ListSetInsn originally was. ie: having a separate field for result.

compiler/modules/bir/verify.bal
```diff
@@ -65,13 +65,223 @@ class VerifyContext {
...
+
+    foreach BasicBlock b in blocks {
         check verifyBasicBlock(cx, b);
     }
 }
 
+// Verify forward edgers form a DAG
+// Verify backward branch points to a predecessor XXX: need to also verify it's a dominator
+// Calculates forward in-degree for each block
+function verifyStructure(VerifyContext vc, BasicBlock[] blocks, Label current, int[] fwdInDegree, boolean[] preds, Position predPos) returns Error? {
```
jclark: `Graph` or `BlockGraph` better than `Structure`

compiler/modules/bir/verify.bal
```diff
@@ -65,13 +65,223 @@ class VerifyContext {
...
+    Position funcPos = defn.position;
+    BasicBlock[] blocks = code.blocks;
+    int numBlocks = blocks.length();
+    if numBlocks == 0 {
+        return cx.invalidErr("no basic blocks for function", funcPos);
+    }
+    // Checks are done in two DFS rounds. During first round, forward in degree is calculated for each block.
+    int[] fwdInDegree = [];
+    fwdInDegree[numBlocks - 1] = 0;
+    boolean[] preds = [];
```
jclark: Not clear to me what `preds` is.

compiler/modules/bir/verify.bal
```diff
@@ -65,13 +65,223 @@ class VerifyContext {
...
     VerifyContext cx = new(mod, defn);
-    foreach BasicBlock b in code.blocks {
+    Position funcPos = defn.position;
+    BasicBlock[] blocks = code.blocks;
+    int numBlocks = blocks.length();
+    if numBlocks == 0 {
+        return cx.invalidErr("no basic blocks for function", funcPos);
+    }
+    // Checks are done in two DFS rounds. During first round, forward in degree is calculated for each block.
+    int[] fwdInDegree = [];
```
jclark: Is "degree" the standard name for number of incoming edges?

compiler/modules/bir/verify.bal
```diff
@@ -65,13 +65,223 @@ class VerifyContext {
...
+    }
+    else {
+        fwdInDegree[child] = fid + 1;
+    }
+}
+
+function verifyReachable(VerifyContext vc, int[] fwdInDegree, Position pos) returns Error? {
+    foreach int fid in fwdInDegree {
+        if fid == 0 {
+            return vc.invalidErr("block is not reachable", pos);
```
jclark: Don't we want the position of the unreachable block?

jclark (changes requested):

compiler/modules/bir/verify.bal
```diff
@@ -65,13 +65,223 @@ class VerifyContext {
...
+            check verifyOperandInitialized(vc, merge.operands[i], flow.regs, merge.pos);
+        }
+        else {
+            return vc.invalidErr(`superfluous predecessor ${pred} in TypeMergeInsn`, merge.pos);
+        }
+        i += 1;
+    }
+}
+
+function terminator(VerifyContext vc, Insn[] insns, Position predPos) returns Insn|Error {
```
jclark: blockTerminator(Block)

compiler/modules/bir/verify.bal
```diff
@@ -65,13 +65,223 @@ class VerifyContext {
     }
 }
 
+// Marks which registers are initialized from a given path into a block
+type RegFlow record {|
+    int? origin; // origin == () iff entry
```
jclark: Label

compiler/modules/bir/verify.bal
manuranga: ListSetInsn is not a special case. It only has operands no result.

compiler/modules/bir/verify.bal
manuranga: Yes, 'in-degree' is a [standard term](https://en.wikipedia.org/wiki/Glossary_of_graph_theory#D).

compiler/modules/bir/verify.bal
manuranga: Yes, most of the time. I was considering the case where it may be empty. Added a check.

compiler/modules/bir/verify.bal
```diff
@@ -65,11 +65,242 @@ class VerifyContext {
...
+    Label? origin; // origin == () iff entry
+    RegSet regs;
+|};
+
 public function verifyFunctionCode(Module mod, FunctionDefn defn, FunctionCode code) returns Error? {
-    VerifyContext cx = new(mod, defn);
-    foreach BasicBlock b in code.blocks {
-        check verifyBasicBlock(cx, b);
+    VerifyContext vc = new(mod, defn);
+    Position funcPos = defn.position;
```
jclark: Why are we passing this around rather than getting it from `vc`?

compiler/modules/bir/verify.bal
```diff
@@ -65,11 +65,242 @@ class VerifyContext {
...
+    BasicBlock[] blocks = code.blocks;
```
jclark: Feels to me like `blocks`, `fwdInDegrees`, `preds` and `blocksFlows` should be collected into a single object that is passed around to all these functions.

jclark:
Looks basically OK.

compiler/modules/bir/verify.bal
manuranga: For `verifyGraph` and `verifyRegFlow`, we are not passing around `funcPos`, it gets passed in as the boundary condition, but subsequent recursive calls get a different value.

I can get it via `vc` for `verifyReachable`, but though it's not worth it since I'll have to add a method to `vc`.

PR #1037, opened 2022-07-04
Add support for regular expressions

### LINKED ISSUE #1033: Add support for regular expressions
Idea is to define regular expressions as a list type (as `[head, tail]`) and define strings to be checked as lists. Then we can check if that string belongs to the list type recursively.

heshanpadmasiri:
## Purpose
Add support for
+ testing regular expression full match
+ testing regular expression inclusion
Resolves #1033 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


extra/regex/Makefile
```diff
@@ -0,0 +1,3 @@
+build:
```
jclark: The `build` target should be done in a similar way to the `compiler` directory, so it does nothing if nothing needs to be built.

There should be a `push` target in the compiler directory, with a empty file in the build directory whose timestamp says tells whether the push is up to date. The `build` target should depend on this timestamp file.

You need to control the build directory as the compiler Makefile does.

extra/regex/Dependencies.toml
```diff
@@ -0,0 +1,138 @@
...
+dependencies = [
+	{org = "ballerina", name = "jballerina.java"},
+	{org = "ballerina", name = "observe"}
+]
+modules = [
+	{org = "ballerinai", packageName = "observe", moduleName = "observe"}
+]
+
+[[package]]
+org = "heshan"
```
jclark: wso2

extra/regex/Dependencies.toml
```diff
@@ -0,0 +1,138 @@
+# AUTO-GENERATED FILE. DO NOT MODIFY.
```
jclark: We are not checking in Dependencies.toml files.

jclark (commented):
The Makefile needs a lot of work to be consistent with everything else.

extra/regex/Dependencies.toml
heshanpadmasiri: file deleted

extra/regex/Dependencies.toml
heshanpadmasiri: file deleted

extra/regex/Makefile
heshanpadmasiri: Fixed such that all build artifacts (`bala` file and regex `jar` file) will end up in separate subdirectories of the `./build` directory. I am using the `bala` file itself as the timestamp file. I assumed that we will have multiple extras in the future and will prefer to be able to build them separately (with the option to build all of them together)

compiler/Makefile
```diff
@@ -8,12 +9,18 @@ BAL_TEST_FLAGS ?= $(BAL_BUILD_FLAGS) --disable-groups longRunning
...
+
 test:
 	$(BAL) test $(BAL_TEST_FLAGS)
 
 $(COMPILER_JAR): $(COMPILER_SRC)
 	$(BAL) build --target-dir ../build $(BAL_BUILD_FLAGS)
 
+$(COMPILER_PACK): $(COMPILER_SRC)
+	$(BAL) pack --target-dir ../build $(BAL_BUILD_FLAGS)
+	$(BAL) push $(COMPILER_PACK) --repository=local
```
jclark, line 22: Doesn't push accept a `--target-dir` option? If not, please create issue.

compiler/Makefile
```diff
@@ -8,12 +9,18 @@ BAL_TEST_FLAGS ?= $(BAL_BUILD_FLAGS) --disable-groups longRunning
 
 all: $(COMPILER_JAR)
 
+pack: $(COMPILER_PACK)
```
jclark, line 12: Needs to be added to .PHONY

Makefile
```diff
@@ -3,9 +3,10 @@ LLVM_SUFFIX?=-12
 CLANG?=clang$(LLVM_SUFFIX)
 CFLAGS=-O2
 JAVA?=$(shell test/findJava.sh)
-TARGETS=all clean test
+TARGETS=all clean test pack extra
 export CLANG CFLAGS BAL JAVA LLVM_SUFFIX
-SUBDIRS=compiler.d runtime.d test.d
+SUBDIRS=compiler.d runtime.d test.d extra.d
+EXTRAS=regex
```
jclark: Don't see why you need EXTRAS or extras here? Can't you just use extra.d?

jclark (changes requested):
The division of responsibility between Makefile and extra/Makefile isn't right. Knowledge of subdirs of extra should be in either Makefile (in which you don't need extra/Makefile) or extra/Makefil, but not both.

compiler/Makefile
heshanpadmasiri: fixed

Makefile
heshanpadmasiri: What I want to do is run `make regex` (or any other extra in the future)  from the root and be able to build just that (while retaining the ability to build all the extras in one command).

heshanpadmasiri:
> The division of responsibility between Makefile and extra/Makefile isn't right. Knowledge of subdirs of extra should be in either Makefile (in which you don't need extra/Makefile) or extra/Makefil, but not both.

Removed extra/Makefile

compiler/Makefile
heshanpadmasiri: created issue https://github.com/ballerina-platform/ballerina-lang/issues/37163

jclark:
> What I want to do is run `make regex` (or any other extra in the future)  from the root and be able to build just that (while retaining the ability to build all the extras in one command). 

We don't have that capability for other directories.  You can instead cd to the directory and do make there.

Makefile
```diff
@@ -15,6 +15,9 @@ test clean:
 
 test: all
 
+regex:
```
jclark: We don't have this for other subdirectories. Why are we special-casing regex?

heshanpadmasiri: Removed from the top level make file

heshanpadmasiri: Removed from the top level make file

jclark (changes requested):
I think there is some confusion here: `make all` /`make test` in the top-level directory needs to apply to all subdirectories including extra.

heshanpadmasiri:
> I think there is some confusion here: `make all` /`make test` in the top-level directory needs to apply to all subdirectories including extra.

Fixed (moved the default build directory to `build/compiler` and made `bal pack` build to `build/pack` to prevent one command clearing the shared build directory `build` when both are running in parallel)

Issue #1031, opened 2022-06-08
Implement bir textual format


manuranga:
Design a textual format to represent bir. Implement a serializer and a deserializer for it.

manuranga:
Please find the first draft [here](http://htmlpreview.github.io/?https://gist.github.com/manuranga/912810e0cd1baac51c25f5c41c13e10c/raw/93f6b8f89b9f0d36662e266225a3244e94bfe358/nbir-s-expr.html).
(html styling doesn't seems to work in preview, download from [here](https://gist.github.com/manuranga/912810e0cd1baac51c25f5c41c13e10c/raw/93f6b8f89b9f0d36662e266225a3244e94bfe358/nbir-s-expr.html))

jclark:
Let's look at printEq

```
 (function "printEq" (boolean boolean) ()
        (registers
            (% boolean "b1")
            (% boolean "b2")
            (% boolean 2)
            (% (& (! readonly) list) 3)
            (% () 4)
            (% (& (! readonly) list) 5)
            (% () 6)
        )
        (blocks
            (block 0 (anon) (no-panic)
                (== (% "b1") (% "b2") (% 2))
                (cond_branch (% 2) 1 2)
            )
            (block 1 (anon) (no-panic)
                ([].new 1 (% 3))
                (call (id "ballerina" ("io")) "println" (% 3) (% 4))
                (branch 3)
            )
            (block 2 (anon) (no-panic)
                ([].new 0 (% 5))
                (call (id "ballerina" ("io")) "println" (% 5) (% 6))
                (branch 3)
            )
            (block 3 (anon) (no-panic)
                (ret ())
            )
        )
    )
```

jclark:
I would do:

```
(function "printEq" ("b1" "b2") ()
        (registers
            (0 boolean)
            (1 boolean)
            (2 boolean)
            (3 (& (! readonly) list))
            (4 ())
            (5 (& (! readonly) list))
            (6 ())
        )
```

Then 

```
(blocks
            (0 (anon no-panic)
                (== (% 0) (% 1) (% 2))
                (cond-branch (% 2) 1 2)
            )
```

jclark:
I think it's cleaner to use identifiers (e.g. `r0`) rather than numbers are identifiers for blocks/registers, even though internally we use numbers.

jclark:
[...]

PR #1042, opened 2022-07-17
Implement type sexpr based testing

manuranga:
Part of #1031


compiler/modules/types.sexpr/schema.bal
```diff
@@ -0,0 +1,22 @@
+import wso2/nballerina.comm.sexpr;
+
+public type Field [sexpr:Str, Type];
+public type Atom ["list", Type[]] | ["list", Type[], int] | ["array", Type] | ["mapping", Type] | ["mapping", Field[], Type]; 
+// JBUG compile time stack overflow if ["!", Type] is added
+public type Type AtomRef | ["|", Type...] | ["&", Type...] | ["!", Type...] | Int | IntSubtype ;
```
lochana-chathura, line 6: Any reason for not using the already defined types?
```suggestion
public type Atom NonRepeatingTuple | RepeatingTuple | Array | Mapping | OpenMapping; 
// JBUG compile time stack overflow if ["!", Type] is added
public type Type AtomRef | Union | Intersection | Not | Int | IntSubtype ; 
```

manuranga: Ya it gives `invalid cyclic type reference in ...` error. I'll add a comment.
(@SasinduDilshara  please note)

PR #1050, opened 2022-07-25
Fix bug in subtype relation between map and record

### LINKED ISSUE #1049: Bug in subtype relation between `map` and `record`
**Description:**
Fallowing type relation should be correct
```
// @type Cood < IntMap
type Cood record {| int x; int...; |};
type IntMap map<int>;
```
But currently it think `Cood = IntMap`

heshanpadmasiri:
## Purpose
Fix bug in subtype relation between map and record caused due to difference between default values
Resolves #1049

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


heshanpadmasiri:
@jclark while this fixes the problem I think the core issue that is causing this problem is the [fallowing](https://github.com/ballerina-platform/nballerina/issues/1049#issuecomment-1193823509). So I am not sure if it makes more sense to rework NEVER such that rest|NEVER can be represented different from rest

compiler/modules/types/mapping.bal
```diff
@@ -225,7 +225,7 @@ function mappingInhabited(Context cx, TempMappingSubtype pos, Conjunction? negLi
         }
         foreach var { name, index1, type1: posType, type2: negType } in pairing {
             SemType d = diff(posType, negType);
-            if !isEmpty(cx, d) {
+            if !isEmpty(cx, d) || index1 is () {
```
jclark: This doesn't look like a good fix to me. This might insert a named field with a type that is empty, which is generally not allowed.

jclark: This is almost right. See https://github.com/ballerina-platform/nballerina/issues/1049#issuecomment-1197583882

But it's not doing quite the right thing when `index1 == ()`.

heshanpadmasiri: fixed

compiler/modules/types/mapping.bal
```diff
@@ -225,6 +225,10 @@ function mappingInhabited(Context cx, TempMappingSubtype pos, Conjunction? negLi
         }
         foreach var { name, index1, type1: posType, type2: negType } in pairing {
             SemType d = diff(posType, negType);
+            if index1 is () && isEmpty(cx, d) {
```
jclark: Why do you need the `isEmpty`?

heshanpadmasiri: fixed

jclark (changes requested):
Can we have some more test cases?

heshanpadmasiri:
> Can we have some more test cases?

Fixed

PR #1052, opened 2022-07-28
Implement cell type

### LINKED ISSUE #1048: Implement cell type
**Description:**
$subject.

lochana-chathura:
## Purpose
$subject.

Fixes #1048

## Samples
```
(let ((cell-a (cell (int (0 255)) 0))
      (cell-b (cell int 0)))
     (assert< cell-a cell-b))
```


compiler/modules/types/cell.bal
```diff
@@ -0,0 +1,173 @@
+// Implementation specific to basic type cell.
+
+public const READONLY_CELL = 0;
+public const MUTABLE_CELL = 1;
```
jclark: This isn't the right name, because it isn't necessarily mutable.

compiler/modules/types/cell.bal
```diff
@@ -0,0 +1,173 @@
...
+    return createComplexSemType(0, [[UT_CELL, bdd]]);
+}
+
+function cellSubtypeIsEmpty(Context cx, SubtypeData t) returns boolean {
+    Bdd b = <Bdd>t;
+    BddMemo? mm = cx.cellMemo[b];
+    BddMemo m;
+
+    if mm == () {
+        m = {bdd: b};
```
jclark: Style is to put spaces here `{ bdd: b }`

compiler/modules/types/cell.bal
```diff
@@ -0,0 +1,173 @@
...
+        cx.cellMemo.add(m);
+    }
+    else {
+        m = mm;
+        boolean? res = m.isEmpty;
+        if res == () {
+            // we've got a loop
+            // XXX is this right???
+            return true;
```
jclark: Don't see how we can get a loop, since we don't have recursive cells.

compiler/modules/types/cell.bal
```diff
@@ -0,0 +1,173 @@
...
+        }
+    }
+
+    if isEmpty(cx, combined) {
+        return true;
+    }
+
+    match minM {
+        READONLY_CELL => {
+            return checkReadonlyCellSubptyeRelation(cx, combined, negList);
```
jclark: ptye tye type???

compiler/modules/types/cell.bal
```diff
@@ -0,0 +1,173 @@
...
+        n = n.next;
+    }
+
+    SemType? negListUnionResult = negListInfUnion(cx, negList);
+    return negListUnionResult != () && isEmpty(cx, diff(posCombined, negListUnionResult));
+}
+
+function negListInfUnion(Context cx, Conjunction? negList) returns SemType? {
+    SemType? nUnion = ();
+    Conjunction? n = negList;
```
jclark: `n` by itself usually suggests integer type.  Maybe `neg`?

compiler/modules/types/cell.bal
```diff
@@ -0,0 +1,173 @@
...
+            minM = minM > newM ? newM : minM;
+            p = p.next;
+        }
+    }
+
+    if isEmpty(cx, combined) {
+        return true;
+    }
+
+    match minM {
```
jclark: This looks strange. Aren't they all looping over the negList?

compiler/modules/types/cell.bal
```diff
@@ -0,0 +1,173 @@
...
+        while true {
+            if p == () {
+                break;
+            }
+
+            cellAtomType = cx.cellAtomType(p.atom);
+            combined = intersect(combined, cellAtomType.t);
+
+            CellMutability newM = cellAtomType.m;
+            minM = minM > newM ? newM : minM;
```
jclark: Clearer to use `int:min`

ie

```
minM = int:min(minM, cellAtomType.m);
```

compiler/modules/types/cell.bal
```diff
@@ -0,0 +1,173 @@
...
+
+public const READONLY_CELL = 0;
+public const MUTABLE_CELL = 1;
+public const MATCHING_CELL = 2;
+
+public type CellMutability READONLY_CELL|MUTABLE_CELL|MATCHING_CELL;
+
+public type CellAtomicType readonly & record {|
+    SemType t;
+    CellMutability m;
```
jclark: `m` is a bit short here. Maybe `mut`

compiler/modules/types/cell.bal
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    boolean isEmpty = bddEvery(cx, b, (), (), cellFormulaIsEmpty);
+    m.isEmpty = isEmpty;
+    return isEmpty;
+}
+
+function cellFormulaIsEmpty(Context cx, Conjunction? posList, Conjunction? negList) returns boolean {
+    SemType combined;
+    CellMutability minM;
```
jclark: minMut

jclark (changes requested):

compiler/modules/types/cell.bal
```diff
@@ -0,0 +1,151 @@
...
+            minMut = <CellMutability>int:min(minMut, cellAtomType.mut);
+            p = p.next;
+        }
+    }
+
+    CellAtomicType combinedAtomicCell = { t: combined, mut: minMut };
+    return !cellInhabited(cx, combinedAtomicCell, negList);
+}
+
+function cellInhabited(Context cx, CellAtomicType atomicCell, Conjunction? negList) returns boolean {
```
jclark: Better for `atomicCell` to have `pos` in the name. Maybe `posCell`?

compiler/modules/types/cell.bal
```diff
@@ -0,0 +1,151 @@
+// Implementation specific to basic type cell.
+
+public const CELL_MUT_NONE = 0;
+public const CELL_MUT_LIMITED = 1;
+public const CELL_MUT_UNLIMITED = 2;
+
+public type CellMutability CELL_MUT_NONE|CELL_MUT_LIMITED|CELL_MUT_UNLIMITED;
+
+public type CellAtomicType readonly & record {|
+    SemType t;
```
jclark, line 10: I suggest we maintain the t field in a slightly normalized form, specifically if t is equivalent to TOP or NEVER, it should be represented by TOP or NEVER exactly.  Then instead of having to do isEmpty on t, we just do isNever (which is much cheaper). Similarly instead of doing isSameType(t, TOP) we can simply do `t === TOP`.  We do a similar thing in list/mapping, but just for NEVER.

lochana-chathura: if we consider isNever() logic vs isEmpty() logic, they both do
```bal
if t is UniformTypeBitSet {
        return t == 0;
} 
```
at the very beginning. So could there be a difference? 

Also, in `isSameType(t, TOP)` we do `t1 == t2` at the beginning. Doesn't it simply do the job for `TOP` and `NEVER`?

lochana-chathura: @jclark if I got [this](https://github.com/ballerina-platform/nballerina/pull/1052/files#r938425308) right, you are suggesting to do the `isEmpty()`, `isSameType(t, TOP)` check upfront and do the normalization for `t` when creating the `CellAtomicType`. Isn't that more expensive than doing `isEmpty()` once when we have the combinedCell? In the combined cell, types can be canceled out when taking the intersection and would have a more simplified `semType` that we can do `isEmpty()` on?

lochana-chathura: Resolving as per the offline discussion.

PR #1057, opened 2022-08-04
Fix isEmpty check for recursive bdd

### LINKED ISSUE #1041: Types containing only cyclic values are treated as empty
**Description:**

```ballerina
type R1 record {| R2 f; |};

type R2 record {| R1 f; |};

// @type R1 < L2
type L2 string[];

// @type R1 < I
type I int;

type RL1 [int, RL2];
// @type RL1 < I
type RL2 [int, RL1];

```

Subtyping relation between strictly recursive types seems to work as expected.

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**Steps to reproduce:**

**Related Issues:**

lochana-chathura:
## Purpose
$subject.

Fixes #1041

### DIFF
--- a/compiler/modules/types/function.bal
+++ b/compiler/modules/types/function.bal
@@ -41,8 +41,7 @@ function functionSubtypeIsEmpty(Context cx, SubtypeData t) returns boolean {
         if res == () {
             // we've got a loop
             io:println("got a function loop");
-            // XXX is this right???
-            return true;
+            return false;
         }
         else {
             return res;
--- a/compiler/testSuite/semtype/recursive-t.bal
+++ b/compiler/testSuite/semtype/recursive-t.bal
@@ -1,0 +1,14 @@
+type R1 record {| R2 f; |};
+
+type R2 record {| R1 f; |};
+
+// @type R1 <> L
+type L string[];
+
+type RL1 [int, RL2];
+
+type RL2 [int, RL1];
+
+// @type R1 <> I
+// @type RL1 <> I
+type I int;
--- a/compiler/modules/types/list.bal
+++ b/compiler/modules/types/list.bal
@@ -179,8 +179,7 @@ function listSubtypeIsEmpty(Context cx, SubtypeData t) returns boolean {
         boolean? res = m.isEmpty;
         if res == () {
             // we've got a loop
-            // XXX is this right???
-            return true;
+            return false;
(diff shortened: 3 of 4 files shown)


compiler/modules/types/function.bal
```diff
@@ -41,8 +41,7 @@ function functionSubtypeIsEmpty(Context cx, SubtypeData t) returns boolean {
         if res == () {
             // we've got a loop
             io:println("got a function loop");
-            // XXX is this right???
```
jclark, line 44: This is right: I have checked with Giuseppe.

compiler/testSuite/semtype/recursive-t.bal
```diff
@@ -0,0 +1,14 @@
+type R1 record {| R2 f; |};
+
+type R2 record {| R1 f; |};
+
+// @type R1 <> L
+type L string[];
```
jclark, line 6: Do we currently fail this? They are different basic types...

lochana-chathura: Currently `// @type R1 < L` evaluates to `true`.  `R1^~L` becomes `R1` and when we check `isEmpty(R1)` that becomes `true` due to `if res == () { return true}` condition as in R1 we get the same bdd again when checking the emptiness.

jclark: So R1 and R2 are treated as equivalent to `never`?

lochana-chathura: At the moment that's the behaviour.

jclark: OK, so that's the real issue then.

jclark: Can you update issue accordingly?

jclark: Simple test case is then:

```
type L [L];
type N never;
// Should we have N == L
// @type N < L
```

compiler/modules/types/function.bal
jclark: I am starting to wonder whether this fix might be the right one, given we have cells. I will ask.

lochana-chathura:
Closing the PR since this is not the correct fix.

PR #1059, opened 2022-08-12
Serialize bir

manuranga:
Part of #1031

### DIFF
--- a/compiler/modules/nback/module.bal
+++ b/compiler/modules/nback/module.bal
@@ -1,7 +1,10 @@
 import wso2/nballerina.bir;
 import wso2/nballerina.comm.err;
+import wso2/nballerina.comm.sexpr;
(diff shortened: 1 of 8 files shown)


compiler/modules/nback/module.bal
```diff
@@ -50,19 +53,37 @@ public function buildModule(bir:Module birMod, *Options options) returns [llvm:M
         stackGuard: llMod.addGlobal(llvm:pointerType("i8"), mangleRuntimeSymbol("stack_guard")),
         llInitTypes: createInitTypes(llContext)
     };  
+    bsexpr:Function[] funcs = [];
```
jclark, line 56: As we discussed in the meeting, I think we should have an option to output the compiled module as BIR sexpr (instead of LLVM). Then nback should not depend on sexpr.

manuranga:
Moving to #1064, which has both Serialize and Deserialize.

PR #1062, opened 2022-08-18
Fix types containing only cyclic values are treated as empty

### LINKED ISSUE #1041: Types containing only cyclic values are treated as empty

### LINKED ISSUE #1061: Report an error for recursive empty intersections
We do not at the moment completely enforce the rule in the spec that:

> It is a error to have an intersection type that denotes an empty set of shapes

Specifically when there are recursive types involved:

https://github.com/ballerina-platform/nballerina/blob/59d47237788061ef72c2fd4b6749e5bc58c36b65/compiler/modules/front/resolveTypes.bal#L146

heshanpadmasiri:
## Purpose
Fix types containing only cyclic values are treated as empty
Report an error for recursive empty intersections
resolves #1041, #1061

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/front/resolveTypes.bal
```diff
@@ -28,6 +28,15 @@ function resolveTypes(ModuleSymbols mod) returns ResolveTypeError? {
     foreach var defn in mod.defns {
         check resolveDefn(mod, defn);
     }
+    if !mod.tc.env.isReady() {
+        // This should never happen
+        panic error("type environment is not ready");
```
jclark: Use err:impossible

compiler/modules/front/resolveTypes.bal
```diff
@@ -28,6 +28,15 @@ function resolveTypes(ModuleSymbols mod) returns ResolveTypeError? {
...
+    }
+    foreach var [ty, loc] in mod.possiblyEmptyTypes {
+        if t:isEmpty(mod.tc, ty) {
+            return err:semantic("empty or infinite type", loc);
```
jclark: Need a better error message. How about

`invalid recursive type (contains no finite shapes)`

?

compiler/modules/front/resolveTypes.bal
```diff
@@ -329,6 +350,14 @@ function resolveTypeDesc(ModuleSymbols mod, s:ModuleLevelDefn modDefn, int depth
     panic error("unimplemented type-descriptor");
 }
 
+// This defer isEmpty check if Env is not ready. But if Env is ready caller must validate type
```
jclark: I don't see the caller checking.  In any case, this should be responsible for the complete check.

Call it something like `nonEmptyType` and return an error or t:SemType.  Then instead of 

```
            t:SemType ty = d.define(env, fields, rest);
            deferIsEmptyCheck(mod, modDefn, td, ty);
            return ty;
```

do 

```
return check nonEmptyType(mod, modDefn, td, d.define(env, fields, rest));
```

Oh you will need to pass in env too.

jclark (changes requested):

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: fixed

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: fixed

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: fixed. I am using the `env` from the `mod` (`ModuleSymbols`).

compiler/testSuite/14-bug/infiniteArray-v.bal
```diff
@@ -0,0 +1,7 @@
+import ballerina/io;
+type Arr Arr[];
```
jclark: Disallowing types like this is a spec change, so I think we should create a separate directory for this - maybe 14-infinite.  The intersection cases are bugs, so they should go in 14-bug.

compiler/modules/front/symbols.bal
```diff
@@ -11,6 +12,7 @@ type ModuleSymbols record {|
     map<Import>[] partPrefixes = [];
     t:Context tc;
     boolean allowAllTypes = false;
+    [t:SemType, d:Location][] possiblyEmptyTypes = [];
```
jclark: I think the name should better capture the idea that these are types that must be checked for non-emptiness.

compiler/modules/front/resolveTypes.bal
```diff
@@ -329,6 +338,18 @@ function resolveTypeDesc(ModuleSymbols mod, s:ModuleLevelDefn modDefn, int depth
     panic error("unimplemented type-descriptor");
 }
 
+function nonEmptyType(ModuleSymbols mod, s:ModuleLevelDefn modDefn, s:TypeDesc td, t:SemType ty) returns t:SemType|ResolveTypeError {
+    final t:Env env = mod.tc.env;
+    d:Location loc = s:locationInDefn(modDefn, { startPos: td.startPos, endPos: td.endPos });
```
jclark: It bothers me that we are creating a loc here, but most of the time we won't be using it.

compiler/modules/front/resolveTypes.bal
```diff
@@ -329,6 +338,18 @@ function resolveTypeDesc(ModuleSymbols mod, s:ModuleLevelDefn modDefn, int depth
     panic error("unimplemented type-descriptor");
 }
 
+function nonEmptyType(ModuleSymbols mod, s:ModuleLevelDefn modDefn, s:TypeDesc td, t:SemType ty) returns t:SemType|ResolveTypeError {
+    final t:Env env = mod.tc.env;
```
jclark: Not much point to this variable since we are using it only once.

compiler/modules/front/resolveTypes.bal
```diff
@@ -329,6 +338,18 @@ function resolveTypeDesc(ModuleSymbols mod, s:ModuleLevelDefn modDefn, int depth
...
+    d:Location loc = s:locationInDefn(modDefn, { startPos: td.startPos, endPos: td.endPos });
+    if !env.isReady() {
+        mod.possiblyEmptyTypes.push([ty, loc]);
+    }
+    else if t:isEmpty(mod.tc, ty) {
```
jclark: We should factor these three lines out, because they are used in two places. We don't want the same error message in two places.

jclark (commented):
Looks great.  I think we can still improve with a little refactoring

compiler/testSuite/14-bug/infiniteArray-v.bal
heshanpadmasiri: fixed

compiler/modules/front/symbols.bal
heshanpadmasiri: renamed to `deferredEmptinessChecks`. Now it will have information to build the location only when needed

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: Fixed

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: Fixed

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: Refactored to `testTypeForFiniteShape`. (Used that name since even though we are only checking for emptiness error is more specific)

compiler/modules/front/resolveTypes.bal
```diff
@@ -28,6 +28,20 @@ function resolveTypes(ModuleSymbols mod) returns ResolveTypeError? {
...
+    if !mod.tc.env.isReady() {
+        // This should never happen
+        panic err:impossible("type environment is not ready");
+    }
+    foreach var { semType, modDefn, startPos, endPos } in mod.deferredEmptinessChecks {
+        check testTypeForFiniteShape(mod.tc, semType, modDefn, startPos, endPos);
+    }
+}
+
+function testTypeForFiniteShape(t:Context tc, t:SemType semType, s:ModuleLevelDefn modDefn, Position startPos, Position endPos) returns ResolveTypeError? {
```
jclark: We are using this for the intersection case also, which might be empty without having infinite types.

compiler/modules/front/resolveTypes.bal
```diff
@@ -28,6 +28,20 @@ function resolveTypes(ModuleSymbols mod) returns ResolveTypeError? {
...
+    if t:isEmpty(tc, semType) {
+        d:Location loc = s:locationInDefn(modDefn, { startPos, endPos });
+        return err:semantic("invalid recursive type (contains no finite shapes)", loc);
```
jclark, line 59: I think we need to keep track of the original type descriptor and give a different error for the intersection case.

So for the list of deferred types I suggest keeping the type desc instead of the location.

What happens if we have a record type that is empty because it has a field type that is empty? Do we always give the right error in that case.

heshanpadmasiri: Fixed. We had a problem with cases such as this,
```
type T1 record {|
    T2 & T1? next; // @error
|};

type T2 record {|
    int x;
|};
```

where there will be a pending inner type descriptor (intersection) but when we resolve the parent (`T1`) env is ready. Previously we will immediately check `T1` and show the error (for infinite type). Fixed by delaying emptiness checks for any type desc with a pending inner type desc. This should ensure we show the correct error message (i.e error message corresponding to the innermost type desc)

heshanpadmasiri: Currently, I am doing this with a naive approach by

1. In order to find any pending "child" type descriptor I iterate over all the deferred emptiness checks
2. Once the env is ready in `resolveTypes` I sort all the deferred emptiness checks in the descending order of start positions so that inner type desc comes before the outer type desc

The problem with this approach is 1 is slow and 2 doesn't take into account some of the emptiness checks may not be overlapping (as a result in cases like [this](https://github.com/heshanpadmasiri/nballerina/blob/fix-infinite-shapes/compiler/testSuite/14-infinite/infiniteRecord3-e.bal) second type is checked first). I think the correct way to do this is to build a tree/forest structure based on the start and end positions. @jclark do you think it is useful to do that?

jclark:
Sorting by start position seems a bit strange to me. It's to do with dependencies rather than syntactic position.

I want to avoid too much complexity here.  We don't need the perfect message in every case. Just something not too misleading.

jclark:
When you have mutually recursive references, I am not sure what "innermost" means.

heshanpadmasiri:
> When you have mutually recursive references, I am not sure what "innermost" means.

In this case, I think we should show the error at the first (in order they are defined in the code). But in my approach I am showing error at the last.

jclark:
> In this case, I think we should show the error at the first (in order they are defined in the code). But in my approach I am showing error at the last.

I don't think this is worth fixing:

- the real solution is to be able to give multiple errors
- modules can be spread among multiple files, and these is no ordering of these files

heshanpadmasiri:
> Sorting by start position seems a bit strange to me. It's to do with dependencies rather than syntactic position.
> 
> I want to avoid too much complexity here. We don't need the perfect message in every case. Just something not too misleading.

I am using starting position as a heuristic for dependencies. (For cases where we have to deferred isEmpty check if B is dependent on A (i.e type definition of B has A) A tends to have a starting position greater than B)

jclark:
The real problem here is to distinguish cases where the type denoted by the type descriptor contains infinite shapes but no finite shapes.  I don't think we can do this reliably (if we could, we ought to be able to fix the type system to allow them). So the best we can do is try to identify those cases where we are detecting it as empty, but it might contain infinite shapes.  One easy test is to see whether the atom for the record type descriptor is recursive atom or not.

So my suggestion would be:

- build list of types whose emptiness check was deferred
- for now, iterate over the list in order give an error for the first one that is empty
- later, give an error for every type that is empty
- when choosing error message, first look at kind of type descriptor
- if it's an intersection, just say that intersections are not allowed to be empty (give same error as for non-deferred intersection)
- if it's an atomic type (mapping or list), then check whether the atom is a recursive atom
- if it's not recursive, then save its index for now  (I think we should later on get an error from either a recursive atom or an intersection being empty)
- if it is a recursive atom, give an error saying it's an invalid recursive record (contains no finite shapes)
- if we haven't given an error yet, but we have saved the index of a non-recursive atom, then give an error for that

jclark:
> I am using starting position as a heuristic for dependencies. 

I don't think that's a good idea. People have different preferences on how to order things: some prefer top down; some prefer bottom up.

jclark:
Do we give an error message for a field that is explicitly given type `never`?

heshanpadmasiri:
> The real problem here is to distinguish cases where the type denoted by the type descriptor contains infinite shapes but no finite shapes. I don't think we can do this reliably (if we could, we ought to be able to fix the type system to allow them). So the best we can do is try to identify those cases where we are detecting it as empty, but it might contain infinite shapes. One easy test is to see whether the atom for the record type descriptor is recursive atom or not.
> 
> So my suggestion would be:
> 
> * build list of types whose emptiness check was deferred
> * for now, iterate over the list in order give an error for the first one that is empty
> * later, give an error for every type that is empty
> * when choosing error message, first look at kind of type descriptor
> * if it's an intersection, just say that intersections are not allowed to be empty (give same error as for non-deferred intersection)
> * if it's an atomic type (mapping or list), then check whether the atom is a recursive atom
> * if it's not recursive, then save its index for now  (I think we should later on get an error from either a recursive atom or an intersection being empty)
> * if it is a recursive atom, give an error saying it's an invalid recursive record (contains no finite shapes)
> * if we haven't given an error yet, but we have saved the index of a non-recursive atom, then give an error for that

Fixed

heshanpadmasiri:
> Do we give an error message for a field that is explicitly given type `never`?

Yes, but not from this. We given error saying type `never` is not implemented due to [this](https://github.com/ballerina-platform/nballerina/blob/2eae9397791de8ca0fa05ef305eab40c2bbd9719/compiler/modules/front/resolveTypes.bal#L124)

jclark:
We won't get that error if we are using `--showTypes` or doing semtype tests

heshanpadmasiri:
> We won't get that error if we are using `--showTypes` or doing semtype tests

In this case, we will get an error for having saved index of non-recursive atom (last point [here](https://github.com/ballerina-platform/nballerina/pull/1062#issuecomment-1223820886)). We can catch all the infinite cases with recursive atoms only. So may I remove that check?

jclark:
Remove which check?

I think we should

- give an error when resolving a record type if isNever is true for an individual field's type
- make the last point in the list (saved index of non-recursive atom) be a "should not happen" panic (I don't think there's any way this can happen if we make the change to check the field's type), but I'm not completely sure

heshanpadmasiri:
> Remove which check?
> 
> I think we should
> 
> * give an error when resolving a record type if isNever is true for an individual field's type
> * make the last point in the list (saved index of non-recursive atom) be a "should not happen" panic (I don't think there's any way this can happen if we make the change to check the field's type), but I'm not completely sure

Fixed. I misunderstood the earlier comment to mean in showTypes case it is not an error to have never as a field type

compiler/modules/front/resolveTypes.bal
```diff
@@ -28,6 +28,41 @@ function resolveTypes(ModuleSymbols mod) returns ResolveTypeError? {
...
+    }
+}
+
+function testTypeForFiniteShape(ModuleSymbols mod, t:SemType semType, s:ModuleLevelDefn modDefn, s:TypeDesc td) returns ResolveTypeError? {
+    if t:isEmpty(mod.tc, semType) {
+        d:Location loc = s:locationInDefn(modDefn, { startPos: td.startPos, endPos: td.endPos });
+        if td is s:BinaryTypeDesc && td.op is "&" {
+            return err:semantic("intersection must not be empty", loc);
+        }
+        // We are only deffering intersections(already handled), lists and mappings
```
jclark: Spelling: deferring

compiler/modules/front/resolveTypes.bal
```diff
@@ -28,6 +28,41 @@ function resolveTypes(ModuleSymbols mod) returns ResolveTypeError? {
...
+    foreach var { semType, modDefn, td } in mod.deferredEmptinessChecks {
+        // XXX When we can give multiple errors we should check all the deferred emptiness checks
+        check testTypeForFiniteShape(mod, semType, modDefn, td);
+    }
+    if mod.emptySourceTypeIndices.length() != 0 {
+        panic err:impossible("there are non recursive empty types");
+    }
+}
+
+function testTypeForFiniteShape(ModuleSymbols mod, t:SemType semType, s:ModuleLevelDefn modDefn, s:TypeDesc td) returns ResolveTypeError? {
```
jclark: The name isn't right, because we are not just testing finite shapes.  Maybe `nonEmptyTypeDeferred`.

compiler/modules/front/resolveTypes.bal
```diff
@@ -28,6 +28,41 @@ function resolveTypes(ModuleSymbols mod) returns ResolveTypeError? {
...
+    if !mod.tc.env.isReady() {
+        // This should never happen
+        panic err:impossible("type environment is not ready");
+    }
+    foreach var { semType, modDefn, td } in mod.deferredEmptinessChecks {
+        // XXX When we can give multiple errors we should check all the deferred emptiness checks
+        check testTypeForFiniteShape(mod, semType, modDefn, td);
+    }
+    if mod.emptySourceTypeIndices.length() != 0 {
+        panic err:impossible("there are non recursive empty types");
```
jclark: Need a location here.

compiler/modules/front/resolveTypes.bal
```diff
@@ -28,6 +28,41 @@ function resolveTypes(ModuleSymbols mod) returns ResolveTypeError? {
     foreach var defn in mod.defns {
         check resolveDefn(mod, defn);
     }
+    if !mod.tc.env.isReady() {
+        // This should never happen
+        panic err:impossible("type environment is not ready");
+    }
+    foreach var { semType, modDefn, td } in mod.deferredEmptinessChecks {
```
jclark, line 38: Make a separate function starting from here.

compiler/modules/front/resolveTypes.bal
```diff
@@ -28,6 +28,41 @@ function resolveTypes(ModuleSymbols mod) returns ResolveTypeError? {
...
+        // XXX When we can give multiple errors we should check all the deferred emptiness checks
+        check testTypeForFiniteShape(mod, semType, modDefn, td);
+    }
+    if mod.emptySourceTypeIndices.length() != 0 {
```
jclark: This doesn't need to be in `mod`. We need only a single location not the list of indices.

compiler/modules/front/resolveTypes.bal
```diff
@@ -28,6 +28,41 @@ function resolveTypes(ModuleSymbols mod) returns ResolveTypeError? {
...
+        }
+        // We are only deffering intersections(already handled), lists and mappings
+        t:ComplexSemType t = <t:ComplexSemType>semType;
+        t:BddNode subtypeDataList = <t:BddNode>t.subtypeDataList[0];
+        t:Atom atom = subtypeDataList.atom;
+        if atom is t:RecAtom {
+            int baseTypeIndex = atom;
+            int? index = mod.emptySourceTypeIndices.indexOf(baseTypeIndex);
+            if index is int {
+                _ = mod.emptySourceTypeIndices.remove(index);
```
jclark: This approach will be really slow.

compiler/modules/types/core.bal
```diff
@@ -22,11 +22,11 @@ public type UniformTypeCode
     |UT_XML_RW|UT_LIST_RW|UT_MAPPING_RW|UT_TABLE_RW|UT_OBJECT_RW
     |UT_STREAM|UT_FUTURE;
 
-type Atom RecAtom|TypeAtom;
+public type Atom RecAtom|TypeAtom;
```
jclark: Better for types to provide some abstractions to support this: e.g. isAtomRecursive function.

compiler/modules/front/symbols.bal
```diff
@@ -11,6 +11,14 @@ type ModuleSymbols record {|
...
     t:Context tc;
     boolean allowAllTypes = false;
+    DeferredEmptinessCheck[] deferredEmptinessChecks = [];
+    int[] emptySourceTypeIndices = [];
+|};
+
+type DeferredEmptinessCheck record {|
+    t:SemType semType;
+    s:ModuleLevelDefn modDefn;
+    s:TypeDesc td;
```
jclark, line 23: Why do you need both modDefn and td?

jclark (changes requested):

compiler/modules/front/symbols.bal
heshanpadmasiri: `modDefn` can be a `FunctionDefn`

heshanpadmasiri: fixed

compiler/modules/types/core.bal

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: fixed

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: fixed

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: fixed

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: fixed

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: fixed

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: fixed (changed error to syntax error since `err:impossible` don't take a location

compiler/modules/front/symbols.bal
```diff
@@ -11,6 +11,12 @@ type ModuleSymbols record {|
     map<Import>[] partPrefixes = [];
     t:Context tc;
     boolean allowAllTypes = false;
+    map<DeferredEmptinessCheck[]> deferredEmptinessChecks = {};
```
jclark: Why does this need to be a map rather than an array?

compiler/modules/types/core.bal
```diff
@@ -22,7 +22,7 @@ public type UniformTypeCode
...
```
jclark: I don't want this change to expose internal implementation details of this module that were not previously exposed.

compiler/modules/front/resolveTypes.bal
```diff
@@ -28,6 +28,52 @@ function resolveTypes(ModuleSymbols mod) returns ResolveTypeError? {
...
+function nonEmptyTypeDeferred(ModuleSymbols mod, t:SemType semType, s:ModuleLevelDefn modDefn) returns ResolveTypeError? {
+    if t:isEmpty(mod.tc, semType) {
+        s:TypeDesc td = modDefn is s:FunctionDefn ? modDefn.typeDesc : <s:TypeDesc>modDefn.td;
+        d:Location loc = s:locationInDefn(modDefn, { startPos: td.startPos, endPos: td.endPos });
+        if td is s:BinaryTypeDesc && td.op is "&" {
+            _ = mod.deferredEmptinessChecks.removeIfHasKey(modDefn.name); // Currently this has no effect
+            return err:semantic("intersection must not be empty", loc);
+        }
+        // We are only defering intersections(already handled), lists and mappings
+        t:ComplexSemType t = <t:ComplexSemType>semType;
```
jclark: This module shouldn't know about the internals of how the types module deals with atomic types.

I think it would be better to the types module to have a function that takes a SemType and tests whether it is atomic and recursive.

jclark (changes requested):

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: fixed

compiler/modules/types/core.bal

compiler/modules/front/symbols.bal
heshanpadmasiri: Seperated out what I need to check if an empty atomic type is a recursive type into `recursiveEmptinessChecked`.

heshanpadmasiri: I need this to separate cases where we have an empty atomic type that is recursive (infinite) and empty atomic type that is not recursive (something I don't think should be possible).

compiler/modules/types/core.bal
```diff
@@ -1589,3 +1589,20 @@ function init() {
         {} // RW object
    ];
 }
+
+public function isSemTypeRecursive(SemType semType) returns [boolean, int?] {
```
jclark: You can't call it `isSemTypeRecursive` if it doesn't return a `boolean`.

I think you can just return an `int?` here. Then call it something like `recursiveAtomIndex`.

But it would be better if we could avoid exposing this index.

compiler/modules/front/symbols.bal
```diff
@@ -11,6 +12,14 @@ type ModuleSymbols record {|
     map<Import>[] partPrefixes = [];
     t:Context tc;
     boolean allowAllTypes = false;
+    ResolveType[] deferredEmptinessChecks = [];
+    map<[boolean, d:Location]> recursiveEmptinessChecked = {};
+|};
+
+type ResolveType record {|
```
jclark: `ResolvedType` (with a `d` is better)

compiler/modules/front/resolveTypes.bal
```diff
@@ -28,6 +28,41 @@ function resolveTypes(ModuleSymbols mod) returns ResolveTypeError? {
...
+function finishEmptinessCheck(ModuleSymbols mod, t:SemType semType, s:ModuleLevelDefn modDefn, s:TypeDesc td) returns ResolveTypeError? {
+    if t:isEmpty(mod.tc, semType) {
+        d:Location loc = s:locationInDefn(modDefn, { startPos: td.startPos, endPos: td.endPos });
+        if td is s:BinaryTypeDesc && td.op is "&" {
+            return err:semantic("intersection must not be empty", loc);
+        }
+        var [isRecursive, index] = t:isSemTypeRecursive(semType);
+        if isRecursive {
+            if index is int {
+                mod.recursiveEmptinessChecked[index.toString()] = [true, loc];
```
jclark: You should use a table rather than a map with int:toString.

compiler/modules/front/resolveTypes.bal
```diff
@@ -28,6 +28,41 @@ function resolveTypes(ModuleSymbols mod) returns ResolveTypeError? {
...
+    }
+    foreach var [checked, loc] in mod.recursiveEmptinessChecked {
+        if !checked {
+            return err:semantic("non recursive empty type", loc);
+        }
+    }
+}
+
+function finishEmptinessCheck(ModuleSymbols mod, t:SemType semType, s:ModuleLevelDefn modDefn, s:TypeDesc td) returns ResolveTypeError? {
+    if t:isEmpty(mod.tc, semType) {
```
jclark: Start off with `if !t:IsEmpty(...) { return; }`, then you don't have to indent the whole thing.

compiler/modules/front/symbols.bal
```diff
@@ -11,6 +12,14 @@ type ModuleSymbols record {|
     map<Import>[] partPrefixes = [];
     t:Context tc;
     boolean allowAllTypes = false;
+    ResolveType[] deferredEmptinessChecks = [];
+    map<[boolean, d:Location]> recursiveEmptinessChecked = {};
```
jclark: Use a record with meaningful names, not a tuple.

compiler/modules/front/resolveTypes.bal
```diff
@@ -28,6 +28,41 @@ function resolveTypes(ModuleSymbols mod) returns ResolveTypeError? {
...
+function finishDeferredEmptinessChecks(ModuleSymbols mod) returns ResolveTypeError? {
+    if !mod.tc.env.isReady() {
+        panic err:impossible("type environment is not ready");
+    }
+    foreach var { semType, modDefn, td } in mod.deferredEmptinessChecks {
+        // XXX When we can give multiple errors we should check all the deferred emptiness checks
+        check finishEmptinessCheck(mod, semType, modDefn, td);
+    }
+    foreach var [checked, loc] in mod.recursiveEmptinessChecked {
+        if !checked {
```
jclark: Isn't the first half of the tuple saying whether it's recursive? Everything in the map is checked.

jclark (changes requested):
Use recursiveAtomIndex for now. With the changes above, hopefully I will be able to understand what you are doing...

compiler/modules/types/core.bal
heshanpadmasiri: I was trying to get two different things here
1. Given a semtype (it is always a list or mapping) figure out if it a recursive atom
2. Get a common identifier for both recAtom and typeAtom of a given type desc. I was using the index for that

I changed 2 to the pair [startPos, endPos] of the type desc. So reverted this function to just checking 1 (as the name suggests)

compiler/modules/front/symbols.bal
heshanpadmasiri: fixed

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: Fixed

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: fixed

compiler/modules/front/symbols.bal
heshanpadmasiri: Fixed

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: Yes. Changed to `recursive` in the `AtomicEmptyType`

compiler/modules/front/resolveTypes.bal
```diff
@@ -329,6 +369,16 @@ function resolveTypeDesc(ModuleSymbols mod, s:ModuleLevelDefn modDefn, int depth
     panic error("unimplemented type-descriptor");
 }
 
+function nonEmptyType(ModuleSymbols mod, s:ModuleLevelDefn modDefn, s:TypeDesc td, t:SemType semType) returns t:SemType|ResolveTypeError {
+    if !mod.tc.env.isReady() {
+        mod.deferredEmptinessChecks.push({ semType, modDefn, td });
+    }
+    else {
+        check finishEmptinessCheck(mod, semType, modDefn, td);
```
jclark: finishEmptinessCheck is doing the same thing as nonEmptyType except that it assumes the env is ready and so won't defer.  The naming of the two functions should reflect this and the parameter order should be the same.

e.g. rename finishEmptinessCheck should be something like nonEmptyTypeNoDefer

compiler/modules/front/symbols.bal
```diff
@@ -11,6 +12,21 @@ type ModuleSymbols record {|
...
+|};
+
+type ResolvedType record {|
+    t:SemType semType;
+    s:ModuleLevelDefn modDefn;
+    s:TypeDesc td;
+|};
+
+type AtomicEmptyType record {|
+    readonly d:Position startPos;
```
jclark: Using startPos and endPos doesn't uniquely identify the type desc (because there can be multiple source parts with the same position).

You can use partIndex in the identifier and SourceFile to construct the location.

But I'm not sure you need this table.

compiler/modules/front/resolveTypes.bal
```diff
@@ -28,6 +28,42 @@ function resolveTypes(ModuleSymbols mod) returns ResolveTypeError? {
...
+    if !t:isEmpty(mod.tc, semType) {
+        return;
+    }
+    d:Position startPos = td.startPos;
+    d:Position endPos = td.endPos;
+    d:Location loc = s:locationInDefn(modDefn, { startPos, endPos });
+    if td is s:BinaryTypeDesc && td.op is "&" {
+        return err:semantic("intersection must not be empty", loc);
+    }
+    if t:isSemTypeRecursive(semType) {
```
jclark, line 58: I don't understand why you need the `atomicEmptyTypeRecursionCheck` table.

What I suggested was not to give an error right away for a non-recursive atomic empty type, because there will should be an empty intersection or recursive atomic type that is the real cause of the problem.  So we just need to save one of the non-recursive empty types that we can use to give a panic in case (unexpectedly) we don't find a recursive error.  This also requires catching a member type for which isNever is true.

jclark (changes requested):

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: fixed. (I am using a sematic error instead of impossible panic so we can have a location in the error)

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: fixed

compiler/modules/front/symbols.bal
heshanpadmasiri: removed table

compiler/modules/front/resolveTypes.bal
heshanpadmasiri: fixed

PR #1067, opened 2022-09-04
Fix SemType recursive types failure

### LINKED ISSUE #1065: SemType recursive types failure
If the following is added as a test case to e.g. `compiler/testSuite/semtype/recurse-t.bal`

```
type T_str ["a", T4] & readonly;
type T4 ["d", ()] & readonly;

// @type T_str < T_regex
type T_regex ["a", T2] & readonly;
type T2 (["b", T3] & readonly)| (["d", ()] & readonly);
type T3 ["c", T2] & readonly;
```

then it fails with the error:

```
                [fail] testSemTypes#'semtype/recurse-t.bal':

                    T_str is not a proper subtype of T_regex

                        expected: <anydata[]> '[true,false]'
                        actual  : <boolean[2]> '[true,true]'

                        at ballerina.test.0:createBallerinaError(assert.bal:43)
                        ballerina.test.0:assertEquals(assert.bal:112)
                        wso2.nballerina$test.0.tests.testSuite:testSubtypes(tests/testSuite.bal:210)
                        wso2.nballerina$test.0.tests.testSuite:testSemTypes(tests/testSuite.bal:38)
```

But when run with `--showTypes` it shows the correct relationships

```
T4<:T2
T_str<:T_regex
```

This test comes from translating the regex test

```
"ad" < "a(bc)*d"
```

and adding `readonly`.

The same thing happens when `readonly` is replaced by `!readonly`.  The error also happens with the tuple is replaced by a closed record with two fields.

Deepening the mystery further, if `readonly` is replaced by `any|error`, the test passes.

heshanpadmasiri:
## Purpose
Fix SemType recursive types failure
Resolves #1065 
Revert #1066 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/types/list.bal
```diff
@@ -118,25 +118,32 @@ function listSubtypeIsEmpty(Context cx, SubtypeData t) returns boolean {
...
-        else {
+        if res is boolean {
             return res;
         }
+        // we've got a loop
+        m.isEmpty = true;
+        originalKeys = cx.listMemo.keys();
     }
     boolean isEmpty = bddEvery(cx, b, (), (), listFormulaIsEmpty);
+    if originalKeys !is () && isEmpty != m.isEmpty {
```
heshanpadmasiri: @jclark given that we need to clear some parts of the memo if we got our guess wrong in a loop I am not sure if this is any better than the [original solution](https://github.com/ballerina-platform/nballerina/issues/1065#issuecomment-1236024804).

jclark: Why do we need this `originalKeys` stuff? I thought everything worked without that.

jclark: Don't understand what you are doing with `originalKeys` at all.

jclark: Your comment here doesn't have that https://github.com/ballerina-platform/nballerina/issues/1065#issuecomment-1236092188

jclark: Can you please add a description of the proposed algorithm to #1065?

heshanpadmasiri: Add https://github.com/ballerina-platform/nballerina/issues/1065#issuecomment-1236287740. This is the implementation (not sure this is the best way to do it) of the part in my original comment where I say `// Clean up if guess is wrong`

jclark:
Can you just revert #1066?

PR #1080, opened 2022-09-15
Integrate cells into mapping type

### LINKED ISSUE #1079: Integrate cells into mapping type
**Description:**
$subject.

lochana-chathura:
## Purpose
$subject.
Fixes #1079


compiler/modules/types/core.bal
```diff
@@ -97,6 +97,15 @@ public isolated class Env {
         }
     }
 
+    isolated function cellAtomType(Atom atom) returns CellAtomicType {
+        if atom is RecAtom {
```
jclark: Don't need this: just use the cast below.

compiler/modules/front/expr.bal
```diff
@@ -972,18 +972,18 @@ function selectMappingInherentType(ExprContext cx, t:SemType expectedType, s:Map
...
     return [semType, mat];
 }
 
-function mappingAlternativeAllowsFields(t:MappingAlternative alt, string[] fieldNames) returns boolean {
+function mappingAlternativeAllowsFields(t:Context cx, t:MappingAlternative alt, string[] fieldNames) returns boolean {
     t:MappingAtomicType? pos = alt.pos;
     if pos !is () {
         // SUBSET won't be right with record defaults
-        if pos.rest == t:NEVER {
+        if t:simpleCellAtomicType(cx, pos.rest).t == t:NEVER {
```
jclark: I think we should introduce a function in the types module that takes a mapping atomic type and returns the type inside the cell of the rest field

jclark (changes requested):
We need more consistency about where "deref" occurs in the identifier.

jclark:
Aren't we going to need to do something about roDiff?

manuranga:
`.t` seems to be too short to be appearing outside the local context.

lochana-chathura:
> `.t` seems to be too short to be appearing outside the local context.

will make it `.ty`

lochana-chathura:
> We need more consistency about where "deref" occurs in the identifier.

We have `type/s` at the end of all functions having `deref`. So I changed it to `derefType/s`

lochana-chathura:
> Aren't we going to need to do something about roDiff?

Fixed in [5920e2b](https://github.com/ballerina-platform/nballerina/pull/1080/commits/5920e2b6d50e61b70180562e4107ed80675e7914)

jclark:
> We have `type/s` at the end of all functions having `deref`. So I changed it to `derefType/s` 

I don't think this is a good approach.  Sometimes we have XyzType where Xyz qualified Type. For example, AtomicType.  Inserting Deref between Xyz and Type loses the connection between Xyz and Type. For example, AtomicDerefType makes no sense.  In this case, Atomic and Type need to be put together.

compiler/modules/types/cell.bal
```diff
@@ -23,36 +23,26 @@ function cellSubtypeIsEmpty(Context cx, SubtypeData t) returns boolean {
 }
 
 function cellFormulaIsEmpty(Context cx, Conjunction? posList, Conjunction? negList) returns boolean {
-    SemType combined;
-    CellMutability minMut;
+    CellAtomicType combinedAtomicType;
     if posList == () {
-        combined = TOP;
-        minMut = CELL_MUT_UNLIMITED;
+        combinedAtomicType = { ty:TOP, mut:CELL_MUT_UNLIMITED };
```
jclark: I think we call this something like `intersected` with other basic types.

jclark (changes requested):
We should have more tests.

compiler/modules/types/mapping.bal
```diff
@@ -32,6 +32,13 @@ public class MappingDefinition {
     }
 
     public function define(Env env, Field[] fields, SemType rest) returns SemType {
+            Field[] cellFields = from Field f in fields select [f[0], cellContaining(env, f[1], CELL_MUT_LIMITED)];
+            SemType restCell = cellContaining(env, rest, CELL_MUT_LIMITED);
+            return self.defineWithoutWrapper(env, cellFields, restCell);
+    }
+
+    // Calling directly will not have the members wrapped in cells
+    public function defineWithoutWrapper(Env env, Field[] fields, SemType rest) returns SemType {
```
jclark: This should be called `define` and the function that wraps should be called something else. This is for consistency for the accessing functions, where the unwrapping function uses deref.

jclark (changes requested):

compiler/modules/types/mapping.bal
```diff
@@ -32,6 +32,13 @@ public class MappingDefinition {
     }
 
     public function define(Env env, Field[] fields, SemType rest) returns SemType {
```
jclark: I don't think this function should be a method on MappingDefintion. (Idea is that mapping knows nothing about cells.) Rather it should be in core.bal

compiler/modules/types/cell.bal
lochana-chathura: Used `combined` similar to other places.

compiler/modules/types/mapping.bal
lochana-chathura: Fixed.

compiler/modules/front/expr.bal
```diff
@@ -931,15 +932,15 @@ function codeGenMappingConstructor(ExprContext cx, bir:BasicBlock bb, t:SemType?
         }
         fieldPos[name] = f.startPos;
         if mat.names.indexOf(name) == () {
-            if mat.rest == t:NEVER {
+            if t:cellDeref(tc, mat.rest) == t:NEVER {
```
lochana-chathura: Not just for the mapping `rest`, There are several places we need to basically get the `SemType`(of BT_CELL) and return its deref type. So added the function `cellDeref()` and used it for mapping atom rest type deref instead of having a separate function for rest alone.

compiler/modules/types/mapping.bal
lochana-chathura: Please check with the latest changes.

compiler/modules/types/core.bal
```diff
@@ -705,21 +706,7 @@ function maybeRoDiff(SemType t1, SemType t2, Context? cx) returns SemType {
...
-            else if data2 == () {
-                data = data1;
-            }
-            else {
-                var diff = ops[code].diff;
-                data = diff(data1, data2);
-            }
-        }
-        else {
+        if cx != () && (code == BT_LIST || code == BT_TABLE) {
```
lochana-chathura, line 720: Needed to keep the `cx != ()` check as we call `isEmpty()` inside. Can get rid of the cx, once the list and table are integrated with cells.

compiler/modules/types/core.bal
```diff
@@ -97,6 +97,10 @@ public isolated class Env {
         }
     }
 
+    isolated function cellAtomType(Atom atom) returns CellAtomicType {
```
jclark: This shouldn't be a member function of Env, because it doesn't use the state in Env.  If you make it a module-level function, then I think there will be several other functions that do not need a Context or Env argument.

compiler/modules/types/mapping.bal
```diff
@@ -138,12 +136,12 @@ function mappingInhabited(Context cx, TempMappingSubtype pos, Conjunction? negLi
             // so we can move on to the next one
 
             // Deal the easy case of two closed records fast.
-            if isNever(pos.rest) && isNever(neg.rest) {
+            if isNever(cellDeref(cx, pos.rest)) && isNever(cellDeref(cx, neg.rest)) {
```
jclark: I would add an isNeverDeref function (which actually won't need a `cx` argument).

compiler/modules/types/core.bal
```diff
@@ -306,6 +302,13 @@ public type ComplexSemType readonly & record {|
     ProperSubtypeData[] subtypeDataList;
 |};
 
+// This is to represent a SemType belonging to cell basic type
+public type MemberSemType readonly & record {|
```
jclark: I thought we decided that we were weren't going to do this and would instead just do

```
type MemberSemType SemType;
```

Then use MemberSemType whenever a SemType satisfied the constraint that if it contains anything beginning to the Cell basic type, then it is a single Node cell Bdd.

compiler/modules/types/core.bal
```diff
@@ -649,6 +653,12 @@ public function intersect(SemType t1, SemType t2) returns SemType {
     return createComplexSemType(all, subtypes);    
 }
 
+public function intersectMemberSemTypes(Env env, MemberSemType t1, MemberSemType t2) returns MemberSemType {
```
jclark, line 652: We were going to change this to 

- check if the MemberSemType contains anything from Cell
- if it does, do the special thing for Cells
- otherwise, do regular intersection

compiler/modules/types/cell.bal
```diff
@@ -7,52 +7,42 @@ public const CELL_MUT_UNLIMITED = 2;
...
 |};
 
-public function cellContaining(Env env, SemType t, CellMutability mut) returns SemType {
-    CellAtomicType atomicCell = { t, mut };
+public function cellContaining(Env env, SemType ty, CellMutability mut) returns SemType {
+    CellAtomicType atomicCell = { ty, mut };
     Atom atom = env.cellAtom(atomicCell);
     BddNode bdd = bddAtom(atom);
-    return createComplexSemType(0, [[BT_CELL, bdd]]);
+    return basicSubtype(BT_CELL, bdd);
```
jclark, line 18: This should return MemberSemType.

compiler/modules/types/mapping.bal
```diff
@@ -9,6 +9,10 @@ public type MappingAtomicType readonly & record {|
     SemType rest;
 |};
 
+public function mappingAtomicTypeMemberAtDeref(MappingAtomicType mat, string k) returns SemType {
```
jclark, line 12: In MappingAtomicType should be using MemberSemType.

jclark (changes requested):

compiler/modules/types/core.bal
lochana-chathura: I added these changes with integrating cells into list type PR. I'll add it here as well. :)

jclark: Oh I see. Sorry for the confusion.

compiler/modules/types/core.bal
```diff
@@ -653,10 +649,16 @@ public function intersect(SemType t1, SemType t2) returns SemType {
     return createComplexSemType(all, subtypes);    
 }
 
-public function intersectMemberSemTypes(Env env, MemberSemType t1, MemberSemType t2) returns MemberSemType {
-    // JBUG #37994 cannot use mapping binding pattern with readonly records
-    CellAtomicType cat = intersectCellAtomicType(<CellAtomicType>cellAtomicType(t1), <CellAtomicType>cellAtomicType(t2));
-    return <MemberSemType>cellContaining(env, cat.ty, cat.mut);
+public function intersectMemberTypes(Env env, SemType t1, SemType t2) returns SemType {
```
lochana-chathura: Here I have made the arguments `SemType` instead of `MemberSemType` as we support the intersection for both normal SemTypes and MemberSemTypes in this function. With that the function name `intersectMemberSemTypes` was misleading. Therefore, changed that to `intersectMemberTypes`.

PR #1086, opened 2022-09-22
Add sexpr types

manuranga:
Add recursive sexpr types
Refactor using match stmt
Part of #1031 relates to #1042


compiler/modules/types/core.bal
```diff
@@ -1015,10 +1016,30 @@ public function stringSubtype(SemType t) returns StringSubtype|boolean {
     return <boolean|StringSubtype>subtypeData(t, BT_STRING);
 }
 
+public function listSubtype(SemType t) returns Bdd {
```
jclark, line 1018: Order: list, mapping, table, xml, error, function

compiler/modules/types/error.bal
```diff
@@ -1,6 +1,6 @@
 // Implementation specific to basic type error.
 
-public function errorDetail(Context cx, SemType detail) returns SemType {
+public function errorDetail(SemType detail) returns SemType {
```
jclark, line 3: Let's not do this in this PR.

compiler/modules/types/sexprSerialize.bal
```diff
@@ -0,0 +1,370 @@
...
+|};
+type MemoTable table<Memo> key (id);
+
+type SerializationContext record {|
+    Context tc;
+    map<ts:Atom> bindings = {};
+    MemoTable memos = table[];
+|};
+
+public function sexprFormSemType(Context tc, SemType ty) returns [ts:Type, map<ts:Atom>] {
```
jclark: `From`

jclark: `map<ts:Atom>` should be an argument

jclark: toSexpr

compiler/modules/types/error.bal
manuranga: 1) Currently type sexpr parser only refer to `t:Env`, never to `t:Context`.
2) `Context cx` was only added with Lochana's changes, even then it was never used.

Therefor, I propose we remove it in (or before) this PR.

jclark (approved):

Issue #1088, opened 2022-09-26
Invalid type cycles in regex


heshanpadmasiri:
Nesting a star inside a star (`(a*)*`) produces an invalid type cycle as
```
type T1 T2 | T0;
type T2 T4 | T1;
type T4 [T3, T2];
type T3 "a";
type T0 ();

```

heshanpadmasiri:
Reopening the issue since I found another case with the same bug that was not solved by the earlier fix #1089. 
```
// (a|((b|c)*))*

type T1 T2 | T0;
type T2 T4 | T5;
type T4 [T3, T1];
type T3 "a";
type T5 T6 | T1;
type T6 T8 | T10;
type T8 [T7, T5];
type T7 "b";
type T10 [T9, T5];
type T9 "c";
type T0 ();
```

The root cause of the problem seems to be when we have a nested star or a union including a star inside a star, the type definition of the inner star will include an element that refers to the outer star. Since stars are represented as unions this ultimately leads to two unions that refer to each other.

lochana-chathura:
The test case: 
`["<", "(RgeRgF(b((OS)*cDVB)*H)*kYLy)*W", "(RgeRgF(b(((glPrQ)*OS)*cDVB)*H)*kYLy)*W"]`

added with PR #1089 failing intermittently with StackOverflow after integrating readonly cells for regex in PR #1091.
I have temporary disabled it with https://github.com/ballerina-platform/nballerina/pull/1091/commits/8e939fcca77074873bdfecfd3208cb0f65428ae8

PR #1091, opened 2022-09-29
Integrate cells into list type

### LINKED ISSUE #1085: Integrate cells into list type
**Description:**
$subject.

lochana-chathura:
## Purpose
$subject. 
Fixes #1085


compiler/modules/types/cell.bal
```diff
@@ -128,10 +128,46 @@ function intersectCellAtomicType(CellAtomicType c1, CellAtomicType c2) returns C
...
+}
+
+function cellSubtypeComplement(SubtypeData t) returns SubtypeData {
+    SubtypeData subtypeData = bddSubtypeComplement(t);
+    return cellFixSubtypeData(subtypeData);
+}
+
+// SubtypeData being a boolean would result in a BasicTypeBitSet which could either be 0 or 1 << BT_CELL.
+// This is to avoid cell type being a BasicTypeBitSet, as we don't want to lose the cell wrapper.
+function cellFixSubtypeData(SubtypeData subtypeData) returns SubtypeData {
```
lochana-chathura, line 153: Well, when we have two cells in which the diff is empty, we end up creating `0` as the resultant semType. This is contrary to having MemberSemType enforced with the following type definition.
``` bal 
public type MemberSemType readonly & record {|
    0 all = 0;
    BasicTypeBitSet some = CELL;
    ProperSubtypeData[1] subtypeDataList;
|};
```
Therefore I did this fix.

The option would be to make MemberSemType 
```bal
type MemberSemType NEVER | CELL | record {| 0 all = 0 BasicTypeBitSet some = CELL; ProperSubtypeData[1] subtypeDataList; |};
```
However, with the second option we will not have the cell wrapper all the time.

compiler/modules/types.regex/tests/data/regexInclusionTests.json
```diff
@@ -17,8 +18,6 @@
     ["<", "(AC)*F", "(A(B)*C)*F"],
     ["=", "(AB*C)*F", "(A(B)*C)*F"],
 
-    ["<", "(RgeRgF(b((OS)*cDVB)*H)*kYLy)*W", "(RgeRgF(b(((glPrQ)*OS)*cDVB)*H)*kYLy)*W"],
```
lochana-chathura, line 20: https://github.com/ballerina-platform/nballerina/issues/1088#issuecomment-1276165054

compiler/modules/types/core.bal
```diff
@@ -1008,22 +1012,36 @@ public function intSubtypeConstraints(SemType t) returns IntSubtypeConstraints?
...
 
-public function listAtomicSimpleArrayMemberType(ListAtomicType? atomic) returns BasicTypeBitSet? {
+public function defineListTypeWrapped(ListDefinition ld, Env env, SemType[] initial = [], int fixedLength = initial.length(), SemType rest = NEVER, CellMutability mut = CELL_MUT_LIMITED) returns SemType {
+    MemberSemType[] initialCells = from var i in initial select cellContaining(env, i, mut);
+    MemberSemType restCell = cellContaining(env, rest, mut);
+    return ld.define(env, initialCells, fixedLength, restCell);
+}
+
+public function tupleTypeWrapped(Env env, SemType... members) returns SemType {
+    ListDefinition def = new;
```
jclark: I would inline this and similar below.

compiler/modules/types/core.bal
```diff
@@ -1008,22 +1012,36 @@ public function intSubtypeConstraints(SemType t) returns IntSubtypeConstraints?
     } 
 }
 
-public function listAtomicSimpleArrayMemberType(ListAtomicType? atomic) returns BasicTypeBitSet? {
+public function defineListTypeWrapped(ListDefinition ld, Env env, SemType[] initial = [], int fixedLength = initial.length(), SemType rest = NEVER, CellMutability mut = CELL_MUT_LIMITED) returns SemType {
```
jclark: Wrapped/Deref isn't very symmmetric, but I don't have a better idea.

compiler/modules/types/mapping.bal
```diff
@@ -1,6 +1,8 @@
 // Implementation specific to basic type list.
 
-public type Field [string, MemberSemType];
+public type Field [string, SemType];
```
jclark, line 3: Why do we need both Field and MemberField here?

jclark (commented):
This looks very good generally.

We need to think a bit more about the extent to which List/Mapping know about cells.  We started off trying to keep these completely separate i.e. List/Mapping know nothing about cells.  But we have moved away from this, because of the need to maintain cell atomic types and because we can use readonly cells for function argument lists.

The current concept is that List/Mappings know that they have "member" semtypes but not that these semtypes are cells.  But this abstraction isn't fully maintained: list.bal accesses cells directly and has Deref functions.  It's not clear whether it is helpful to maintain distinction between MemberSemType and semtype that is a subtype of Cell.

I think we need to get clearer on this and implement a policy consistently.

Related to this is the Deref/Wrapped naming convention.

compiler/modules/types/mapping.bal
lochana-chathura: Well, in the `defineListTypeWrapped()` we take a `SemType[]` and convert it to a `CellSemType[]`.
Similarly in the `defineMappingTypeWrapped()` we take a `Field[]` and convert it to a `CellField[]`. So we need two type definitions to represent `[string, SemType]`(input) and `[string, CellSemType]`(after wrapping) The current occurrence of `CellField` is only `mapping.bal`.

compiler/modules/types/core.bal
lochana-chathura: replaced `deref` with `inner` with [1bddfa0](https://github.com/ballerina-platform/nballerina/pull/1091/commits/1bddfa0c4c379d8e38a70c5e7968a6f797dc461f) as per our discussion

Issue #1076, opened 2022-09-09
Subtype check for some complex, nested recursive types takes excessive time and stack space


heshanpadmasiri:
Running the program in extra/regex with the following arguments

- `BDGB(gSUNd(Pp(xMsTI(TcnBE)*ke)*)*S)*v`
- `BDGB(gSUNd(Pp(xMsTI((CvGb)*TcnBE)*ke)*)*S)*v`

(which tests whether the first regex is a subtype of the second) uses a lot of time and stack space.

It caused the memo table to grow to 243 (checked after [this line](https://github.com/ballerina-platform/nballerina/blob/c1a12d9d09a0a8d69211809e89fb3e7c657b058a/compiler/modules/types/common.bal#L65)) and provisional empty count to change between 21 and 22 (checked after [this line](https://github.com/ballerina-platform/nballerina/blob/c1a12d9d09a0a8d69211809e89fb3e7c657b058a/compiler/modules/types/common.bal#L55))

Could be the same cause as  #1074 but didn't caused at stack overflow in about 10-20 minutes (#1074 will cause it within few minutes)

jclark:
It seems like this technique is successfully stressing our recursive types implementation!

jclark:
I think this shows algorithm in #1065 isn't right.

jclark:
Can you add a test case for this? It should only run if long-running tests are enabled.

heshanpadmasiri:
One problem we have in the current implementation is that if we mark one type as provisional it will make every subsequent provisional if that type is empty. This don't take into account the fact the subsequent type may have no dependency on any provisionally empty type. 

One idea on how to deal with this is for a given Bdd (that comes after a provisionally empty type and we think is empty), if all the child nodes in the bdd are known to be nonprovisional then that type must be nonprovisional as well.

jclark:
As of Update 2 and with regex fixes, reproduce the problem by doing `make` in `extra/regex` and then running:

```
java -Xss100M -jar ../../build/extra/bin/nballerina.regex.jar "BDGB(gSUNd(Pp(xMsTI(TcnBE)*ke)*)*S)*v" "BDGB(gSUNd(Pp(xMsTI((CvGb)*TcnBE)*ke)*)*S)*v"
```

jclark:
This pair completes in a second:

```
[...]

PR #1092, opened 2022-10-03
Simplify subtype checks for some complex, nested recursive types

heshanpadmasiri:
## Purpose
Simplify subtype checks for some complex, nested recursive types
Related to #1076 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/types/list.bal
```diff
@@ -572,13 +572,42 @@ function listSubtypeIntersect(Context cx, SubtypeData t1, SubtypeData t2) return
...
+        ListAtomicType negAtom = cx.listAtomType(t2.atom);
+        // we only handle tuples
+        if !isNever(posAtom.rest) || !isNever(negAtom.rest) {
+            return false;
+        }
+        SemType[] posInitial = posAtom.members.initial; 
+        SemType[] negInitial = negAtom.members.initial; 
+        if posInitial.length() == negInitial.length() {
+            foreach int i in 0 ..< posInitial.length() {
+                if !isEmpty(cx, intersect(posInitial[i], negInitial[i])) {
```
jclark: Not sure about using `isEmpty` here. Don't we want `isNever`?

jclark: Is the logic right? The two lists are disjoint if any of the pairs of members are disjoint.

compiler/modules/types/list.bal
```diff
@@ -572,13 +572,42 @@ function listSubtypeIntersect(Context cx, SubtypeData t1, SubtypeData t2) return
...
         //io:println("fresh BDD: ", bddToString(b));
     }
     return b;
 }
 
+function isDisjoint(Context cx, Bdd t1, Bdd t2) returns boolean {
+    if t1 !is BddNode || t2 !is BddNode {
+        return false;
+    }
+    if t1.left is true && t1.middle is false && t1.right is false &&
```
jclark: I would expect we would need to handle BDDs that are unions of positive atoms.

heshanpadmasiri: fixed

compiler/modules/types/list.bal
heshanpadmasiri: fixed

jclark (commented):
This doesn't seem to me the best implementation strategy. Instead I suggest calling intersect, and then have a isEmptyShortcut function that returns `boolean?`, which does a simplified check for emptiness and returns true/false if it is known empty, non-empty, and `()` if it is a case that it cannot handle.
 
It aims to handle cases with only positive atoms. It can just intersect member types for each index; if any of those are never, it can return true.

jclark:
You can implement this with bddEvery, where the predicate checks whether one of the member indices has an index for which the intersection of the positive member types is never.

heshanpadmasiri:
> You can implement this with bddEvery, where the predicate checks whether one of the member indices has an index for which the intersection of the positive member types is never.

Fixed

compiler/modules/types/list.bal
```diff
@@ -582,58 +582,44 @@ function listSubtypeDiff(Context cx, SubtypeData t1, SubtypeData t2) returns Sub
     return b;
 }
 
-function isDisjoint(Context cx, Bdd t1, Bdd t2) returns boolean {
-    ListAtomicType[]? posAtoms = bddListAtoms(cx, t1);
-    ListAtomicType[]? negAtoms = bddListAtoms(cx, t2);
-    if posAtoms is () || negAtoms is () {
-        return false;
+function isTupleDisjointSimple(Context cx, Bdd t1, Bdd t2) returns boolean? {
```
jclark: Make it `listIsEmptySimple(Context cx, Bdd) returns boolean?`.

compiler/modules/types/list.bal
```diff
@@ -582,58 +582,44 @@ function listSubtypeDiff(Context cx, SubtypeData t1, SubtypeData t2) returns Sub
...
+    if t1 !is BddNode || t1.left !is true || t1.middle !is false || t1.right !is false {
```
jclark: Don't need to be this restrictive.

Just intersect the positive atoms, and if we get empty, return true.

compiler/modules/types/list.bal
jclark: Note that we can memoize the result when it's true.

heshanpadmasiri: fixed

compiler/modules/types/list.bal
heshanpadmasiri: fixed

compiler/modules/types/list.bal
```diff
@@ -572,13 +572,59 @@ function listSubtypeIntersect(Context cx, SubtypeData t1, SubtypeData t2) return
 }
 
 function listSubtypeDiff(Context cx, SubtypeData t1, SubtypeData t2) returns SubtypeData {
+    if listIsEmptySimple(cx, bddIntersect(noBddCache, <Bdd> t1, <Bdd> t2)) {
```
jclark: Why `noBddCache`?

compiler/modules/types/list.bal
```diff
@@ -572,13 +572,59 @@ function listSubtypeIntersect(Context cx, SubtypeData t1, SubtypeData t2) return
...
+        return t1;
+    }
     Bdd b = memoSubtypeDiff(cx.listMemo, <Bdd>t1, <Bdd>t2);
     if b != false && cx.listMemo[b] == () {
         //io:println("fresh BDD: ", bddToString(b));
     }
     return b;
 }
 
+function listIsEmptySimple(Context cx, Bdd bdd) returns boolean {
```
jclark: This is confusing because returning false doesn't mean it's not empty.

I think it's conceptually simple to have this return `boolean?` even if we don't yet ever return false.

compiler/modules/types/list.bal
```diff
@@ -572,13 +572,59 @@ function listSubtypeIntersect(Context cx, SubtypeData t1, SubtypeData t2) return
...
+    BddMemo? m = cx.listMemo[bdd];
+    if m !is () && m.empty is boolean {
+        return <boolean>m.empty;
+    }
+    boolean empty = bddEvery(cx, bdd, (), (), isTupleIntersectionEmptySimple);
+    if empty is true {
```
jclark: I prefer `==` rather than `is` for testing a value is equal to a constant.

compiler/modules/types/list.bal
```diff
@@ -572,13 +572,59 @@ function listSubtypeIntersect(Context cx, SubtypeData t1, SubtypeData t2) return
...
+        current = current.next;
+    }
+    return false;
+}
+
+function tupleMembers(ListAtomicType atom) returns SemType[]? {
+    if !isNever(atom.rest) {
+        return ();
+    }
+    return atom.members.initial;
```
jclark: Cannot be right to ignore members.fixedLength.

compiler/modules/types/list.bal
```diff
@@ -572,13 +572,59 @@ function listSubtypeIntersect(Context cx, SubtypeData t1, SubtypeData t2) return
...
+    if memberTypes is () {
+        return false;
+    }
+    Conjunction? current = pos.next;
+    while current !is () {
+        SemType[]? currentMemberTypes = tupleMembers(cx.listAtomType(current.atom));
+        if currentMemberTypes is () || currentMemberTypes.length() != memberTypes.length() {
+            return false;
+        }
+        foreach int i in 0 ..< currentMemberTypes.length() {
```
jclark: We ought to be using fixedArrayGet here.

compiler/modules/types/list.bal
```diff
@@ -572,13 +572,59 @@ function listSubtypeIntersect(Context cx, SubtypeData t1, SubtypeData t2) return
...
+    if neg !is () || pos is () {
+        return false;
+    }
+    SemType[]? memberTypes = tupleMembers(cx.listAtomType(pos.atom));
+    if memberTypes is () {
+        return false;
+    }
+    Conjunction? current = pos.next;
+    while current !is () {
+        SemType[]? currentMemberTypes = tupleMembers(cx.listAtomType(current.atom));
```
jclark: Can't we use listIntersectWith here?

compiler/modules/types/list.bal
```diff
@@ -572,13 +572,59 @@ function listSubtypeIntersect(Context cx, SubtypeData t1, SubtypeData t2) return
...
+        return <boolean>m.empty;
+    }
+    boolean empty = bddEvery(cx, bdd, (), (), isTupleIntersectionEmptySimple);
+    if empty is true {
+        cx.listMemo.put({ bdd, empty }); // handle both memoized values being () and not in memo
+    }
+    return empty;
+}
+
+function isTupleIntersectionEmptySimple(Context cx, Conjunction? pos, Conjunction? neg) returns boolean {
```
jclark: This name isn't consistent with the rest of this file. This is trying to do listFormulaIsEmpty in a simple way.

This needs to return boolean to work with bddEvery. So I would call it listFormulaIsDefinitelyEmpty. This reflects the semantics: is is true then it definitely is empty; if false, then it's not definitely empty (but may prove to be empty later).

jclark (commented):
I had some pending comments left over from yesterday that I forgot to submit.

compiler/modules/types/list.bal
heshanpadmasiri: fixed

compiler/modules/types/list.bal
heshanpadmasiri: fixed

compiler/modules/types/list.bal
heshanpadmasiri: fixed

compiler/modules/types/list.bal
heshanpadmasiri: fixed. (with the use of `listIntersectWtih` we will get `()` if the `fixedLengths` are different (we have checked both `rest` fields tobe `never`)

compiler/modules/types/list.bal
heshanpadmasiri: fixed

compiler/modules/types/list.bal
heshanpadmasiri: fixed

compiler/modules/types/list.bal
heshanpadmasiri: fixed

compiler/modules/types/list.bal
```diff
@@ -572,13 +572,60 @@ function listSubtypeIntersect(Context cx, SubtypeData t1, SubtypeData t2) return
...
+        var { members: currentMembers, rest: currentRest } = cx.listAtomType(current.atom);
+        if !isNever(currentRest) {
+            return false;
+        }
+        var intersection = listIntersectWith(cx, members, rest, currentMembers, currentRest); 
+        if intersection is () {
+            return false;
+        }
+        FixedLengthArray intersectionMembers = intersection[0];
+        foreach int i in 0 ..< memberCount {
```
jclark: Let's split this into a separate function fixedArrayDefinitelyEmpty

compiler/modules/types/list.bal
```diff
@@ -572,13 +572,60 @@ function listSubtypeIntersect(Context cx, SubtypeData t1, SubtypeData t2) return
...
 }
 
+function listIsEmptySimple(Context cx, Bdd bdd) returns boolean? {
+    BddMemo? m = cx.listMemo[bdd];
+    if m !is () && m.empty is boolean {
+        return <boolean>m.empty;
+    }
+    boolean empty = bddEvery(cx, bdd, (), (), listFormulaIsDefinitelyEmpty);
+    if empty == true {
+        cx.listMemo.put({ bdd, empty }); // handle both memoized values being () and not in memo
```
jclark, line 578: More idiomatic: `cx.listMemo[bdd] = empty;`

compiler/modules/types/list.bal
```diff
@@ -572,13 +572,60 @@ function listSubtypeIntersect(Context cx, SubtypeData t1, SubtypeData t2) return
...
+    return ();
+}
+
+function listFormulaIsDefinitelyEmpty(Context cx, Conjunction? pos, Conjunction? neg) returns boolean {
+    if neg !is () || pos is () {
+        return false;
+    }
+    var { members, rest } = cx.listAtomType(pos.atom);
+    int memberCount = members.fixedLength;
+    if !isNever(rest) {
```
jclark: We should figure out the right requirement on `rest`.

I have `S = [S1, Sr...]`, and `T = [T1, Tr...]` and `S1&T1` is never, then `S&T` is empty.

compiler/modules/types/list.bal
heshanpadmasiri: I don't think this is possible due `cannot update 'table<wso2/nballerina.types:0.1.0:BddMemo> key(bdd)' with member access expression` error

jclark: My suggestion was wrong.

compiler/modules/types/list.bal
heshanpadmasiri: fixed

compiler/modules/types/list.bal
heshanpadmasiri: fixed. (I think as long as any member of the intersection is never intersection is empty. Originally I had this check to limit this to tuples only)

heshanpadmasiri:
@jclark I removed the `bddIntersect` optimization (if we intersect just two lists then we can calculate the resulting list eagerly) since it is unrelated to this specific optimization. I will send that as a separate pr.

compiler/modules/types/list.bal
```diff
@@ -572,11 +560,56 @@ function listSubtypeIntersect(Context cx, SubtypeData t1, SubtypeData t2) return
 }
 
 function listSubtypeDiff(Context cx, SubtypeData t1, SubtypeData t2) returns SubtypeData {
-    Bdd b = memoSubtypeDiff(cx.listMemo, <Bdd>t1, <Bdd>t2);
-    if b != false && cx.listMemo[b] == () {
-        //io:println("fresh BDD: ", bddToString(b));
+    MemoBddCache cache = new (cx.listMemo);
+    // This is checking if t1 and t2 are disjoint (i.e intersection(t1, t2) is empty). If so diff(t1, t2) == t1.
+    if listIsEmptySimple(cx, bddIntersect(cache, <Bdd> t1, <Bdd> t2)) is true {
```
jclark: Prefer `== true`.

compiler/modules/types/list.bal
```diff
@@ -572,11 +560,56 @@ function listSubtypeIntersect(Context cx, SubtypeData t1, SubtypeData t2) return
...
+        return t1;
+    }
+    return memoSubtypeDiff(cx.listMemo, <Bdd>t1, <Bdd>t2);
+}
+
+function listIsEmptySimple(Context cx, Bdd bdd) returns boolean? {
+    BddMemo? m = cx.listMemo[bdd];
+    if m !is () && m.empty is boolean {
```
jclark: `!= ()`

compiler/modules/types/list.bal
```diff
@@ -572,11 +560,56 @@ function listSubtypeIntersect(Context cx, SubtypeData t1, SubtypeData t2) return
...
+    boolean empty = bddEvery(cx, bdd, (), (), listFormulaIsDefinitelyEmpty);
+    if empty == true {
+        cx.listMemo.put({ bdd, empty }); // handle both memoized values being () and not in memo
+        return true;
+    }
+    return ();
+}
+
+function listFormulaIsDefinitelyEmpty(Context cx, Conjunction? pos, Conjunction? neg) returns boolean {
+    if neg !is () || pos is () {
```
jclark: As a matter of style, please use `==` instead of `is` for comparing against constants.

jclark (changes requested):
Why has `function.bal` changed?

compiler/modules/types/list.bal
heshanpadmasiri: fixed

compiler/modules/types/list.bal
heshanpadmasiri: fixed

compiler/modules/types/list.bal
heshanpadmasiri: fixed

heshanpadmasiri:
> Why has `function.bal` changed?

Fixed (It was a new  line)

compiler/modules/types/list.bal
```diff
@@ -572,11 +560,56 @@ function listSubtypeIntersect(Context cx, SubtypeData t1, SubtypeData t2) return
 }
 
 function listSubtypeDiff(Context cx, SubtypeData t1, SubtypeData t2) returns SubtypeData {
-    Bdd b = memoSubtypeDiff(cx.listMemo, <Bdd>t1, <Bdd>t2);
-    if b != false && cx.listMemo[b] == () {
-        //io:println("fresh BDD: ", bddToString(b));
+    MemoBddCache cache = new (cx.listMemo);
+    // This is checking if t1 and t2 are disjoint (i.e intersection(t1, t2) is empty). If so diff(t1, t2) == t1.
+    if listIsEmptySimple(cx, bddIntersect(cache, <Bdd> t1, <Bdd> t2)) == true {
```
heshanpadmasiri, line 565: @jclark we can do this check similar to `simpleIntersect` in the [other PR](https://github.com/ballerina-platform/nballerina/blob/d12899fb5bf882448e0c7f0be895a2e9b6f2b37d/compiler/modules/types/bdd.bal#L85). Not sure which one is more preferable (do that change similar to this i.e in `listSubtypeIntersect` or do this check similar to that in `bddDiff`)

jclark: I think it's better for the complexity to be in the functions that test emptiness.

compiler/modules/types/list.bal
```diff
@@ -572,11 +560,56 @@ function listSubtypeIntersect(Context cx, SubtypeData t1, SubtypeData t2) return
...
+    if neg != () || pos == () {
+        return false;
+    }
+    var { members, rest } = cx.listAtomType(pos.atom);
+    Conjunction? current = pos.next;
+    while current != () {
+        var { members: currentMembers, rest: currentRest } = cx.listAtomType(current.atom);
+        var intersection = listIntersectWith(cx, members, rest, currentMembers, currentRest); 
+        if intersection == () {
+            return false;
```
jclark: Shouldn't this be `true`. listIntersectWith returns nil when the result is empty because the list lengths are inconsistent.

jclark (changes requested):

compiler/modules/types/list.bal
heshanpadmasiri: fixed

PR #1103, opened 2022-10-23
Better detect recursive types that only contain infinite shapes

### LINKED ISSUE #1070: Better detect recursive types that only contain infinite shapes
With #1065 we should be able to detect such cases without the need for deferring emptiness checks as done in #1062

heshanpadmasiri:
## Purpose
Better detect recursive types that only contain infinite shapes
Resolves #1070

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/types/boolean.bal
```diff
@@ -54,5 +54,6 @@ final BasicTypeOps booleanOps = {
     intersect: booleanSubtypeIntersect,
     diff: booleanSubtypeDiff,
     complement: booleanSubtypeComplement,
-    isEmpty: notIsEmpty
+    isEmpty: notIsEmpty,
```
jclark: This is too confusing.

compiler/modules/types/core.bal
```diff
@@ -170,12 +170,13 @@ function cellAtomType(Atom atom) returns CellAtomicType {
...
 
-// See memoSubtypeIsEmpty for what these mean.
-type MemoEmpty boolean|"provisional"|();
+// See memoSubtypeIsFiniteAndEmpty for what these mean.
+type MemoEmpty boolean|"loop"|"provisional"|();
 
 type BddMemo record {|
     readonly Bdd bdd;
     MemoEmpty empty = ();
+    boolean finite = true;
```
jclark: Why is this a separate field rather than another value for `empty`?

compiler/modules/types/core.bal
```diff
@@ -290,6 +291,7 @@ type BasicTypeOps readonly & record {|
     BinOp diff = binOpPanic;
     UnaryOp complement = unaryOpPanic;
     UnaryTypeCheckOp isEmpty = unaryTypeCheckOpPanic;
+    UnaryTypeCheckOp isFinite = unaryTypeCheckOpPanic;
```
jclark: This doesn't feel right to me. Do we need isFinite for all basic types?

jclark (changes requested):

compiler/modules/types/boolean.bal
heshanpadmasiri: fixed.

compiler/modules/types/core.bal
heshanpadmasiri: fixed.

compiler/modules/types/core.bal
heshanpadmasiri: fixed. (We only need this for list and mapping types)

compiler/modules/types/core.bal
```diff
@@ -171,7 +172,10 @@ function cellAtomType(Atom atom) returns CellAtomicType {
 }
 
 // See memoSubtypeIsEmpty for what these mean.
```
jclark, line 179: Adding the comment here is not consistent with this line.

compiler/modules/types/common.bal
```diff
@@ -52,19 +56,29 @@ function memoSubtypeIsEmpty(Context cx, BddMemoTable memoTable, BddIsEmptyPredic
...
         m.empty = isEmpty;
     }
+    if cyclic {
+        m.empty = "cyclic";    
+    }
     return isEmpty;
 }
 
+function memoSubtypeIsCyclic(Context cx, BddMemoTable memoTable, BddIsEmptyPredicate isEmptyPredicate, Bdd b) returns boolean {
+    // This assume we have already checked (and confirmed) type to be empty
```
jclark: It only assumes that we have tested (and so memoized) whether the type is empty.

compiler/modules/types/common.bal
```diff
@@ -34,11 +34,15 @@ function memoSubtypeIsEmpty(Context cx, BddMemoTable memoTable, BddIsEmptyPredic
     BddMemo m;
```
jclark, line 34: We should add a comment here explaining the values of MemoEmpty,

compiler/modules/types/core.bal
```diff
@@ -171,7 +172,10 @@ function cellAtomType(Atom atom) returns CellAtomicType {
...
-type MemoEmpty boolean|"provisional"|();
+// loop means type has a recursive shape
```
jclark: These are not adequate explanations of the meaning of the values.

compiler/modules/types/list.bal
```diff
@@ -132,6 +132,10 @@ function listSubtypeIsEmpty(Context cx, SubtypeData t) returns boolean {
     return memoSubtypeIsEmpty(cx, cx.listMemo, listBddIsEmpty, <Bdd>t);
 }
 
+function listSubtypeIsCyclic(Context cx, SubtypeData t) returns boolean {
```
jclark: Better to be listBddIsCyclic and take a Bdd, similarly for mapping.

compiler/modules/types/core.bal
```diff
@@ -778,6 +782,24 @@ public function isNever(SemType t) returns boolean {
     return t is BasicTypeBitSet && t == 0;
 }
 
+public function isCyclic(Context cx, SemType t) returns boolean {
```
jclark, line 756: I don't like the implementation of this function. You're doing similar stuff is isMemberNever.

We need to factor out functions for trying to do conversions between a SemType to an single basic type/Bdd/atom.

heshanpadmasiri: fixed

compiler/modules/types/list.bal
heshanpadmasiri: fixed

compiler/modules/types/common.bal
heshanpadmasiri: fixed

compiler/modules/types/core.bal
heshanpadmasiri: fixed

compiler/modules/types/common.bal
heshanpadmasiri: fixed

compiler/modules/types/core.bal

compiler/modules/types/common.bal
```diff
@@ -52,19 +59,29 @@ function memoSubtypeIsEmpty(Context cx, BddMemoTable memoTable, BddIsEmptyPredic
     int initStackDepth = cx.memoStack.length();
     cx.memoStack.push(m);
     boolean isEmpty = isEmptyPredicate(cx, b);
+    boolean cyclic = isEmpty && (m.empty == "loop");
```
jclark: Clearer to just save isLoop here.

compiler/modules/types/common.bal
```diff
@@ -52,19 +59,29 @@ function memoSubtypeIsEmpty(Context cx, BddMemoTable memoTable, BddIsEmptyPredic
...
     if !isEmpty || initStackDepth == 0 {
         foreach int i in initStackDepth + 1 ..< cx.memoStack.length() {
             MemoEmpty memoEmpty = cx.memoStack[i].empty;
-            if memoEmpty == "provisional" {
+            if memoEmpty is "provisional"|"loop"|"cyclic" {
                 cx.memoStack[i].empty = isEmpty ? isEmpty : ();
             }
         }
         cx.memoStack.setLength(initStackDepth);
         m.empty = isEmpty;
```
jclark: m.empty = isLoop && isEmpty ? "cyclic" : isEmpty;

compiler/modules/types/common.bal
```diff
@@ -34,11 +34,18 @@ function memoSubtypeIsEmpty(Context cx, BddMemoTable memoTable, BddIsEmptyPredic
     BddMemo m;
     if mm != () {
         MemoEmpty res = mm.empty;
+        if res == "cyclic" {
+            // These are types with no non-recursive non-empty shapes (ex: A = [A];)
+            // Since we define types inductively we consider these to be empty
```
jclark, line 38: Say we consider cyclic types to be empty because we define types inductively.

compiler/modules/types/common.bal
jclark: The only way that we have found that this can be empty is by going through a loop. This means that the shapes in the type would all be infinite. But we define types inductively, which means we only consider finite shapes.

compiler/modules/types/common.bal
heshanpadmasiri: fixed

compiler/modules/types/common.bal
heshanpadmasiri: fixed

compiler/modules/types/common.bal
heshanpadmasiri: fixed

jclark (approved):

Issue #1101, opened 2022-10-21
Fix nightly builds


manuranga:
<img width="927" alt="Screenshot 2022-10-21 at 1 23 08 PM" src="https://user-images.githubusercontent.com/1686124/197143179-c23b56b3-c102-4d08-b9e7-1c68c9ff33b4.png">

manuranga:
Looks like [ballerina-distribution/daily-build-2201.0.x.yml](https://github.com/ballerina-platform/ballerina-distribution/actions/workflows/daily-build-2201.0.x.yml) is deleted and [ballerina-distribution/daily-build-2201.1.x.yml](https://github.com/ballerina-platform/ballerina-distribution/actions/workflows/daily-build-2201.1.x.yml) has been failing for a month. So no point of having these at our end.

@keizer619 Please provide us a list of `ballerina-distribution` workflows to rely on.

In absence of the correct versions, these two builds are currently running on some random build form dist, which happens to be `ballerina-2201.2.2-SNAPSHOT-swan-lake-linux-x64`, with following error.
```
[fail] typeTests#'int.typetest#6':

something impossible happened (unreached in semTypeFromSexpr for sexpr:["!",["int",[8,8]]])

	at wso2.nballerina.comm.err.0:impossible(err.bal:37)
	wso2.nballerina.types.sexpr.0:semTypeFromSexpr(convert.bal:64)
	wso2.nballerina.types.sexpr.0:semTypeFromSexpr(convert.bal:39)
	wso2.nballerina.types.sexpr$test.0.tests.typeTest:typeTests(tests/typeTest.bal:51)
```
Sent a PR for this #1102.

Attempting to reproduce this error locally, I tested with a `ballerina-2201.2.2-SNAPSHOT-swan-lake` but it is giving me a different (more severe) error.

https://github.com/ballerina-platform/ballerina-lang/issues/38339

I couldn't test the PR locally due to above, when I tried to test it in GitHub actions, it gave yet another error:
`ERROR [tests/exprBinaryDiv.bal:(29:55,29:74)] '9223372036854775808' is out of range for 'int'`

manuranga:
Had a chat with @keizer619 We decided test nBallerina during daily full build of jBallerina. Please add following steps to the build
```sh
git clone --depth=1 git@github.com:ballerina-platform/nballerina.git
(cd nballerina/compiler && path/to/newly/created/bal test)
```

[...]

PR #1104, opened 2022-10-24
Fix nightly - refactor int min/max to consts

manuranga:
Nightly jballerina doesn't allow `-9223372036854775808` anymore.

https://github.com/ballerina-platform/ballerina-spec/issues/902#issuecomment-975150745
https://github.com/ballerina-platform/ballerina-lang/issues/32577


manuranga:
Part of #1101

PR #1112, opened 2022-11-09
Compile to native image with GraalVM

### LINKED ISSUE #1082: Compile to native image with GraalVM
See if we can compile nballerina using the support for GraalVM that is being worked on by jBallerina team.

heshanpadmasiri:
## Purpose
Compile to native image with GraalVM
Resolves #1082 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


heshanpadmasiri:
@jclark I updated the llvm version used by jni from 12 to 13 due to a bug in the jni side (the intrinsic `llvm.ptrmask.p1i8.i64` is not available in the bytedeco version of llvm 12). Besides that, I have 2 concerns that I am not sure how important to us at the moment

1. When running the native image it produces a warning `Warning: Version of org.bytedeco:llvm could not be found.` Though this doesn't seem to cause any other issues (it passes all our tests cases except two mentioned [here](https://github.com/ballerina-platform/nballerina/issues/1082#issuecomment-1306998115))
2. Based on [this problem](https://github.com/ballerina-platform/nballerina/issues/1082#issuecomment-1306998115) as well as the fact we need `-lm` flag [here](https://github.com/ballerina-platform/nballerina/blob/ce995c42d0d407fe518581410d64f7103dace524/test/native-sub.mk#L56) it looks to me like some of the optimizations we get when we run our normal build with `-O2` are not happening with the object files we are getting (We have these problems in the normal build only if we use `-O0`)

test/sub.mk
```diff
@@ -11,8 +11,9 @@
 # You can do `make -f ../../sub.mk tdir=$(basename "$PWD") testIntermediate` to test the ll/o files.
 # Failing tests are listed in fail.txt
 COMPILER_JAR=../../../build/compiler/bin/nballerina.jar
-OBJECT_COMPILER=../../../testbuild/target/bin/nballerina.jar
+OBJECT_COMPILER ?=../../../testbuild/target/bin/nballerina.jar
 COMPILER_OUTPUT ?= ll
+WHICH_OBJECT_COMPILER = $(suffix $(OBJECT_COMPILER))
```
jclark: The `WHICH_` prefix is too weird.

test/sub.mk
```diff
@@ -1,6 +1,6 @@
 # The default target is `test`, which runs in 3 phases
 # 1. Remove the compile.stamp file if the compiler jar has changes
-# 2. If COMPILER_OUTPUT is ll: compile the .bal test cases into .ll files 
+# 2. If COMPILER_OUTPUT_EXTENSION is .ll: compile the .bal test cases into .ll files
```
jclark: How does this get set? Does the user have to say `make COMPILER_OUTPUT_EXTENSION=.o" in the JNI case?

heshanpadmasiri: Yes
For running tests with JNI (jar file, since it is the default), it is `make test COMPILER_OUTPUT_EXTENSION=.o` 
If you need to use the executable (one from Graal VM) file, it also needs to be specified as `make test COMPILER_OUTPUT_EXTENSION=.o O_COMPILER=../../../testbuild/target/bin/nballerina`

heshanpadmasiri: Changed above two commands to `make test LINK_FILE_EXTENSION=.o` and `make test LINK_FILE_EXTENSION=.o O_COMPILER=../../../testbuild/target/bin/nballerina` respectively

jclark (approved):

PR #1118, opened 2022-11-11
Add ability to test the jni version of the compiler against the test suite

### LINKED ISSUE #1117: Add ability to test the jni version of the compiler against the test suite
Currently, jni version of the compiler (compiler that uses `jni.llvm` instead of `print.llvm`), is tested only against its own unit tests. It is better if we have an on-demand test that can test it against the whole test suite similar to how we test the default version of the compiler

I don't think it makes that much sense to have this test run automatically, especially with #1112 we will need a lot of resources and time to compile the native version (which we will need to support as a third option along with this) and unit tests in `jni.llvm` are suppose to make sure both that as well as `print.llvm` generate the same llvm-ir.

heshanpadmasiri:
## Purpose
Add the ability to test the JNI version of the compiler against the test suite
Resolves #1117

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/test/sub.mk
+++ b/test/sub.mk
@@ -1,6 +1,6 @@
 # The default target is `test`, which runs in 3 phases
 # 1. Remove the compile.stamp file if the compiler jar has changes
-# 2. If COMPILE_MODE is default: compile the .bal test cases into .ll files 
+# 2. If COMPILER_OUTPUT is ll: compile the .bal test cases into .ll files 
 # 	 Else: compile the .bal test cases into .o files (using the jni version of the compiler)
 # 3. Compile, execute and check the output of .ll/.o files
 # Phase 2 only updates .ll if they have changed.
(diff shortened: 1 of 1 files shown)


test/sub.mk
```diff
@@ -1,15 +1,18 @@
...
 # This is used in phases 1 and 2.
 # To run this, first navigate to out/<category> dir, then
-# You can do `make -f ../../sub.mk tdir=$(basename "$PWD") compile` to compile all changed test cases from .bal to .ll
-# You can do `make -f ../../sub.mk tdir=$(basename "$PWD") testll` to test the ll files.
+# You can do `make -f ../../sub.mk tdir=$(basename "$PWD") compile` to compile all changed test cases from .bal to .ll/.o files
+# You can do `make -f ../../sub.mk tdir=$(basename "$PWD") testIntermediate` to test the ll/o files.
 # Failing tests are listed in fail.txt
 COMPILER_JAR=../../../build/compiler/bin/nballerina.jar
+JNI_COMPILER_JAR=../../../testbuild/target/bin/nballerina.jar
+COMPILE_MODE ?= default
```
jclark: Use a variable name and value that makes use of kind of output produced: .ll or .o

jclark (approved):

PR #1120, opened 2022-11-14
Avoid generating registers for invalid narrow vars

### LINKED ISSUE #1114: Invalid code generation in type narrowing
In the llvm-ir generated for the following two test cases, there is a call for `_bal_tagged_to_float` that is getting called even if the type check shows the value is not a `float`. (I believe what is happening here is that the (implicit) else part that narrows the type to be float if the type check fails is getting merged into rest of the code)
1. [12-narrow/if21-v.bal](https://github.com/ballerina-platform/nballerina/blob/main/compiler/testSuite/12-narrow/if21-v.bal)
2. [14-bug/loopassign-v.bal](https://github.com/ballerina-platform/nballerina/blob/main/compiler/testSuite/14-bug/loopassign-v.bal)

Corresponding function calls
1. [12-narrow/if21-v.ll](https://gist.github.com/heshanpadmasiri/7f1d61f5a369659cefe01169e60e02e2#file-if21-v-ll-L57)
2. [14-bug/loopassign-v.ll](https://gist.github.com/heshanpadmasiri/7f1d61f5a369659cefe01169e60e02e2#file-loopassign-v-ll-L48)

It seems these function calls are getting optimized away under the default optimization setting (`-O2`) we are using but if we use (`-O0`) we will get a segfault.

[Link to the original comment that discusses this issue](https://github.com/ballerina-platform/nballerina/issues/1082#issuecomment-1306998115)

manuranga:
Fixes #1114
This uses solution 3 described in the issue.

### DIFF
--- a/compiler/modules/nback/basic.bal
+++ b/compiler/modules/nback/basic.bal
@@ -1,4 +1,5 @@
 import wso2/nballerina.bir;
+import wso2/nballerina.comm.err;
 import wso2/nballerina.types as t;
 import wso2/nballerina.print.llvm;
 
@@ -45,7 +46,10 @@ function buildPrologue(llvm:Builder builder, Scaffold scaffold, bir:Position pos
 function buildBasicBlock(llvm:Builder builder, Scaffold scaffold, bir:BasicBlock block) returns BuildError? {
     scaffold.setBasicBlock(block);
     builder.positionAtEnd(scaffold.basicBlock(block.label));
-    foreach var insn in block.insns {
+    bir:Insn[] insns = block.insns;
+    int typeMergeCount = check buildTypeMerges(builder, scaffold, block);
+    foreach int j in typeMergeCount ..< insns.length() {
+        bir:Insn insn = insns[j];
         scaffold.setCurrentPosition(builder, insn.pos);
         if insn is bir:IntArithmeticBinaryInsn {
             buildArithmeticBinary(builder, scaffold, insn);
@@ -116,9 +120,6 @@ function buildBasicBlock(llvm:Builder builder, Scaffold scaffold, bir:BasicBlock
         else if insn is bir:TypeBranchInsn {
             check buildTypeBranch(builder, scaffold, insn);
         }
-        else if insn is bir:TypeMergeInsn {
-            check buildTypeMerge(builder, scaffold, insn);
-        }
         else if insn is bir:AbnormalRetInsn {
             buildAbnormalRet(builder, scaffold, insn);
         }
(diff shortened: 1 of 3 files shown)


compiler/modules/nback/basic.bal
```diff
@@ -45,8 +45,22 @@ function buildPrologue(llvm:Builder builder, Scaffold scaffold, bir:Position pos
 function buildBasicBlock(llvm:Builder builder, Scaffold scaffold, bir:BasicBlock block) returns BuildError? {
     scaffold.setBasicBlock(block);
     builder.positionAtEnd(scaffold.basicBlock(block.label));
+    boolean narrowRegsInited = false;
```
jclark: It feels ugly to me to have this variable here and the rest of the related state in scaffold. Can we put narrowRegsInited into scaffold and then have a startBasicBlock function on scaffold?

compiler/modules/nback/basic.bal
```diff
@@ -45,8 +45,22 @@ function buildPrologue(llvm:Builder builder, Scaffold scaffold, bir:Position pos
...
+
     foreach var insn in block.insns {
         scaffold.setCurrentPosition(builder, insn.pos);
+        if insn is bir:TypeMergeInsn {
+            foreach var reg in insn.operands {
```
jclark: I would push the loop into the scaffold.

I want the logic for this in one place.

compiler/modules/nback/scaffold.bal
```diff
@@ -373,6 +375,27 @@ class Scaffold {
...
+        foreach var reg in self.blockUninitRegs[label] {
+            check buildNarrowReg(builder, self, reg);
+        }
+    }
+
+    function scheduleNarrowRegInit(bir:Label label, bir:NarrowRegister reg) {
+        self.blockUninitRegs[label].push(reg);
+    }
+
+    function unscheduleNarrowRegInit(bir:Label label, bir:Register reg) {
```
jclark: This is called in a loop. So we have a loop over a loop. Is this really the best data structure for this.

jclark (changes requested):
I feel we can make this patch cleaner.

compiler/modules/nback/basic.bal
manuranga: Renamed to `typeMergeFinished`, I think now it belongs in this function.

compiler/modules/nback/basic.bal
manuranga: Pushed it into `buildTypeMerge` instead.

compiler/modules/nback/scaffold.bal
```diff
@@ -373,6 +375,29 @@ class Scaffold {
...
+        var regs = self.blockUninitRegs[label];
+        foreach int i in 0..<regs.length() {
```
jclark: Space around `..<`

compiler/modules/nback/scaffold.bal
```diff
@@ -373,6 +375,29 @@ class Scaffold {
             return used;
         }
     }
+
+    // Should be called after all type merge insns.
+    // Will initialize narrow regs that are not initialized as part of type merges
+    function initNarrowRegs(llvm:Builder builder, bir:Label label) returns BuildError? {
```
jclark: I think we need `block` in the name of these functions.

compiler/modules/nback/basic.bal
```diff
@@ -45,8 +45,19 @@ function buildPrologue(llvm:Builder builder, Scaffold scaffold, bir:Position pos
 function buildBasicBlock(llvm:Builder builder, Scaffold scaffold, bir:BasicBlock block) returns BuildError? {
     scaffold.setBasicBlock(block);
     builder.positionAtEnd(scaffold.basicBlock(block.label));
+    boolean typeMergeFinished = false;
+
     foreach var insn in block.insns {
         scaffold.setCurrentPosition(builder, insn.pos);
+        if insn is bir:TypeMergeInsn {
+            check buildTypeMerge(builder, scaffold, block.label, insn);
+            continue;
```
jclark: This feels ugly to me.

compiler/modules/nback/basic.bal
```diff
@@ -45,7 +46,25 @@ function buildPrologue(llvm:Builder builder, Scaffold scaffold, bir:Position pos
 function buildBasicBlock(llvm:Builder builder, Scaffold scaffold, bir:BasicBlock block) returns BuildError? {
     scaffold.setBasicBlock(block);
     builder.positionAtEnd(scaffold.basicBlock(block.label));
-    foreach var insn in block.insns {
+    bir:Insn[] insns = block.insns;
+    int i = 0;
+
+    BlockNarrowRegBuilder narrowRegBuilder = scaffold.narrowRegBuilder(block.label);
```
jclark: Split into a function that returns number of TypeMergeInsn's.

jclark (approved):

PR #1124, opened 2022-11-22
Fix `xml<never>` to be a non-sequence

### LINKED ISSUE #1123: Fix `xml<never>` to be a non-sequence
`xml<never>` is currently represented as a sequence (just like `xml<Element>`) when it should just be a primitive.

manuranga:
Fixes  #1123


jclark:
Can we find a test case?

jclark:
Can we find a test case?

jclark (dismissed):

jclark (approved):

PR #1128, opened 2022-11-24
Fix optimizations not happening with object file generation

### LINKED ISSUE #1122: Look in to why certain optimizations not happening with object files
As noted in (2) [here](https://github.com/ballerina-platform/nballerina/pull/1112#issuecomment-1308288438) it seems that certain optimizations llvm / clang does with the normal llvm-ir (output of compiler version using `print.llvm`) are not happening with the object files (output of compiler version using `jni.llvm).

heshanpadmasiri:
## Purpose
Fix optimizations not happening with object file generation
Resolves #1122

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


heshanpadmasiri:
@jclark it turned out we do need to [use the `-lm` flag ](https://github.com/ballerina-platform/nballerina/pull/1128/files#diff-89a5dd5857c744afeb3204f39a51524aaa7ec4460293f2a090d3b9202870cf57R101) for both cases (with `print.llvm` and `jni.llvm`). It just happened to be the case with [06-float/22-v.bal](https://github.com/ballerina-platform/nballerina/blob/main/compiler/testSuite/06-float/22-v.bal) llvm can optimize away all the float modulo arithmetic operations at the compile time to constants. I added a new test case [14-bug/floatRem-v.bal](https://github.com/ballerina-platform/nballerina/blob/3ae1b6ab76915a18053fc13047a8876c6020fa95/compiler/testSuite/14-bug/floatRem-v.bal) where it wont do that.

test/sub.mk
```diff
@@ -98,9 +98,9 @@ clean:
 ifeq ($(COMPILER_OUTPUT),ll)
 .SECONDEXPANSION:
 $(exe_files): $$(patsubst %.exe,%.bc,$$@) $$(filter $$(patsubst %.exe,%,$$@).%.bc, $(mod_bc_files)) $(RT)
-	$(CLANG) $(CFLAGS) -g -o $@ $^ 
+	$(CLANG) $(CFLAGS) -lm -g -o $@ $^
```
jclark: `-lm` should be at the end. `-lm` isn't like a regular option: its order relative to `.o` files is significant.

jclark (changes requested):

test/sub.mk
jclark: Here and below in one place

jclark (approved):

PR #1131, opened 2022-11-30
Refactor function call via value

manuranga:
Refactor function call via value to multiple statements.

Inlined func apply no longer works on jBallerina 2201.3.0


manuranga:
It was due to this change:
https://github.com/ballerina-platform/ballerina-lang/issues/31586

jclark (approved):

PR #1139, opened 2022-12-13
Change jni.llvm to use TypeInContext

### LINKED ISSUE #1138: Change `jni.llvm` to use TypeInContext
In order to translate [our representations of llvm types](https://github.com/ballerina-platform/nballerina/blob/6710eabe3f50bdb5105dd319200893e3c110b1ec/compiler/modules/print.llvm/common.bal#L55) to `LLVMTypeRef` jni.llvm uses [`typeToLLVMType`](https://github.com/ballerina-platform/nballerina/blob/6710eabe3f50bdb5105dd319200893e3c110b1ec/llvm_jni/modules/jni.llvm/Type.bal#L3) function which optionally takes a `Context`. We need to make `Context` required so we can use `LLVM{TypeName}TypeInContext` over `LLVM{TypeName}Type` [here](https://github.com/ballerina-platform/nballerina/blob/6710eabe3f50bdb5105dd319200893e3c110b1ec/llvm_jni/modules/jni.llvm/Type.bal#L29). (This is required for #1116)

Apart from the following 3 cases, we can do this simply due to the fact types are used with other llvm constructs such as `Builder`, `Module`, etc. all of which have a reference to the `Context`
1.[ `constInt`](https://github.com/ballerina-platform/nballerina/blob/6710eabe3f50bdb5105dd319200893e3c110b1ec/compiler/modules/print.llvm/llvm.bal#L70)
2.[ `constFloat`](https://github.com/ballerina-platform/nballerina/blob/6710eabe3f50bdb5105dd319200893e3c110b1ec/compiler/modules/print.llvm/llvm.bal#L79)
3. [`constNull`](https://github.com/ballerina-platform/nballerina/blob/6710eabe3f50bdb5105dd319200893e3c110b1ec/compiler/modules/print.llvm/llvm.bal#L109)

With regard to refactoring the above functions, the fallowing ideas have been discussed.
1. In most cases we are using these functions (especially `constInt`) we are using it to create either `i64` (primarily for ballerina integer values), `i32` (indices since GEP instructions don't support `i64`), `i1` (represent booleans). Therefore create dedicated helper functions for these common cases such as `constInt64`
2. Have these helper functions take a `Builder` as an argument (in most cases we call the above 3 functions we don't have direct reference to `Context` but we have a reference to `Builder`), and the function will get the `Context` via `Builder`

One problem with 2 is in the [`init.bal`](https://github.com/ballerina-platform/nballerina/blob/main/compiler/modules/nback/init.bal) we don't use a `Builder` in most functions (but in all those cases we have access to `Context` via the `InitModuleContext` so one workaround would be for 2 to take `Builder|Context` instead of `Builder`) 

Another option for 2 is to take the `Context` itself instead of `Builder`. For this `Scaffold` will have something like `function llContext() returns llvm:Context => self.mod.llContext` (and `int.bal` will get the context via `InitModuleContext`)

heshanpadmasiri:
## Purpose
Change jni.llvm to use TypeInContext
Resolves #1138

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/nback/build.bal
+++ b/compiler/modules/nback/build.bal
@@ -124,15 +124,11 @@ final bir:ModuleId runtimeModule = {
     names: ["runtime"]
 };
 
-type Context object {
-    function llContext() returns llvm:Context;
-};
-
 function buildCheckPanicCode(llvm:Builder builder, Scaffold scaffold, llvm:Value valWithErr, bir:Position pos) returns llvm:Value {
     llvm:BasicBlock continueBlock = scaffold.addBasicBlock();
     llvm:BasicBlock errBlock = scaffold.addBasicBlock();
     llvm:Value panicCode = builder.extractValue(valWithErr, 1);
-    builder.condBr(builder.iCmp("ne", panicCode, buildConstInt(scaffold, 0)), errBlock, continueBlock);
+    builder.condBr(builder.iCmp("ne", panicCode, constInt(scaffold, 0)), errBlock, continueBlock);
     builder.positionAtEnd(errBlock);
     builder.store(buildErrorForPanic(builder, scaffold, panicCode, pos), scaffold.panicAddress());
     builder.br(scaffold.getOnPanic());
@@ -141,11 +137,11 @@ function buildCheckPanicCode(llvm:Builder builder, Scaffold scaffold, llvm:Value
 }
 
 function buildErrorForConstPanic(llvm:Builder builder, Scaffold scaffold, PanicIndex panicIndex, bir:Position pos) returns llvm:PointerValue {
-    return buildErrorForPackedPanic(builder, scaffold, buildConstInt(scaffold, panicIndex | (scaffold.lineNumber(pos) << 8)));
+    return buildErrorForPackedPanic(builder, scaffold, constInt(scaffold, panicIndex | (scaffold.lineNumber(pos) << 8)));
 }
 
 function buildErrorForPanic(llvm:Builder builder, Scaffold scaffold, llvm:Value panicIndex, bir:Position pos) returns llvm:PointerValue {
-    return buildErrorForPackedPanic(builder, scaffold, builder.iBitwise("or", panicIndex, buildConstInt(scaffold, scaffold.lineNumber(pos) << 8)));
+    return buildErrorForPackedPanic(builder, scaffold, builder.iBitwise("or", panicIndex, constInt(scaffold, scaffold.lineNumber(pos) << 8)));
 }
 
 function buildErrorForPackedPanic(llvm:Builder builder, Scaffold scaffold, llvm:Value packedPanic) returns llvm:PointerValue {
@@ -287,10 +283,10 @@ function buildConvertRepr(llvm:Builder builder, Scaffold scaffold, Repr sourceRe
 }
 
 function buildTaggedBoolean(llvm:Builder builder, Scaffold scaffold, llvm:Value value) returns llvm:Value {
-    return builder.getElementPtr(buildConstNilTaggedPtr(scaffold),
+    return builder.getElementPtr(constNilTaggedPtr(scaffold),
                                      [builder.iBitwise("or",
                                                        builder.zExt(value, LLVM_INT),
-                                                       buildConstInt(scaffold,TAG_BOOLEAN))]);
+                                                       constInt(scaffold,TAG_BOOLEAN))]);
 }
 
 function buildTaggedInt(llvm:Builder builder, Scaffold scaffold, llvm:Value value) returns llvm:PointerValue {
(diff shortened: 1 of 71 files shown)


compiler/modules/nback/build.bal
```diff
@@ -128,7 +128,7 @@ function buildCheckPanicCode(llvm:Builder builder, Scaffold scaffold, llvm:Value
     llvm:BasicBlock continueBlock = scaffold.addBasicBlock();
     llvm:BasicBlock errBlock = scaffold.addBasicBlock();
     llvm:Value panicCode = builder.extractValue(valWithErr, 1);
-    builder.condBr(builder.iCmp("ne", panicCode, llvm:constInt("i64", 0)), errBlock, continueBlock);
+    builder.condBr(builder.iCmp("ne", panicCode, constInt(scaffold, 0)), errBlock, continueBlock);
```
jclark: panicCode isn't a Ballerina int.

compiler/modules/nback/build.bal
```diff
@@ -275,47 +275,47 @@ function buildConvertRepr(llvm:Builder builder, Scaffold scaffold, Repr sourceRe
...
 }
 
-function buildTaggedBoolean(llvm:Builder builder, llvm:Value value) returns llvm:Value {
-    return builder.getElementPtr(llvm:constNull(LLVM_TAGGED_PTR),
+function buildTaggedBoolean(llvm:Builder builder, Scaffold scaffold, llvm:Value value) returns llvm:Value {
+    return builder.getElementPtr(constNilTaggedPtr(scaffold),
                                      [builder.iBitwise("or",
                                                        builder.zExt(value, LLVM_INT),
-                                                       llvm:constInt(LLVM_INT, TAG_BOOLEAN))]);
+                                                       constInt(scaffold,TAG_BOOLEAN))]);
```
jclark: Missing space

compiler/modules/nback/build.bal
```diff
@@ -349,18 +349,18 @@ function buildReprValue(llvm:Builder builder, Scaffold scaffold, bir:Operand ope
...
         }
         else if value is int {
             boolean alwaysInImmediateRange = IMMEDIATE_INT_MIN <= value && value <= IMMEDIATE_INT_MAX;
             IntRepr repr = { constraints: { min: value, max: value, all: true }, alwaysInImmediateRange };
-            return [repr, llvm:constInt(LLVM_INT, value)];
+            return [repr, constInt(scaffold, value)];
         }
         else if value is float {
-            return [REPR_FLOAT, llvm:constFloat(LLVM_DOUBLE, value)];
+            return [REPR_FLOAT, constDouble(scaffold, value)];
```
jclark: Isn't constFloat more consistent for this module?

compiler/modules/nback/init.bal
```diff
@@ -192,9 +197,9 @@ function createListDescInit(InitModuleContext cx, int tid, t:SemType semType) re
     t:ListAtomicType atomic = <t:ListAtomicType>t:listAtomicType(tc, semType);
     FunctionRef[] functionRefs = getListDescFunctionRefs(cx, atomic);
     llvm:Value[] initStructValues = [
-        llvm:constInt(LLVM_TID, tid),
-        llvm:constInt("i32", atomic.members.initial.length()),
-        llvm:constInt("i64", atomic.members.fixedLength)
+        constTid(cx, tid),
+        constIndex(cx, atomic.members.initial.length()),
```
jclark: I don't think changing from "i32" to Index or "i64" to Int is an improvement.  These need to match the definitions in balrt.h which use int32 and int64.

jclark (changes requested):
I don't want to have to look at every single use of llvm:constInt.  Can you not change things (leave  "i32" as constI32 etc) in this PR? And then separately do a PR to change any that you feel are necessary?

compiler/modules/nback/build.bal
heshanpadmasiri: My idea for the naming scheme was `constFoo` would build something of type `LLVM_FOO` (ex: `constInt` -> `LLVM_INT`). In retrospect, I feel using `LLVM_FLOAT` may be a better choice than `LLVM_DOUBLE`, so it will be similar to `LLVM_INT`, `LLVM_BOOLEAN` etc. (the name is derived from what ballerina type they represent over what llvm type they are using). May I change `LLVM_DOUBLE` to `LLVM_FLOAT` as well or just change this to `constFloat`?

jclark: The llvm module is naming things from an LLVM perspective. The nback module is naming things from the Ballerina perspective.  So the llvm module should use DOUBLE (because that is what LLVM calls it), but the nback module should call it FLOAT (because that's what Ballerina calls).  So just change this module.

heshanpadmasiri: Fixed by changing `LLVM_DOUBLE` to `LLVM_FLOAT` (`LLVM_DOUBLE` was a type alias defined in [`scaffold.bal`](https://github.com/ballerina-platform/nballerina/blob/016c743511e152dc889e9e6b11c3481c742fbed3/compiler/modules/nback/scaffold.bal#L12) for `double`) and `constDouble` to `constFloat`

compiler/modules/nback/build.bal
heshanpadmasiri: fixed

compiler/modules/nback/build.bal
heshanpadmasiri: fixed

compiler/modules/nback/init.bal
heshanpadmasiri: fixed

heshanpadmasiri:
> I don't want to have to look at every single use of llvm:constInt. Can you not change things (leave "i32" as constI32 etc) in this PR? And then separately do a PR to change any that you feel are necessary?

fixed. (Will send a separate PR)

jclark (approved):

PR #1147, opened 2023-01-04
Fix result type of shift expressions

### LINKED ISSUE #1145: Result type of shift operators is incorrect
Currently, the result type for shift expression is the union of type of the two operands (according to code [here](https://github.com/ballerina-platform/nballerina/blob/016c743511e152dc889e9e6b11c3481c742fbed3/compiler/modules/front/expr.bal#L839)). However according to [spec](https://ballerina.io/spec/lang/master/#section_6.25) "If the operator is >> or >>> and the left hand operand is a subtype of int:UnsignedK when K is 8, 16 or 32, then the static type of the result is int:UnsignedN where N is the smallest such K; otherwise, the static type of the result is int. In all cases the static type of the result is modified by the usual singleton typing rules."

This leads to invalid results in cases such as,
```
import ballerina/io;
public function main() {
    int:Unsigned8 x = 250;
    int:Signed8 a = 2;
   
    int:Unsigned8 x1 = x >> a; // error: value is not a subtype of the LHS
}
```
```
import ballerina/io;
public function main() {
    int:Unsigned8 a = 255;
    int:Unsigned8 b = 63;
    io:println(a << b); // should be -9223372036854775808 but gets 0
}
```

heshanpadmasiri:
## Purpose
Fix result type of shift expressions
Resolves #1145

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/front/expr.bal
+++ b/compiler/modules/front/expr.bal
@@ -836,7 +836,13 @@ function codeGenBitwiseBinaryExpr(ExprContext cx, bir:BasicBlock bb, s:BinaryBit
     ValueFlags resultFlags = leftFlags & rightFlags;
     t:SemType lt = t:widenUnsigned(lhs.semType);
     t:SemType rt = t:widenUnsigned(rhs.semType);
-    t:SemType resultType = op == "&" ? t:intersect(lt, rt) : t:union(lt, rt);
+    t:SemType resultType;
+    if op is "<<"|">>"|">>>" {
+        resultType = op is ">>"|">>>" && t:isSubtype(cx.mod.tc, lt, t:intWidthUnsigned(32)) ? lt : t:INT;
+    }
+    else {
+        resultType = op == "&" ? t:intersect(lt, rt) : t:union(lt, rt);
+    }
     if resultFlags != 0 {
         int value = bitwiseEval(op, leftValue, rightValue);
         if (resultFlags & VALUE_SINGLE_SHAPE) != 0 {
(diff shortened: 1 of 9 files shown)


compiler/modules/front/expr.bal
```diff
@@ -836,7 +836,13 @@ function codeGenBitwiseBinaryExpr(ExprContext cx, bir:BasicBlock bb, s:BinaryBit
     ValueFlags resultFlags = leftFlags & rightFlags;
     t:SemType lt = t:widenUnsigned(lhs.semType);
     t:SemType rt = t:widenUnsigned(rhs.semType);
-    t:SemType resultType = op == "&" ? t:intersect(lt, rt) : t:union(lt, rt);
+    t:SemType resultType;
+    if op is "<<"|">>"|">>>" {
+        resultType = op is ">>"|">>>" && t:isSubtype(cx.mod.tc, lt, t:intWidthUnsigned(32)) ? lt : t:INT;
```
jclark, line 841: Think about case where lt is a proper subtype of unisgnedN.

compiler/modules/front/expr.bal
```diff
@@ -836,7 +836,13 @@ function codeGenBitwiseBinaryExpr(ExprContext cx, bir:BasicBlock bb, s:BinaryBit
     ValueFlags resultFlags = leftFlags & rightFlags;
     t:SemType lt = t:widenUnsigned(lhs.semType);
     t:SemType rt = t:widenUnsigned(rhs.semType);
-    t:SemType resultType = op == "&" ? t:intersect(lt, rt) : t:union(lt, rt);
+    t:SemType resultType;
+    if op is "<<"|">>"|">>>" {
```
jclark: Can we use a define for this?

heshanpadmasiri: fixed (added `BitwiseShiftOp` to AST)

compiler/modules/front/expr.bal
heshanpadmasiri: This is correct since `lt` is the result of [`widenUnsigned`](https://github.com/ballerina-platform/nballerina/blob/2d43f999afe0c44e4be2b55b575b15537c8ac99b/compiler/modules/types/core.bal#L978) (added a test case to cover this)

jclark (approved):

PR #1149, opened 2023-01-05
Fix bug in xml is empty

### LINKED ISSUE #1148: Error when defining xml sequence type of length 2 or more
```ballerina
xml<xml:Element> & !xml:Element & !xml<never>
```

Defining above types gives the following error.

`intersection must not be empty`

manuranga:
Fixes #1148

### DIFF
--- a/compiler/modules/types/xml.bal
+++ b/compiler/modules/types/xml.bal
@@ -126,7 +126,10 @@ function xmlFormulaIsEmpty(Context cx, Conjunction? pos, Conjunction? neg) retur
 }
 
 function xmlCollectAllPrimitives(Conjunction? con) returns int {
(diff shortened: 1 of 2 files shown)


compiler/modules/types/xml.bal
```diff
@@ -126,13 +126,14 @@ function xmlFormulaIsEmpty(Context cx, Conjunction? pos, Conjunction? neg) retur
 }
 
 function xmlCollectAllPrimitives(Conjunction? con) returns int {
-    int bits = 0;
+    int? bits = ();
```
jclark: I think we can make this cleaner

jclark (approved):

PR #1161, opened 2023-01-30
Avoid maintaining the order of pre-defined atoms in two places

### LINKED ISSUE #1130: Avoid maintaining the order of pre-defined atoms in two places
**Description:**
$subject.

We need to refactor as follows.
- Use constant for the indexes.
- Define an array of atoms and loop over it to initialize. e.g. `final readonly & Atom[] PREDEFINED_ATOMS = [ ...];`
- Can be accessed as follows: `PREDEFINED_ATOMS[CELL_ATOM_TOP_INDEX];`

Related code segment:
https://github.com/ballerina-platform/nballerina/blob/941f899f9358705332a1f1e665775dde84403ca4/compiler/modules/types/core.bal#L1286-L1293

lochana-chathura:
## Purpose
$subject.

Depends on PR #1160
Fixes #1130


anupama-pathirage:
Closing this PR due to inactivity. Please re-open if this is still needed.

PR #1162, opened 2023-02-02
Extend undef for list type

### LINKED ISSUE #1158: Extend undef for list type
**Description:**
$subject.

lochana-chathura:
## Purpose
$subject.

Fixes #1158


compiler/modules/types/list.bal
```diff
@@ -19,6 +19,10 @@ public type FixedLengthArray record {|
 // The SemTypes in this list are not `never`.
 public type ListMemberTypes [Range[], SemType[]];
 
+public function listAtomicTypeMemberAtInnerVal(ListAtomicType atomic, int i) returns SemType {
```
heshanpadmasiri, line 22: This seems identical to `listAtomicTypeMemberAtInner` (unlike [`mappingAtomicTypeMemberAtInnerVal`](https://github.com/ballerina-platform/nballerina/blob/49cfa7b757a21ff94f0d949925f36d27bfcfc88e/compiler/modules/types/mapping.bal#L18). Is there a reason why we don't need to remove `UNDEF` here, for an arbitrary `ListAtomicType`?

lochana-chathura: Fixed in [83989d4](https://github.com/ballerina-platform/nballerina/pull/1162/commits/83989d418030e0b9f889aafda7a8eba3282ff571)

lochana-chathura:
@jclark PR is ready to review.

compiler/modules/types/list.bal
```diff
@@ -185,8 +188,8 @@ function listFormulaIsEmpty(Context cx, Conjunction? pos, Conjunction? neg) retu
             return true;
         }
         // Ensure that we can use isNever on rest in listInhabited
-        if cellInner(rest) != NEVER && isEmpty(cx, rest) {
-            rest = cellContaining(cx.env, NEVER);
+        if cellInnerVal(rest) != NEVER && isEmpty(cx, rest) {
```
heshanpadmasiri: Given that the rest type of a list (at least in cases where it goes via `defineListTypeWrapped`) is a union of "rest type" and UNDEF (whereas previously it was just the "rest type") I am not sure that their intersections can have a rest type that is empty (It must at least have UNDEF). Either way, I think we can get rid of this whole if block since `listInhabited` is no longer dependent on any member being NEVER on the positive side (I think we got rid of that some time ago but never removed this part here).

heshanpadmasiri:
Other changes look okay to me

heshanpadmasiri (dismissed):

compiler/modules/types/list.bal
lochana-chathura: +1 it never gets empty with undef.

jclark (approved):

PR #1166, opened 2023-02-13
Remove restrictions on where list/mapping constructors are allowed

### LINKED ISSUE #1003: Remove restrictions on where list/mapping constructors are allowed
The subset restricts the syntactic context where list and mapping constructors are allowed, to ensure that there is always a contextually expected type.

Now that we have tuple types and record types, we can handle this case by inferring the inherent type as the spec requires.

### LINKED ISSUE #1167: Fix list comparison when the is no contextually expected type
With #1166 we should be able to run the following test case
```
import ballerina/io;
public function main() {
    io:println([1,2] < [1,2]); // @output false
}
```
However, we are getting an error saying
```
test.bal:4:22: error: operands of '<' do not belong to an ordered type
    io:println([1,2] < [1,2]); // @output true
                     ^
```
I believe this is caused by `codeGenRelationalExpr` passing its expected type to each of the operands ([here](https://github.com/ballerina-platform/nballerina/blob/49cfa7b757a21ff94f0d949925f36d27bfcfc88e/compiler/modules/front/expr.bal#L1067))

In this particular test case arguments to the `println` function have expected type of `any` which when passed to the `codeGenListConstructor` causes it to select the type of `[1,2]` to be `any[]` (instead of `[1,2]` according to spec). To fix it we need #1166 (so we can parse this test case) but I think issue is separate from #1003 (since it is caused by Relational Expr)

heshanpadmasiri:
## Purpose
Remove restrictions on where list/mapping constructors are allowed
Resolves #1003, #1167

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/compiler/modules/front.syntax/parseExpr.bal
+++ b/compiler/modules/front.syntax/parseExpr.bal
@@ -3,26 +3,6 @@ import wso2/nballerina.comm.err;
 type SpecialMethodName "map";
 
 function parseExpr(Tokenizer tok) returns Expr|err:Syntax {
-    Token? t = tok.current();
-    Position startPos = tok.currentStartPos();
-    if t == "[" {
-        check tok.advance();
-        var [members, _] = check parseExprList(tok, "]");
-        Position endPos = tok.previousEndPos();
-        ListConstructorExpr expr = { startPos, endPos, opPos: startPos, members };
-        return expr;
-    }
-    else if t == "{" {
-        check tok.advance();
-        Field[] fields = check parseFields(tok);
-        Position endPos = tok.previousEndPos();
-        MappingConstructorExpr expr = { startPos, endPos, opPos: startPos, fields };
-        return expr;
-    }
-    return parseInnerExpr(tok);
-}
-
-function parseInnerExpr(Tokenizer tok) returns Expr|err:Syntax {
     return parseLogicalOrExpr(tok);
 }
 
@@ -339,7 +319,7 @@ function startPrimaryExpr(Tokenizer tok) returns Expr|err:Syntax {
             LiteralExpr expr = { startPos, endPos, value: () };
             return expr;
         }
-        Expr innerExpr = check parseInnerExpr(tok);
+        Expr innerExpr = check parseExpr(tok);
         endPos = check tok.expectEnd(")");
         return { startPos, endPos, innerExpr };
     }
@@ -361,6 +341,20 @@ function startPrimaryExpr(Tokenizer tok) returns Expr|err:Syntax {
         endPos = check tok.expectEnd(")");
         return { startPos, endPos, message, kwPos };
     }
+    else if t == "[" {
+        check tok.advance();
+        var [members, _] = check parseExprList(tok, "]");
+        endPos = tok.previousEndPos();
+        ListConstructorExpr expr = { startPos, endPos, opPos: startPos, members };
+        return expr;
+    }
+    else if t == "{" {
+        check tok.advance();
+        Field[] fields = check parseFields(tok);
+        endPos = tok.previousEndPos();
+        MappingConstructorExpr expr = { startPos, endPos, opPos: startPos, fields };
+        return expr;
+    }
     else {
         return parseError(tok);
     }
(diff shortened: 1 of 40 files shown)


compiler/modules/front.syntax/parseExpr.bal
```diff
@@ -339,7 +319,7 @@ function startPrimaryExpr(Tokenizer tok) returns Expr|err:Syntax {
             LiteralExpr expr = { startPos, endPos, value: () };
             return expr;
         }
-        Expr innerExpr = check parseInnerExpr(tok);
+        Expr innerExpr = check parseExpr(tok);
```
jclark: Using `innerExpr` no longer seems right

compiler/testSuite/14-list/7-v.bal
```diff
@@ -0,0 +1,42 @@
...
+
+// function call arguments get their expected type from the function type
+function foo(any a) {
+    if a is [2, 3] {
+        io:println("unexpected");
+    }
+    if a is [int, int] {
+        io:println("unexpected");
+    }
+    else {
```
jclark, line 21: int[] is also unexpected

compiler/testSuite/14-list/7-v.bal
```diff
@@ -0,0 +1,42 @@
...
+        io:println("working");
```
jclark: io:println(a is any)

compiler/testSuite/14-list/11-v.bal
```diff
@@ -0,0 +1,13 @@
+import ballerina/io;
+public function main() {
+    match [1, 2] {
+        _ => { io:println("working");} // @output working
```
lochana-chathura:
```suggestion
        _ => { io:println("working"); } // @output working
```

compiler/testSuite/14-mapping/1-v.bal
```diff
@@ -0,0 +1,14 @@
+import ballerina/io;
+
+public function main() {
+    io:println({a: 1, b: 2} == {a: 1, b: 2}); // @output true
```
lochana-chathura:
```suggestion
    io:println({ a: 1, b: 2 } == { a: 1, b: 2 }); // @output true
```

lochana-chathura (commented):
There seem to be several formatting issues in the test cases

compiler/testSuite/14-list/7-v.bal
heshanpadmasiri: fixed

compiler/testSuite/14-list/11-v.bal
heshanpadmasiri: fixed

compiler/testSuite/14-mapping/1-v.bal
heshanpadmasiri: fixed

compiler/testSuite/14-list/7-v.bal
heshanpadmasiri: fixed

compiler/testSuite/14-list/7-v.bal
heshanpadmasiri: fixed

compiler/modules/front.syntax/parseExpr.bal
heshanpadmasiri: fixed (Changed `CallStmt` to avoid type conflicts. I felt `GroupingExpr` having the field named `expr` makes more sense than having it in `CallStmt`)

docs/subset14.md
```diff
@@ -396,13 +398,14 @@ Two kinds of `import` are supported.
 
 ## Notes
 
-* The syntax restricts where a `list-constructor-expr` or `mapping-constructor-expr` can occur so as to avoid the need to infer a type for the constructed list.
+* `list-constructor-expr` and `mapping-constructor-expr` are not allowed as `const-expr`.
```
jclark: within a `const-expr`

docs/subset14.md
```diff
@@ -234,13 +234,13 @@ simple-const-expr =
    | "-" floating-point-literal
    | const-reference-expr
 
-expression = inner-expr | list-constructor-expr | mapping-constructor-expr
+expression = logical-or-expr
 
-const-expr = inner-expr # must also satisfy restrictions of const-expr as in Ballerina language spec
+const-expr = logical-or-expr # must also satisfy restrictions of const-expr as in Ballerina language spec and should be neither list-constructor-expr nor mapping-constructor-expr
```
jclark: Don't need this, given that it is mentioned below.

docs/subset14.md
heshanpadmasiri: fixed

docs/subset14.md
heshanpadmasiri: fixed

jclark (approved):

PR #1177, opened 2023-03-02
Fix empty list normalization

### LINKED ISSUE #1176: Empty list type has multiple internal representation
List atomic type for empty list contain the the member type. While it behaves correctly with respect to `listInhabited`, it leads to sub-optimal behavior in places where types are compared by value (eg: de-duplication of inherent types in `init.bal`).

manuranga:
Fixes #1176


compiler/modules/types/list.bal
```diff
@@ -109,6 +109,9 @@ public class ListDefinition {
 }
 
 function fixedLengthNormalize(FixedLengthArray array) returns FixedLengthArray {
+    if array.fixedLength == 0 {
```
jclark, line 112: Add a comment above FixedLengthArray

jclark (approved):

PR #1183, opened 2023-03-13
Refactor `diagnostic:File` to support non-source-aware implementations

manuranga:
Since current `File` API can't be supported on top of bir sexpr, it's broken in to two APIs, `File` and `SourceFile`.
Part of #1031

### DIFF
--- a/compiler/modules/comm.diagnostic/diagnostic.bal
+++ b/compiler/modules/comm.diagnostic/diagnostic.bal
@@ -26,8 +26,9 @@ public type File readonly & object {
     public function filename() returns string;
     public function directory() returns string?;
     public function lineColumn(Position pos) returns LineColumn;
-    public function lineContent(Position|Range range) returns [string, string, string];
-    public function qNameRange(Position startPos) returns Range;
+    // Returns source splitted at the given range into prefix, content, and suffix.
+    // Returns nil if underlying file does not support textual source.
+    public function lineContent(Position|Range range) returns [string, string, string]?;
 };
 
 public type Location readonly & record {|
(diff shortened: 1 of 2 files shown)


manuranga:
@jclark 

jBallerina object subtypeing is working correctly, I may have missed a readonly previously.

Still I can't define a File subtype in bir module. I need them both in diagnostics module. Because, diagnostics are crated by both front and back, it needs to be aware of both kinds of Files. If bir were to define a super type, this will not be an issue, but we need a sub type.

compiler/modules/comm.diagnostic/diagnostic.bal
```diff
@@ -126,17 +132,19 @@ public function format(Diagnostic d) returns string[] {
     line += ": error: " + d.message;
     lines.push(line);
     Range|Position range = loc.range;
-    string[] content = loc.file.lineContent(range);
-    if content[1] == "" {
-        panic error(string `error range ${lc.toString()} is empty in ${loc.file.filename()}`);
+    File file = loc.file;
+    if file is SourceFile {
```
manuranga: This doesn't feel right, since type of the `loc.file` is `File`, not `File|SourceFile`. But If I do that, jBallerina doesn't allow me to call the methods.

compiler/modules/comm.diagnostic/diagnostic.bal
```diff
@@ -26,8 +26,9 @@ public type File readonly & object {
     public function filename() returns string;
     public function directory() returns string?;
     public function lineColumn(Position pos) returns LineColumn;
-    public function lineContent(Position|Range range) returns [string, string, string];
-    public function qNameRange(Position startPos) returns Range;
+    // Returns source splitted at the given range into prefix, content, and suffix.
+    // Returns nil if underling file does not support textual source.
```
jclark: underlying

jclark:
OK apart from typo

jclark (approved):

PR #1218, opened 2023-07-18
Add support for object type

### LINKED ISSUE #1217: Add support for object type
At this stage, I am planning to add support to [object type descriptors](https://ballerina.io/spec/lang/2022R1/#section_5.5.3) with the following limitations

1. We will not support object type values (only type)
2. We will not support visibility modifiers
3. We will not support importing object types from other modules
4. We will not support remote method declarations
5. We will not support object type qualifiers (network qualifiers and isolated)

Therefore we will support the following addition to the grammar
```
object_type_descriptor = "object" "{" object_member_descriptor* "}"
object_member_descriptor = object_field_descriptor | method_decl
object_field_descriptor = type_descriptor field_name;
method_decl = "function" method_name function_signature ";"
method_name = identifier|special_method_name
special_method_name = "map" (currently we don't tree "start" and "join" as keywords)
```
# SemType
As far as semtype representation of an object type goes I think we can represent them as an open record type,
``` ts
object O {
    int foo;
    string bar;
    function baz(int, int) returns int;
}

record R {
    int foo;
    string bar;
    // NOTE: function signature is different since we need to pass `self` to methods
    // also note this is going to be insufficient to separate remote methods when we support them
    (function (R, int, int) returns int) baz;
}
```
_To deal with the method kind in the future (maybe we start with it already even if there is no difference), we can represent them as a separate list of (method kind, method type) pairs and limit the record type to only fields_

In addition, type representation of the object type must contain information such as,
1. Isolated bit
2. Network interaction kind
3. Field kind (for methods we need to encoded if they are remote function or not)
4. Read-only bit

While 4 is handled by cell types I think Atomic type will need fields for others (at this stage we support only 1 value for each, so we can ignore them as well)

With these in mind, I think we can have an Object atomic type as,

```ts
type MethodKind "remote method"|"method"
type Method record {|
   string name;
   MethodKind kind;
   Semtype type; // I don't think we need it here if we add the method types to the record type itself
|}

type ObjectAtomicType record {|
   SemType memberTypes; // This is the record type
   Method[] methods; // For the time being we can ignore this as well since there one `MethodKind`
   // In the future
   "isolated"? 'isolated;
   "client"|"server"|() network;
|}
```

Then for subtyping, we will check the subtyping of memberTypes record type and go over each method and see if their method kinds match (Not entirely sure how to do the latter)

# AST changes
I am planning to add the following AST nodes
```
public type ObjectTypeDesc record {|
    *PositionFields;
    MemberDesc[] member;
    t:ObjectDefinition? defn = (); // I think we need to support recursive types similar to mapping
|}

public type MemberDesc record {|
    *PositionFields;
    string name;
    TypeDesc typeDesc;
    "value"|"method"|"remote method" kind;
|};
```

heshanpadmasiri:
## Purpose
Add support for object type
Resolves #1217 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


compiler/modules/types/object.bal
```diff
@@ -0,0 +1,64 @@
...
+}
+
+function restMemberType(Env env) returns CellSemType {
+    SemType fieldMemberType = defineMappingTypeWrapped(new, env, [{ name: "value", ty: ANY },
+                                                                  { name: "kind", ty: MEMBER_KIND_FIELD }],
+                                                       NEVER);
+    SemType methodMemberType = defineMappingTypeWrapped(new, env, [{ name: "value", ty: FUNCTION, ro: true },
+                                                                   { name: "kind", ty: MEMBER_KIND_METHOD }],
+                                                        NEVER);
+    return cellContaining(env, union(fieldMemberType, methodMemberType));
```
heshanpadmasiri, line 44: @jclark It turned out that every parameter in the function type is also represented as individual cells ([ref](https://github.com/ballerina-platform/nballerina/blob/0331ffaf618ed6e5ece4efa405e8712105cdeabc/compiler/modules/front/resolveTypes.bal#L336C40-L336C40)) and list type also expect each member to be a cell. Should I still change the mapping type to such that members are not cells?

heshanpadmasiri: Draft PR with changes that will be needed #1219

compiler/modules/types/object.bal
```diff
@@ -0,0 +1,68 @@
...
+
+    public function getSemType(Env env) returns SemType {
+        return objectContaining(self.mappingDefn.getSemType(env));
+    }
+
+    public function define(Env env, Member[] members) returns SemType {
+        if members.length() == 0 {
+            return OBJECT;
+        }
+        CellField[] fieldCells = from Member member in members where member.kind == "field" select fieldMember(env, member);
```
jclark: Combine into one loop

compiler/modules/types/object.bal
```diff
@@ -0,0 +1,68 @@
...
+                           defineMappingTypeWrapped(new, env, [{ name: "value", ty: member.valueTy, ro: true },
+                                                               { name: "kind", ty: MEMBER_KIND_METHOD }],
+                                                    NEVER))];
+}
+
+final BasicTypeOps objectOps = {
+    union: bddSubtypeUnion,
+    intersect: bddSubtypeIntersect,
+    diff: bddSubtypeDiff,
+    complement: bddSubtypeComplement,
```
jclark: Need to do something similar to tables here.

docs/subset15.md
```diff
@@ -162,6 +164,13 @@ param-list = param ["," param]* ["," rest-param]
 param = type-desc [identifier] # identifier can only be omitted when occurring in function-type-desc
 rest-param = type-desc "..." identifier
 
+object-type-desc = "object" "{" object-member-desc* "}"
+object-member-desc = object-field-desc | method-decl
+object-field-desc = "public" type-desc identifier ";"
+method-decl = "public" "function" method-name "(" [param-list] ")" ["returns" type-desc] ";"
```
jclark: Use signature here

jclark: And in function-type-desc above.

heshanpadmasiri: Fixed

compiler/modules/types/object.bal
heshanpadmasiri: Fixed

compiler/modules/types/object.bal
heshanpadmasiri: Fixed. However, I have the same [concern as the function type](https://github.com/ballerina-platform/nballerina/pull/1221#discussion_r1274418942).

jclark (approved):

PR #1219, opened 2023-07-24
Allow non-cell mapping members

heshanpadmasiri:
## Purpose
Remove the restriction requiring mapping members to be CellSemTypes.


heshanpadmasiri:
Decided not to implement this since functions also use cells for each parameter type.

PR #1221, opened 2023-07-25
Make compliment op for function types more explicit

### LINKED ISSUE #1220: Make complement op for function type explicit
Currently, we are using `bddSubtypeComplement` ([ref](https://github.com/ballerina-platform/nballerina/blob/0331ffaf618ed6e5ece4efa405e8712105cdeabc/compiler/modules/types/function.bal#L228)) as the complement operation for function types.  Furthermore, the function type represents its parameters as a read-only tuple. ~Therefore when we take the complement for a function type we calculate the parameter type by removing the parameter type from all list types. This is wrong and we must instead use the set of all read-only tuples as the top type for parameters.~

Edit: above statement is wrong since function parameters are contravariant. Therefore "top type" for parameters is NEVER. Therefore current implementation is technically correct. But it may still be worth it to make the function top type explicit.

heshanpadmasiri:
## Purpose
$Subject
Resolves #1220

### DIFF
--- a/compiler/modules/types/core.bal
+++ b/compiler/modules/types/core.bal
@@ -1357,6 +1357,7 @@ final CellAtomicType CELL_ATOMIC_INNER_RO = { ty: INNER_READONLY, mut: CELL_MUT_
 final MappingAtomicType MAPPING_ATOMIC_INNER = { names: [], types: [], rest: CELL_SEMTYPE_INNER };
 final ListAtomicType LIST_ATOMIC_INNER = { members: { initial: [], fixedLength: 0 }, rest: CELL_SEMTYPE_INNER };
 final ListAtomicType LIST_ATOMIC_MAPPING = { members: {initial: [], fixedLength: 0 }, rest: CELL_SEMTYPE_INNER_MAPPING };
+final FunctionAtomicType FUNCTION_ATOMIC_TOP = [NEVER, ANY];
 
 final TypeAtom ATOM_CELL_VAL = { index: 0, atomicType: CELL_ATOMIC_VAL };
 final TypeAtom ATOM_CELL_NEVER = { index: 1, atomicType: CELL_ATOMIC_NEVER };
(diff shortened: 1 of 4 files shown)


compiler/modules/types/core.bal
```diff
@@ -1357,6 +1357,7 @@ final CellAtomicType CELL_ATOMIC_INNER_RO = { ty: INNER_READONLY, mut: CELL_MUT_
 final MappingAtomicType MAPPING_ATOMIC_INNER = { names: [], types: [], rest: CELL_SEMTYPE_INNER };
 final ListAtomicType LIST_ATOMIC_INNER = { members: { initial: [], fixedLength: 0 }, rest: CELL_SEMTYPE_INNER };
 final ListAtomicType LIST_ATOMIC_MAPPING = { members: {initial: [], fixedLength: 0 }, rest: CELL_SEMTYPE_INNER_MAPPING };
+final FunctionAtomicType FUNCTION_ATOMIC_TOP = [NEVER, ANY];
```
heshanpadmasiri, line 1360: @jclark I don't think our original implementation was wrong for function types since the top type for the parameter part is not a list type. This is because parameters are contravariant, thus top type must be a subtype of all parameter list types, which is `NEVER`. But it might still be worth it to make it explicit to avoid any confusion in the future.

compiler/modules/types/function.bal
```diff
@@ -221,6 +221,10 @@ function bddFunctionAtomicType(Env env, Bdd bdd) returns FunctionAtomicType? {
     return ();
 }
 
+function functionSubtypeComplement(ProperSubtypeData t) returns SubtypeData {
+    return bddSubtypeDiff(FUNCTION_TOP, t);
```
heshanpadmasiri: @jclark I don't think this actually makes a difference since I can replace the `FUNCTION_TOP` with any arbitrary BDD node (say `LIST_SUBTYPE_MAPPING_RO` -> `readonly & (map<readonly>)[]`) and everything works. This is true for table type as well. The only test that fails is [this serialization test](https://github.com/ballerina-platform/nballerina/blob/0331ffaf618ed6e5ece4efa405e8712105cdeabc/compiler/modules/types.sexpr/README.md?plain=1#L143) since they produce slightly different BDDs.

compiler/modules/types/core.bal
jclark: I see your point. I suggest you comment FunctionAtomicType accordingly.

PR #1223, opened 2023-07-31
Implement anonymous functions w/o variable capture

### LINKED ISSUE #908: Implement anonymous functions w/o variable capture
This builds on #907 to add support for anonymous functions.

This means supporting [explicit-anonymous-function-expr](https://ballerina.io/spec/lang/2022R1/#section_6.18) with a block-function-body.

Initially, we would have the restriction that these functions could not refer to block-scope variables defined outside the function body. This would be handled in #909.

Further we would _not_ support
- `=>` syntax (i.e. expr-function-body)
- inferring types for parameters/return type (infer-anonymous-function-expr)

heshanpadmasiri:
## Purpose
Implement anonymous functions w/o variable capture
Resolves #908


compiler/modules/bir/bir.bal
```diff
@@ -42,17 +42,30 @@ public type Label int;
...
 
+public type Function FunctionDefn | AnonFunction;
+
+public type FunctionBase record {|
+    FunctionDecl decl;
+    # Index of source part in which the definition occurs
+    int partIndex;
+    # The position of the definition
+    Position position;
+    readonly...;
```
jclark, line 53: Why `readonly...` rather than just an open record?

compiler/modules/bir/bir.bal
```diff
@@ -42,17 +42,30 @@ public type Label int;
...
+|};
 
-    # Index of source part in which the definition occurs
-    int partIndex;
-    # The position of the definition
-    Position position;
+public type AnonFunction readonly & record {|
+    *FunctionBase;
+    # The index of the function in the module
+    int index;
```
jclark: Why do we need this in AnonFunction but not in FunctionDefn?

jclark: For references I guess, but can't we just refer to it directly?

compiler/modules/bir/bir.bal
```diff
@@ -83,6 +104,8 @@ public type FunctionCode record {|
     Register[] registers = [];
     # Single-entry single-exit regions
     Region[] regions = [];
+    # TODO:index in bir:Module functions 
+    [int, FunctionCode][] childAnnonFunctions = [];
```
jclark: Spelling: `Annon`

jclark: I think we can get rid of this.

heshanpadmasiri: Fixed. Decided to get rid of the `childAnonFunctions` fields completely since nothing is using it

compiler/modules/bir/bir.bal
heshanpadmasiri: Moved `index` into the `*FunctionBase`

compiler/modules/bir/bir.bal
```diff
@@ -62,8 +73,16 @@ public type InternalSymbol readonly & record {|
 
 public type Symbol InternalSymbol|ExternalSymbol;
 
-public type FunctionRef readonly & record {|
-    Symbol symbol;
+public type FunctionRef ExternalFunctionRef | InternalFunctionRef;
```
heshanpadmasiri, line 76: @jclark One problem with completely factoring `FunctionRef` to be the same between Anonymous function and function definitions is we can't give an index to langlib functions [here](https://github.com/ballerina-platform/nballerina/blob/84ccc96b98b72e86b4cfef68400cdcb1439cb49e/compiler/modules/front/expr.bal#L1631), since we don't really have an actual module for langlib modules. Also, I am not sure if is a good idea to use an index for external functions (we need at least two indices one for the module and another as the index withing the module). As such I decided to factor FunctionRef into `ExternalFunctionRef` (which will use a symbol) and `InternalFunctionRef` (common for both anonymous functions and normal ones)

compiler/modules/bir/bir.bal
```diff
@@ -381,11 +401,25 @@ public type StringConstOperand readonly & record {|
     string value;
 |};
 
-public type FunctionConstOperand readonly & record {|
+public type FunctionConstOperand InternalFunctionConstOperand | ExternalFunctionConstOperand;
```
jclark: Can simplify this

heshanpadmasiri: Fixed

heshanpadmasiri:
[Serialization examples ](https://heshanpadmasiri.github.io/BIR-report/).
 _(For capture see `capture-v.bal` example)_

compiler/modules/bir/bir.bal
```diff
@@ -663,6 +691,14 @@ public type CallIndirectInsn readonly & record {|
     boolean restParamIsList;
 |};
 
+# Create a closure by enclosing the first operand with the remaining operands.
+public type CaptureInsn readonly & record {|
+    *ResultInsnBase;
+    INSN_CAPTURE name = INSN_CAPTURE;
+    # Index of the function, registers holding the free variables
+    [int, CapturedRegister|DeclRegister...] operands;
```
jclark: int isn't an operand

compiler/modules/bir/bir.bal
```diff
@@ -42,17 +42,28 @@ public type Label int;
 # A RegionIndex is an index of a region in the regions array.
 public type RegionIndex int;
 
+public type Function FunctionDefn | AnonFunction;
+
+public type FunctionBase record {|
```
jclark, line 47: InternalFunctionBase

compiler/modules/bir/bir.bal
```diff
@@ -663,6 +691,14 @@ public type CallIndirectInsn readonly & record {|
     boolean restParamIsList;
 |};
 
+# Create a closure by enclosing the first operand with the remaining operands.
```
jclark: Create a function value from an AnonFunction.
The operands are the values for the captured registers.

compiler/modules/front/expr.bal
```diff
@@ -47,11 +47,13 @@ type AssignmentBinding record {|
     Position pos;
 |};
 
+type FunctionMarker "func";
```
jclark: Comment explaining abbrev

compiler/modules/nback/symbol.bal
```diff
@@ -61,6 +61,10 @@ function mangleInternalSymbol(bir:ModuleId modId, bir:InternalSymbol symbol) ret
     return "_B_" + symbol.identifier;
 }
 
+function anonFunctionSymbol(int n) returns string {
+    return ".anonFn" + n.toString();
```
jclark: .anonFunction for consistency with .uniformFunction

compiler/modules/bir/bir.bal
heshanpadmasiri: Fixed

compiler/modules/front/expr.bal
heshanpadmasiri: Fixed

compiler/modules/bir/bir.bal
heshanpadmasiri: Fixed

jclark (approved):

PR #1230, opened 2023-08-24
Fix breaking code after SL U8

rdulmina:
## Purpose
Fix breaking code after https://github.com/ballerina-platform/ballerina-lang/pull/41117

### DIFF
--- a/compiler/showTypes.bal
+++ b/compiler/showTypes.bal
@@ -19,7 +19,7 @@ function subtypeRels(front:SourcePart[] sources) returns string[]|err:Diagnostic
 
     var tc = t:typeContext(env);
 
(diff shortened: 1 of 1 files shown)


pcnfernando (approved):
As per 
> If no query-construct-type is specified, the applicable contextually expected type determines the basic type of the value constructed. If there is no contextually expected type, then the basic type of the value constructed is the basic type of expression following in; it is a compile-time error if the static type of this expression is not a subtype of one of the basic types list, table, string or xml.

LGTM

jclark (approved):

jclark:
Yes, I agree that unfortunately this is wrong according to the spec.

PR #1231, opened 2023-09-04
Fix compiler warning in SL U8

heshanpadmasiri:
We get a warning on [`ComplexSemType`](https://github.com/ballerina-platform/nballerina/blob/eba6100af833071f45d0c81ca2548c5e57148477/compiler/modules/types/core.bal#L352) since it exposes `ProperSubtypeData` which is not public. Also, note each individual type in the union is already public.

### DIFF
--- a/compiler/modules/types/core.bal
+++ b/compiler/modules/types/core.bal
@@ -309,7 +309,7 @@ public class Context {
     }
 }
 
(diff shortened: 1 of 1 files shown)


jclark (approved):

jclark:
Not actually a bug in the code according to the spec, but it's fine to change.

PR #1228, opened 2023-08-17
Implement closures for final values

heshanpadmasiri:
## Purpose
Implement closures by extending #1223
Resolves #1225 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests
   > Details about the test cases and coverage

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes/no
 - Ran FindSecurityBugs plugin and verified report? yes/no
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes/no

PR #1232, opened 2023-09-04
Implement closures

### LINKED ISSUE #909: Implement closures
This builds on #908 to support closures i.e. allow capturing of local variables.

### LINKED ISSUE #1224: Extend BIR to support closures
Required for #908 and #909

# BIR instructions
## `ConstructLocalFunction` 
+ I think this should take as the operand some sort of handle that points to the `FunctionCode` for the lambda.
+ Then `nback` will lift the function to the module scope (see LLVM representation of closures)

# LLVM representation of closures
- LLVM doesn't have a concept of nested functions*, therefore we have to lift the function to the module level. 
  - LLVM has the[ `trampoline` intrinsic ](https://llvm.org/docs/LangRef.html#trampoline-intrinsics) which are used to support [GCC nested functions ](https://gcc.gnu.org/onlinedocs/gcc/Nested-Functions.html). But I don't think this is what we need since we want our closures to be in the heap not stack.
- Typically (I check c++ and rust) this lifted function will take a struct that "holds the closure" as the first argument and the rest will be the arguments to the lambda
  - In cpp this is the `%class.anon` part  
- Then at the beginning of the lambda you load each value in the closure struct to separate registers
- [examples](https://github.com/heshanpadmasiri/closure-tests)

cc @manuranga @jclark

heshanpadmasiri:
## Purpose
Extend  #1228 to support the capture of non-final values
Resolves #909 #1224


compiler/testSuite/15-closure/shadowing-e.bal
```diff
@@ -0,0 +1,20 @@
+import ballerina/io;
+type F function(int) returns int;
+
+function foo(boolean flag, int offset) returns F {
+    return function(int x) returns int {
+        int offset = 10; // @error
+        if (flag) {
+            return x + offset;
+        } else {
```
lochana-chathura:
```suggestion
        }
        else {
```

compiler/testSuite/15-closure/narrow-v.bal
```diff
@@ -0,0 +1,19 @@
+import ballerina/io;
+type F function(int) returns int;
+
+function foo(boolean flag, int offset) returns F {
+    return function(int x) returns int {
+        if (flag) {
+            return x + offset;
+        } else {
```
lochana-chathura:
```suggestion
        }
        else {
```

compiler/modules/bir/bir.bal
```diff
@@ -12,6 +12,7 @@ public type Module object {
     // which are in arrays in this.
     public function getTypeContext() returns t:Context;
     public function getFunctions() returns Function[];
+    public function isEnclosingFunction(int i) returns boolean;
```
jclark: hasCaptureInsn

compiler/modules/front/stmt.bal
```diff
@@ -41,7 +41,7 @@ type ClosureContext object {
 };
 
 type CapturedRegisterMemo readonly & record {|
-    int captured;
+    bir:CapturableRegister captured;
```
jclark: This is the one in outer function

compiler/modules/front/stmt.bal
```diff
@@ -41,7 +41,7 @@ type ClosureContext object {
...
     bir:CapturedRegister reg;
```
jclark: Calling this `reg` is not good.

Maybe inner/outer for the two fields.

compiler/modules/bir/bir.bal
heshanpadmasiri: Fixed

compiler/modules/front/stmt.bal
heshanpadmasiri: Fixed

compiler/modules/front/stmt.bal
heshanpadmasiri: Fixed

jclark (approved):

PR #1235, opened 2023-09-25
Fix type testing with FunctionConstOperands

### LINKED ISSUE #1234: Type testing with `FunctionConstOperands` fail to compile
Currently in [`finishCodeGenTypeTest`](https://github.com/ballerina-platform/nballerina/blob/2ff356ad6ab269c43fb35059f059e76e712ea9a3/compiler/modules/front/expr.bal#L1430) we have the assumption that if the type is not singleton value must be a `bir:Register`. This is not necessarily true since `FunctionConstOperands` are also not singleton. Currently we can't compile compile code such as,
1. Direct reference to functions
 ```
import ballerina/io;
function foo(string x, string y) returns (string) {
        return x + y;
}
public function main() {
        io:println(foo is function (string, string|int) returns (string)); // @output false

}
 ```
2. Directly using anonymous functions that don't capture (internally same as 1) 
```
import ballerina/io;
public function main() {
    io:println(function(string x, string y) returns (string) {
        return x + y;
    } is function (string, string|int) returns (string)); // @output false
}
```

heshanpadmasiri:
## Purpose
Fix type testing with `FunctionConstOperands`
Resolves #1234


compiler/modules/front/expr.bal
```diff
@@ -1425,8 +1425,13 @@ function finishCodeGenTypeTest(ExprContext cx, t:SemType semType, bir:Operand op
     if t:isEmpty(tc, intersect) {
         return { result: singletonBooleanOperand(tc, negated), block: nextBlock };
     }
-    bir:TmpRegister result = cx.createTmpRegister(t:BOOLEAN, pos);
     // Either diff or intersect should be empty if the operand is singleton
+    if operand is bir:FunctionConstOperand {
```
jclark, line 1428: Let's avoid cast to bir:Register below.

heshanpadmasiri: Fixed

jclark: Shouldn't test against bir:FunctionConstOperand twice

jclark (approved):

PR #1241, opened 2023-10-24
Add a "release" build target to jni.llvm

heshanpadmasiri:
## Purpose
Add a single "release" build target that will,

1. Compile nBallerina compiler that compile `.bal` files to `.o` using the `jni.llvm`
2. Convert that into a native executable using GraalVM
3. Move that along with the runtime (`balrt.a`) and a script that user can use to run the nballerina compiler to a single directory


llvm_jni/dist/native/nballerina.sh
```diff
@@ -0,0 +1,47 @@
...
+            exit 1
+        fi
+    else
+        print_usage_and_exit 1
+    fi
+else
+    "$scriptDir"/./compiler --outDir "$buildDir" "$src"
+fi
+
+objects=$(find "$buildDir" -name "*.o" | tr '\n' ' ')
```
jclark: This won't work if there are spaces in filenames.

llvm_jni/dist/native/nballerina.sh
```diff
@@ -0,0 +1,47 @@
...
+src="$1"
+buildDir="$(pwd)/build"
+rm -rf "$buildDir"
+mkdir -p "$buildDir"
+
+c_compiler=cc
+runtime="$scriptDir/./balrt.a"
+if [ $# -gt 1 ]; then
+    if [ "$2" = "--target" ]; then
+        if [ $# -lt 3 ]; then
```
jclark, line 27: It should be a usage error if the next argument is not aarch64.

llvm_jni/dist/native/nballerina.sh
jclark: Will the .o files be at any depth?

llvm_jni/dist/native/nballerina.sh
```diff
@@ -0,0 +1,47 @@
...
+fi
+
+if [ "$1" = "--help" ] || [ "$1" = "-h" ]; then
+    print_usage_and_exit 0
+fi
+
+scriptDir=$(dirname "$0")
+src="$1"
+buildDir="$(pwd)/build"
+rm -rf "$buildDir"
```
jclark, line 20: Doing rm -fr on a directory in the current directory seems a bit drastic. What if the user already had created a `build`` directory?

jclark:
Can we use a more specific wildcard to catch just the o files we want rather than removing the whole directory?

llvm_jni/dist/native/nballerina.sh
```diff
@@ -0,0 +1,51 @@
...
+    fi
+else
+    "$scriptDir"/./compiler --outDir "$buildDir" "$src"
+fi
+
+srcName=$(basename "$src" .bal)
+objects=()
+while IFS= read -r -d $'\0' object; do
+  objects+=("$object")
+done < <(find "$buildDir" -name "$srcName*.o" -print0)
```
jclark: Why can't we use a wildcard here?

I think you're relying on bash features, so the first line should reference /bin/bash rather than /bin/sh.

llvm_jni/dist/native/nballerina.sh
heshanpadmasiri: Fixed and limited the search depth to 1

llvm_jni/dist/native/nballerina.sh
heshanpadmasiri: Fixed

llvm_jni/dist/native/nballerina.sh
heshanpadmasiri: Fixed (Re-implemented using sh only)

heshanpadmasiri:
[Build artifacts](https://drive.google.com/drive/folders/1Rvqo2_G5-gM-iTVEzLC7of91WV1hJW0H?usp=share_link)

jclark (approved):

PR #1245, opened 2023-11-15
Fix verification complaints from llvm

### LINKED ISSUE #1244: Calculating sub-element address using GEP fails on AArch64
In AArch64 (both Apple and Linux) some instances of calculating sub-element address using GEP leads to invalid results. (Note that reading these address don't cause page faults, but data at those address is invalid). Fallowing two instances exhibit this behavior,
1. [Reading list length](https://github.com/ballerina-platform/nballerina/blob/02cd742092d094d122d6f709b366225589525602/compiler/modules/nback/structure.bal#L236) 
2. [Reading code pointer in function values](https://github.com/ballerina-platform/nballerina/blob/02cd742092d094d122d6f709b366225589525602/compiler/modules/nback/function.bal#L172)

~Most likely culprit is in our LLVM IR we don't explicitly provide the alignment, which causes LLVM to assume ABI alignment where as Clang uses explicit alignments.~ (Validated alignment is correct for AArch64 as well)

Required for #1241

heshanpadmasiri:
## Purpose
+ Add a module verification stage to jni.llvm to ensure "in memory representation" of the module is valid
  + Fixed raised issues by validator
+ Add a more simplified optimization pipeline using the LLVM's `PassManagerBuilder`
Resolves #1244

### DIFF
--- a/compiler/modules/nback/build.bal
+++ b/compiler/modules/nback/build.bal
@@ -91,7 +91,7 @@ final RuntimeFunction taggedToIntFunction = {
         returnType: "i64",
         paramTypes: [LLVM_TAGGED_PTR]
     },
-    attrs: []
+    attrs: [[0, "readonly"]]
 };
 
 final RuntimeFunction taggedToFloatFunction = {
@@ -100,7 +100,7 @@ final RuntimeFunction taggedToFloatFunction = {
         returnType: "double",
         paramTypes: [LLVM_TAGGED_PTR]
     },
-    attrs: []
+    attrs: [[0, "readonly"]]
 };
 
 final RuntimeFunction taggedClearExactAnyFunction = {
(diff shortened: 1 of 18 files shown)


compiler/modules/nback/build.bal
```diff
@@ -91,7 +91,7 @@ final RuntimeFunction taggedToIntFunction = {
         returnType: "i64",
         paramTypes: [LLVM_TAGGED_PTR]
     },
-    attrs: ["readonly"]
+    attrs: [[0, "readonly"]]
```
jclark: This is rather ugly. I don't yet understand the problem this is solving, but I hope we can find something nicer.

jclark:
Can I merge this?

jclark (approved):

heshanpadmasiri:
> Can I merge this?

Yes

